PR #9, opened 2019-02-22
Improved performance of loading embedding matrix

qubixes:
Use Queue's from the multiprocessing library to speed up the loading times of the embedding layer. Also, use the knowledge of the sampling to not having to load the whole matrix into memory.

On a MacBook Pro Retina 2012 there is a 10x speed up, now loading in approximately 30s.

### DIFF
--- a/asr/models/embedding.py
+++ b/asr/models/embedding.py
@@ -1,9 +1,96 @@
-"""Script to create the memory of an embeeding layer."""
+from multiprocessing import Process, Queue, cpu_count
 
 import numpy as np
 
 
-def load_embedding(fp, verbose=1):
+def _embedding_reader(filename, input_queue, block_size=1000):
+    """ Process that reads the word embeddings from a file.
+
+    Parameters
+    ----------
+    filename: str
+        File of trained embedding vectors.
+    input_queue: Queue
+        Queue to store jobs in.
+    block_size: int
+        Number of lines for each job.
+    """
+
+    with open(filename, 'r', encoding='utf-8', newline='\n') as f:
+        # Throw away the first line, since we don't care about the dimensions.
+        f.readline()
+
+        i_line = 0
+        buffer = []
+        # Read the embedding file line by line.
+        for line in f:
+            i_line += 1
+            buffer.append(line)
+            # If the buffer is full, write it to the queue.
+            if i_line == block_size:
+                input_queue.put(buffer)
+                i_line = 0
+                buffer = []
+        if i_line > 0:
+            input_queue.put(buffer)
+
+    # Put the string "DONE" in the queue, to ensure that the
+    # worker processes finish.
+
+    input_queue.put("DONE")
+
+
+def _embedding_worker(input_queue, output_queue, emb_vec_dim, word_index=None):
+    """ Process that reads the word embeddings from a file.
+
+    Parameters
+    ----------
+    input_queue: Queue
+        Queue in which the jobs are submitted.
+    output_queue: Queue
+        Queue to store the embedding in dictionary form.
+    emb_vec_dim: int
+        Dimension of each embedding vector.
+    word_index: dict
+        Dictionary of the sample embedding.
+    """
+
+    embedding_dict = {}
+    badInput = False
+    badValues = {}
+    while not badInput:
+        buffer = input_queue.get()
+        if buffer == "DONE":
+            break
+
+        for line in buffer:
+            values = line.split(' ')
+
+            if len(values) != emb_vec_dim + 2:
+                badInput = True
+                badValues = values
+                break
+
+            word = values[0]
+            if word_index is not None and word not in word_index:
+                continue
+            coefs = values[1:emb_vec_dim + 1]
+
+            # store the results
+            embedding_dict[word] = np.asarray(coefs, dtype=np.float32)
+
+    # We removed the "DONE" from the input queue, so put it back in for
+    # the other processes.
+    input_queue.put("DONE")
+
+    # Store the results in the output queue
+    if badInput:
+        output_queue.put({"ErrorBadInputValues": badValues})
+    else:
+        output_queue.put(embedding_dict)
+
+
+def load_embedding(fp, word_index=None, n_jobs=None, verbose=1):
     """Load embedding matrix from file.
 
     The embedding matrix needs to be stored in the
@@ -13,8 +100,13 @@ def load_embedding(fp, verbose=1):
     ----------
     fp: str
         File path of the trained embedding vectors.
+    word_index: dict
+        Sample word embeddings.
+    n_jobs: int
+        Number of processes to parse the embedding (+1 process for reading).
     verbose: int
         The verbosity. Default 1.
+
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


J535D165:
Nice PR and nice speed up.

Is there a reason to choose queus over a pool?

qubixes:
I'm no expert on either, but Queue's seem more convenient with a simultaneous IO/worker process configuration? Pool's seem more useful when all the data is already present.

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
-"""Script to create the memory of an embeeding layer."""
+'''
+Created on 22 Feb 2019
```
J535D165: Remove

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
...
+
+@author: Raoul Schram, Parisa Zahedi, Jonathan de Bruin
```
J535D165: remove

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
...
+
+Multiprocessing version of embedding file. 
+'''
 
 import numpy as np
 
+from multiprocessing import Process, Queue
```
J535D165: CPython imports before external imports. PEP8

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
...
+
+def _embedding_reader(filename, input_queue, block_size=1):
```
J535D165: default block size 1000, remove in #L108 and #L134.

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
...
+    while True:
+        buffer = input_queue.get()
+        if buffer == "DONE":
+            break
+        
+        for line in buffer:
+            values = line.split(' ')
+            
+            if len(values) != d + 2:
+                raise ValueError("Check values: %s" % values)
```
J535D165: Use `f"Check values: {values}"`

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
...
+            embedding_dict[word] = np.asarray(coefs, dtype=np.float32)
 
-def load_embedding(fp, verbose=1):
+    #We removed the "DONE" from the input queue, so put it back in for the other processes.
+    input_queue.put("DONE")
+    #Store the results in the output queue
+    output_queue.put(embedding_dict)
+    
+
+def load_embedding(fp, verbose=1, word_index=None):
```
J535D165: Make `verbose` third argument.

asr/models/embedding.py
```diff
@@ -15,42 +94,58 @@ def load_embedding(fp, verbose=1):
...
     dict:
         The embedding weights stored in a dict with the word as key and
         the weights as values.
     """
-
-    embedding = {}
-
+    
+    #Granularity of the worker jobs.
+    block_size=1000
```
J535D165: remove

asr/models/embedding.py
```diff
@@ -15,42 +94,58 @@ def load_embedding(fp, verbose=1):
...
+    #Maximum number of jobs in the queue.
+    queue_size=500
+    #Number of worker jobs running in parallel.
+    n_worker=3
```
J535D165: Make this an argument named `n_jobs` (sklearn). Default 1. Value -1 should be all cores.

asr/models/embedding.py
```diff
@@ -15,42 +94,58 @@ def load_embedding(fp, verbose=1):
...
         the weights as values.
     """
-
-    embedding = {}
-
+    
+    #Granularity of the worker jobs.
+    block_size=1000
+    #Maximum number of jobs in the queue.
+    queue_size=500
```
J535D165: Hardcoding looks fine here

asr/models/embedding.py
```diff
@@ -15,42 +94,58 @@ def load_embedding(fp, verbose=1):
...
-            coefs = values[1:d + 1]
-
-            # store the results
-            embedding[word] = np.asarray(coefs, dtype=np.float32)
-
+    embedding = {}
+    
+    worker_procs=[]
+    for _ in range(n_worker):
+        p = Process(target=_embedding_worker,
```
J535D165: outline

J535D165: We sometimes use yapf.

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
...
+    block_size: int
+        Number of lines for each job.
+    """
+    
+    with open(filename, 'r', encoding='utf-8', newline='\n') as f:
+        f.readline()
+        
+        ib=0
+        buffer=[]
+        for _, line in enumerate(f): # This can be done better probably...
```
J535D165: remove comment

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
...
+    filename: str
+        File of trained embedding vectors.
+    input_queue: Queue
+        Queue to store jobs in.
+    block_size: int
+        Number of lines for each job.
+    """
+    
+    with open(filename, 'r', encoding='utf-8', newline='\n') as f:
+        f.readline()
```
J535D165: Your aren't using the info. You can use this to improve for loop.

asr/models/embedding.py
J535D165: While loop with `try except` is what you are looking for.

asr/models/embedding.py
```diff
@@ -1,9 +1,88 @@
...
+def _embedding_worker(input_queue, output_queue, d, word_index=None):
+    """ Process that reads the word embeddings from a file.
+    
+    Parameters
+    ----------
+    input_queue: Queue
+        Queue in which the jobs are submitted.
+    output_queue: Queue
+        Queue to store the embedding in dictionary form.
+    d: int
```
J535D165: Explicit variables names `import zen`

asr/review.py
```diff
@@ -43,8 +43,8 @@ def _load_embedding_matrix(fp, word_index):
 
     from asr.models.embedding import load_embedding, sample_embedding
 
-    embedding, words = load_embedding(fp)
```
J535D165: good catch

J535D165:
Shall we add a unittest before merging?

J535D165:
And a function to download the embedding?

qubixes:
I added unit tests, including testing for the URL you gave me. (Obviously it all passes.) 

I think all the things you mentioned are resolved.

One possible issue: if one of the processes gets an exception, the others will deadlock. Obviously, the error will still be raised. Solution would be a try/except block that only raises an error after everything will be joined. But of course the whole thing would be one more indented to the right...

J535D165:
Please rebase.

J535D165:
Can you add the following line to the __init__.py of the module? 

```
from asr.models.embedding import load_embedding, sample_embedding
```

qubixes:
Should be able to merge now?

J535D165:
What time should I think of? It takes ages (or hangs) on my machine. 

``` python
In [1]: import asr                                                                                                 

In [2]: %timeit asr.load_embedding("/Users/jonathan/asr_data/fasttext.cc.en.300.vec")                              
Reading 2000000 vectors with 300 dimensions.
```

qubixes:
Hmmm, with 1 core and no sampling it should be similar to what it was before, so 5 min perhaps (depending on hardware obviously)?

On 12 Mar 2019, at 16:30, Jonathan de Bruin <notifications@github.com<mailto:notifications@github.com>> wrote:


What time should I think of? It takes ages (or hangs) on my machine.

In [1]: import asr

In [2]: %timeit asr.load_embedding("/Users/jonathan/asr_data/fasttext.cc.en.300.vec")
Reading 2000000 vectors with 300 dimensions.

PR #13, opened 2019-02-26
words is removed from load_embedding() and sample_embedding() parameters

parisa-zahedi:
found a minor error in review.py while loading embedding


J535D165:
Good catch. @qubixes ran into the same problem. See PR #9. I will merge #13 first because #9 is not ready yet.

PR #26, opened 2019-03-12
Add functions to automate downloading embedding files

J535D165:
Function to automate the process of downloading (and storing) embedding files. 

```python
from asr.models.embedding import download_embedding

download_embedding()
```

Use the argument `data_home` to change the default folder. 

One can set the environment variable to change the default folder.

```
import os

from asr.models.embedding import download_embedding

# set the environment variable
os.environ['ASR_DATA'] = "~/my_asr_embedding_files"

# download the files
download_embedding()

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,35 @@ next to this repository. Now you can debug code with the code below:
 python -m asr oracle ../automated-systematic-review-simulations/pickle/ptsd_vandeschoot_words_20000.pkl --n_instances 5
 ```
 
+### Embedding files
+
+Embedding files contains pretrained model weights. The weights are used as
+prior knowledge of the neural network. By default, these weights are stored in
+the users `~/asr_data` folder. You can download embedding files with the
+following command:
+
+```python
+from asr.models.embedding import download_embedding
+
+download_embedding()
+```
+
+One can set the environment variable to change the default folder.
+
+```
+import os
+
+from asr.models.embedding import download_embedding
+
+# set the environment variable
+os.environ['ASR_DATA'] = "~/my_asr_embedding_files"
+
+# download the files
+download_embedding()
+
+```
+
+
 ## Publications
 
 - Dutch newspaper NRC on this project ["Software vist de beste artikelen uit een bibliotheek van duizenden."](https://www.nrc.nl/nieuws/2019/01/14/software-vist-de-beste-artikelen-eruit-a3628952)
--- a/asr/models/embedding.py
+++ b/asr/models/embedding.py
@@ -1,6 +1,19 @@
+import gzip
+import io
+import os
+import shutil
 from multiprocessing import Process, Queue, cpu_count
+from urllib.request import urlopen
 
 import numpy as np
+
+from asr.utils import get_data_home
+
+
+EMBEDDING_EN = {
+    "url": "https://dl.fbaipublicfiles.com/fasttext/vectors-crawl/cc.en.300.vec.gz",  # noqa
+    "name": 'fasttext.cc.en.300.vec'
+}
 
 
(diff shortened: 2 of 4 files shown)


J535D165:
Fix: download files if and only if they don't exist.

J535D165:
I think the new code for the simulations project will look like this:

```python

from pathlib import Path

from asr.models.embedding import EMBEDDING_EN, download_embedding, load_embedding, sample_embedding
from asr.utils import get_data_home

# download data
fp = Path(get_data_home(), EMBEDDING_EN['name'])
if not fp.exists():
   download_embedding()

# read and sample matrix
embedding = load_embedding(PATH_TO_EMBEDDING)
embedding_matrix = sample_embedding(embedding, words)

```

qubixes:
Yes, that makes sense. I still think it makes sense to warn the user that they're about to download a big file (disk + data plan + time usage). Obviously at the moment not the highest priority (since we don't have so many users..), but wouldn't it be nice to have something of a stub, then upgrade it when we're in "production"?

J535D165:
Thanks for the feedback. Good point. For now, the most important thing is to improve our workflow. 

But let's think about a solution.

- A warning/user input and a flag to suppress this. 
- Build a download center with a new entry point. 

Btw, this is a nice idea https://github.com/chakki-works/chakin. But the implementation is poor imo.

J535D165:
Thanks for the feedback. Good point. For now, the most important thing is to improve our workflow. 

But let's think about a solution.

- A warning/user input and a flag to suppress this. 
- Build a download center with a new entry point. 

Btw, this is a nice idea https://github.com/chakki-works/chakin. But the implementation is not very flexible.

qubixes:
I think a download center seems a bit overkill right now. Given the application, it seems unlikely we'll support 10 languages, right? 

I think the first solution makes sense (user input). It should only ask once, so I don't think that would be too much of a bother. They can always type `yes | asr` or `echo "y" | asr`, no?  :)

J535D165:
> I think a download center seems a bit overkill right now. Given the application, it seems unlikely we'll support 10 languages, right?

Not 10 languages, but 10 domain specific pertrained models sounds reasonable. 

> I think the first solution makes sense (user input). It should only ask once, so I don't think that would be too much of a bother. They can always type yes | asr or echo "y" | asr, no? :)

What do you think about the simulation mode?

qubixes:
> Not 10 languages, but 10 domain specific pre-trained models sounds reasonable.

That makes sense. The models would be much smaller than the embeddings I suppose? Either way, I'm also fine with some kind of download manager if that is what you prefer to implement now. 

> What do you think about the simulation mode?

I'm not sure what you mean? You mean some kind of non-interactive mode for simulations? I don't think it makes much sense to download (big) files in a non-interactive mode. I mean, it probably isn't what a user wants if he submits the simulation to a cluster and it starts downloading at xxx kb/s.

J535D165:
> The models would be much smaller than the embeddings I suppose? 

No. Similar. 

> I'm not sure what you mean? You mean some kind of non-interactive mode for simulations? I don't think it makes much sense to download (big) files in a non-interactive mode. I mean, it probably isn't what a user wants if he submits the simulation to a cluster and it starts downloading at xxx kb/s.

Simulation mode is also for the researcher. To measure the potential speedup of the SR. We are using it for a simulation study, but this isn't the same.

J535D165:
Merged, but needs improvement in the future.

qubixes:
https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

On 22 Mar 2019, at 10:37, Jonathan de Bruin <notifications@github.com> wrote:


Merged, but needs improvement in the future.

PR #36, opened 2019-03-26
Add CLI option to set class weights

qubixes:


J535D165:
Thanks! Let's write some documentation on this as well. It is crucial for good results, isn't it?

PR #41, opened 2019-04-03
Add log readers

J535D165:
Based on the work op @qubixes in simulations project. 

Log readers output `Logger` objects. 

@qubixes Shall we rename `Logger` into `ALLog`. Logger is also used in the default lib `logging` I think.


J535D165:
@qubixes Can you review this one?

asr/logging.py
```diff
@@ -0,0 +1,190 @@
...
+        File path of the log file.
+
+    Returns
+    -------
+    Logger:
+        A Logger object with logs.
+    """
+    log_fp_path = Path(log_fp)
+
+    if log_fp_path.is_file() and log_fp_path.suffix == ".json":
```
qubixes: I don't think checking for suffix should be necessary? Just try/except when loading the file?

asr/logging.py
```diff
@@ -0,0 +1,190 @@
...
+            log = Logger()
+            log._log_dict = json.load(f)
+
+            return log
+
+    else:
+        raise ValueError(f"Incorrect file path {log_fp}")
+
+
+def read_logs_from_dir(log_dir, prefix=None):
```
qubixes: How about prefix="", then by default it will use all files in the directory?

asr/logging.py
J535D165: Great! Will change this one.

asr/logging.py
J535D165: This is a bug. `""` will fail as well atm. Will fix this one.

PR #46, opened 2019-04-29
Balance rewrite

qubixes:
Final pull request for the new LSTM model.

* Add new model: lstm_pool, with no dense layer, and max pooling.

* Rewrote review function of Review class
** Simplified code by only training within the loop.

* Always try to work with absolute indices

* Loop abstract instead of padding with zeros.
** Should not have a big effect on the original lstm model, but does for the new one.

* New query strategy: Combination of random and max sampling.

* New training data balance strategies for quicker learning.
** Simple (default): use all training data once
** Undersample: undersample the 0's to make the ratio of 1's to 0's more balanced.
** Triple_Balance: Use a precarious balance of 1's and 0's as well as a balance between max and randomly sampled instances (0's).


qubixes:
Let's merge?

PR #47, opened 2019-04-30
Add info on the Active Learning process

J535D165:


J535D165:
@Sasafrass Can you provide me with the XML file of the image?

Issue #50, opened 2019-05-17
Fix API to be more usable


qubixes:
At the moment using the API one needs a significant amount of hoops to jump through. It should be easier in my opinion.

PR #53, opened 2019-05-20
Make the defaults of the API work.

qubixes:
Related issue: #50

I will update the README afterwards.


qubixes:
Btw, if we're going to break backward compatibility of read_data, we might as well immediately remove load_data.

Also, let's update the version number of ASR.

J535D165:
Closes #55

PR #63, opened 2019-07-16
Asr rename

qubixes:
Changed all the 'asr's into asreview's. I'll go through the docs as well.


setup.py
```diff
@@ -56,8 +56,7 @@
 
     entry_points={
         'console_scripts': [
-            'asr=asr.__main__:main',
```
J535D165: Please keep `asr` with a deprecation warning.

qubixes: You mean only the entry point, or the whole library (I.e. being able to ‘import asr’)?

J535D165: This remark is on the entry point. That's a must-have. The general remark is on the `import asr`.

qubixes:
I'm trying to find ways to do it, but from what I read it looks like it's not how it's normally done:

https://www.python.org/dev/peps/pep-0423/

I don't think our case is really "exceptional", is it? We'd also get into conflict with the asr project (I guess we are technically at the moment), though it has an inactive status.

 Personally, I prefer a clean break? Let me know if you find a way to do it cleanly.

I'll re-add the entry-point.

PR #67, opened 2019-07-18
Fix Travis CI testing

qubixes:
Testing fails on travis using the latest 1.14 tensorflow package. Probably due to memory limitations. Locally the testing works.


J535D165:
Danke danke @qubixes

PR #69, opened 2019-07-24
Update configuration parsing more arguments

qubixes:
Little bit of an ugly hack to set the defaults to -1, but I don't have a neater solution... 

I suggest if we merge the pull request, that we up the version number, since it breaks forward compatibility (while maintaining backward compatibility) with configuration files.

### DIFF
--- a/asreview/__main__.py
+++ b/asreview/__main__.py
@@ -6,7 +6,6 @@
 import sys
 import argparse
 import warnings
-import os
 
 import numpy as np
 
@@ -18,6 +17,8 @@ DEFAULT_MODEL = "lstm_base"
 DEFAULT_QUERY_STRATEGY = "uncertainty"
 DEFAULT_BALANCE_STRATEGY = "simple"
 DEFAULT_N_INSTANCES = 50
+DEFAULT_N_PRIOR_INCLUDED = 10
+DEFAULT_N_PRIOR_EXCLUDED = 10
 
 
 def parse_arguments(mode, prog=sys.argv[0]):
@@ -98,14 +99,14 @@ def parse_arguments(mode, prog=sys.argv[0]):
         # Initial data (prior knowledge)
         parser.add_argument(
             "--n_prior_included",
-            default=None,
+            default=DEFAULT_N_PRIOR_INCLUDED,
             type=int,
             help="Sample n prior included papers. "
                  "Only used when --prior_included is not given.")
 
         parser.add_argument(
             "--n_prior_excluded",
-            default=None,
+            default=DEFAULT_N_PRIOR_EXCLUDED,
             type=int,
             help="Sample n prior excluded papers. "
                  "Only used when --prior_excluded is not given.")
--- a/asreview/__main__.py
+++ b/asreview/__main__.py
@@ -18,6 +18,9 @@ DEFAULT_MODEL = "lstm_base"
 DEFAULT_QUERY_STRATEGY = "uncertainty"
 DEFAULT_BALANCE_STRATEGY = "simple"
 DEFAULT_N_INSTANCES = 50
+DEFAULT_N_QUERIES = 2
+DEFAULT_N_PRIOR_INCLUDED = 10
+DEFAULT_N_PRIOR_EXCLUDED = 10
 
 
 def parse_arguments(mode, prog=sys.argv[0]):
@@ -60,7 +63,7 @@ def parse_arguments(mode, prog=sys.argv[0]):
     parser.add_argument(
         "--n_queries",
         type=int,
-        default=None,
+        default=DEFAULT_N_QUERIES,
         help="The number of queries. Default None"
     )
     parser.add_argument(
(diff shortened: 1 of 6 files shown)


J535D165:
> Little bit of an ugly hack to set the defaults to -1, but I don't have a neater solution...

Can you explain the issue this is solving? 

Maybe it is a good idea to start with release notes...

asreview/__main__.py
```diff
@@ -60,7 +63,7 @@ def parse_arguments(mode, prog=sys.argv[0]):
     parser.add_argument(
         "--n_queries",
         type=int,
-        default=None,
+        default=DEFAULT_N_QUERIES,
```
J535D165: Can you explain this one? This seems odd and confusing to me.

asreview/__main__.py
```diff
@@ -60,7 +63,7 @@ def parse_arguments(mode, prog=sys.argv[0]):
...
         help="The number of queries. Default None"
```
J535D165: Default not `None` any more.

asreview/review.py
```diff
@@ -41,12 +42,16 @@ def _default_settings(model, n_instances, query_strategy, balance_strategy,
...
+        "n_queries": n_queries,
+        "n_prior_included": n_prior_included,
+        "n_prior_excluded": n_prior_excluded,
         "mode": mode,
         "model_param": {},
         "fit_param": {},
         "query_param": {},
         "balance_param": {},
     }
+    print(settings)
```
J535D165: remove?

asreview/review.py
```diff
@@ -31,8 +31,9 @@
 from os.path import splitext
 
 
-def _default_settings(model, n_instances, query_strategy, balance_strategy,
-                      mode, data_fp):
+def _default_settings(model, n_instances, n_queries, n_prior_included,
```
J535D165: I think it is better to make this a class.

asreview/review.py
```diff
@@ -188,6 +196,7 @@ def review(dataset,
             query_strategy=query_fn,
             train_data_fn=train_data_fn,
             n_instances=n_instances,
+            n_queries=n_queries,
```
J535D165: See above. What is the intention?

qubixes:
Okay, give me some time, I'll find a better solution. At this point the first ugly hack (updating from configuration file) needs to many hacks to not break...

qubixes:
Check the new commit; I think it should fix all concerns.

asreview/__main__.py
```diff
@@ -98,17 +98,15 @@ def parse_arguments(mode, prog=sys.argv[0]):
         # Initial data (prior knowledge)
         parser.add_argument(
             "--n_prior_included",
-            default=10,
+            default=None,
             type=int,
-            nargs="*",
```
J535D165: good catch

asreview/__main__.py
```diff
@@ -98,17 +98,15 @@ def parse_arguments(mode, prog=sys.argv[0]):
         # Initial data (prior knowledge)
         parser.add_argument(
             "--n_prior_included",
-            default=10,
+            default=None,
```
J535D165: Why the revert? This was a good change I think.

asreview/base.py
```diff
@@ -216,6 +216,10 @@ def __init__(self,
         self.n_prior_excluded = n_prior_excluded
 
     def _prior_knowledge(self):
+        if self.n_prior_included is not None and self.n_prior_included < 0:
```
J535D165: I don't like the -1 option. If a -1 is given to the program, it should raise a ValueError (by design). At the moment, I don't see the actual problem.

asreview/settings.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC
+import os
+from asreview.utils import config_from_file, _safe_dict_update
+
+
+class ASReviewSettings(ABC):
```
J535D165: Great! Thanks. 

I don't see the reason to make this an ABC class. Inherit 'object'?

asreview/settings.py
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self, model, n_instances, n_queries, n_prior_included,
+                 n_prior_excluded, query_strategy,
+                 balance_strategy, mode, data_fp):
+        data_name = os.path.basename(data_fp)
+        self._settings = {
```
J535D165: Unnest?

qubixes: I'm not sure what you mean?

J535D165: There is no need to wrap the settings in an additional attribute named `_settings`. Please put them on class level. 

I'm not so sure on the type checking as well. Can you clarify this?

qubixes: It seems much easier to loop over elements of a dictionary than those of class elements though?

qubixes: The configuration file doesn't have any information about type, so instead of guessing, I explicitly convert it using these stored types.

PR #70, opened 2019-07-31
Add option to continue review from existing log file.

qubixes:
Pull request based on the other pending pull request. Will rebase if necessary.

### DIFF
--- a/asreview/settings.py
+++ b/asreview/settings.py
@@ -55,7 +55,8 @@ class ASReviewSettings(object):
                     try:
                         setattr(self, key, SETTINGS_TYPE_DICT[key](value))
                     except (KeyError, TypeError):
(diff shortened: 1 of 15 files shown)


asreview/settings.py
```diff
@@ -1,29 +1,65 @@
 import os
-from asreview.utils import config_from_file, _safe_dict_update
+from configparser import ConfigParser
+
+
+settings_type_dict = {
```
J535D165: Global var --> Uppercase

qubixes:
This solution doesn't trigger my OCD, and does what you want, no? The locals() solution is not particularly pretty, but...

J535D165:
Let's merge it. The new features are very cool.

PR #76, opened 2019-08-28
Add very basic mechanism for example datasets

J535D165:
Example usage:

```
asreview oracle example_ptsd
```


asreview/review.py
```diff
@@ -45,6 +45,10 @@ def review(dataset,
            **kwargs
            ):
 
+    # Find the URL of the datasets if the dataset is an exampe dataset.
```
qubixes: exampe -> example

PR #80, opened 2019-09-13
Review rewrite

qubixes:
Restructuring of the Review base class. Should allow for more flexibility in using the class, especially with regards to the granularity. It also makes the "review" member function of the Review base class more readable by having it comprised of smaller functions.

Also put all the Review classes in the a new review package (instead of module), and renamed the "review.py" module to "factory.py". 

Supplants both pull requests #79 and #77.


J535D165:
This looks good. Thanks @qubixes. Ready to merge?

qubixes:
Yes, I think so. There is still some possible improvement for starting the CLI modes with zero inclusions/exclusions, but let's do that in a separate PR.

J535D165:
Many thanks! Good work on restructuring the modules.

PR #89, opened 2019-09-19
Improved query management + README fix

qubixes:
The query counter is now updated after each training session.

The format of the log files have changed: Query 0 are the queries done before the first initial training. Query 1 has information collected *after* training (previously probabilities after training were stored in Query 0 as well). This makes it easier conceptually to understand where training takes place (between the queries).

The handling of query methods and storing this information has been improved.

There is new command line option: -n/--n_papers, which sets the number of papers to be reviewed, including the priors.


J535D165:
I agree on the implementation. Keep going!

Let's think about the variable names
```
n_papers -> n_documents/n_items?
n_instances -> n_batch?          
n_queries -> oke 
```

qubixes:
> I agree on the implementation. Keep going!
> 
> Let's think about the variable names
> 
> ```
> n_papers -> n_documents/n_items?
> n_instances -> n_batch?          
> n_queries -> oke 
> ```

You mean only for the CLI interface or the whole package?

PR #100, opened 2019-09-30
update team members and added hyperlinks

Rensvandeschoot:


J535D165:
Thanks for the update

PR #114, opened 2019-10-14
Fix undersampling (was cheating).

### LINKED ISSUE #113: Undersampling cheats/doesn't work as intended.

qubixes:
The included tests would have caught this particular bug.

resolves #113


J535D165:
Thanks for the unit test!!

PR #122, opened 2019-10-28
Tutorial on oracle mode for first time users

GerbrichFerdinands:
Basic tutorial on how to install and use the oracle mode with the ptsd_example dataset. 

Suggestions to add in the future:
* explanation on when to stop
* explanation on different models, query strategies etc. 
* explanation on reading the log_file

(some md issues need to be fixed, e.g. within document links are not working yet)


J535D165:
Thanks @GerbrichFerdinands. Thanks for the suggestions as well!

PR #124, opened 2019-10-31
Fancy cli

### LINKED ISSUE #7: Hyper-parameter optimisation for NB and SVM models
Find better values for the hyper-parameters of the NB and SVM models. The code should be part of https://github.com/msdslab/automated-systematic-review-simulations

### LINKED ISSUE #49: Add analysis class

qubixes:
Don't squash please!

Changelog:

- Introduce hooks for hyperopt optimization
-- Hyperopt is a package for hyperparameter optimization, and these hooks make it much easier to do the optimization of said hyperparameters.
-- The hyperopt package is optional.

- Model specific action are removed from the review package and moved in the model package.
- Models are now classes, instead of a generator function.

- Add HDF5 as a storage container for logging.
-- It works parallel to the still available JSON logger. It has an advantage in both speed and disk requirements (~x2.4).

- The JSON logger is updated, and it's version updated to 2.0, making it incompatible with log files created with an earlier version of ASReview. Automatically detected if this is the case.

- Add hooks for final_labels
-- This is a slightly obscure feature, where you can have for example two levels of inclusion:
--- After reading the abstract -> model trained on this
--- After reading full text
-- This is a simulation only feature that can be enabled by putting abstract_only=True in the [global_settings] section of a config file.

- Update the command line interface
-- Oracle only change.
-- This makes the command line interface much more useful, asking more questions and giving more options what to do next.
-- It uses the PyInquirer package, which unfortunately means that we cannot test oracle mode anymore with pytest (needs a TTY).

- The analysis part of the simulation project was ported over, since it can be generally useful.

- Documentation was updated.

Resolves #49 
Resolves #7


asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
...
+def _result_group(f, query_i):
+    try:
+        g = f[f'/results/{query_i}']
+    except KeyError:
+        g = f.create_group(f'/results/{query_i}')
+        g.attrs['creation_time'] = np.string_(datetime.now())
+    return g
+
+
+class HDF5_Logger(object):
```
J535D165: Naming conventions

asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def __str__(self):
+        log_str = "Attributes:\n--------------\n"
+        for key, value in self.f.attrs.items():
+            log_str += f"{key}: {value}\n"
+        log_str += "\n\nGroups+datasets:\n---------------\n"
+        log_str += str(list(self.f.keys()))
+        return log_str
+
+    def todict(self):
```
J535D165: Inconsistent name

asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
...
+        ---------
+        fp: str
+            The file path to export the results to.
+
+        """
+        self.f['end_time'] = str(datetime.now())
+        self.f.flush()
+
+    def get(self, variable, query_i=None, idx=None):
+        "Get variables from hdf5 file."
```
J535D165: Docstring

asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
...
+        if variable == "train_idx":
+            array = np.array(g["train_idx"], dtype=np.int)
+        if array is None:
+            return None
+        if idx is not None:
+            return array[idx]
+        return array
+
+    def restore(self, fp, read_only=False):
+        "Restore logger from hdf5 file."
```
J535D165: Docstring

asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
...
+                    f"Log cannot be read: logger version {self.version}, "
+                    f"logfile version {log_version}.")
+            settings_dict = json.loads(self.f.attrs['settings'])
+            if "mode" in settings_dict:
+                self.settings = ASReviewSettings(**settings_dict)
+        except KeyError:
+            self.create_structure()
+
+    def create_structure(self):
+        "Initialize (empty) structure."
```
J535D165: docstring

asreview/logging.py
```diff
@@ -1,230 +1,35 @@
...
-from datetime import datetime
-from pathlib import Path
+import os
 
-import asreview
-import numpy as np
-from asreview.settings import ASReviewSettings
 
+def get_logger_class(fp):
+    "Get logging class from file extension."
```
J535D165: dosctring

asreview/json_logging.py
```diff
@@ -0,0 +1,238 @@
...
+
+import numpy as np
+
+import asreview
+from asreview.settings import ASReviewSettings
+from asreview.analysis.statistics import _get_labeled_order
+from asreview.analysis.statistics import _get_last_proba_order
+
+
+class Dict_Logger(object):
```
J535D165: Name. 

Should this be an abstract base class?

asreview/json_logging.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def close(self):
+        if not self.read_only:
+            self.save()
+
+    def save(self):
+        print(self)
+
+
+class JSON_Logger(Dict_Logger):
```
J535D165: Name

asreview/json_logging.py
```diff
@@ -0,0 +1,238 @@
...
+    def __init__(self, log_fp, *_, read_only=False, **__):
```
J535D165: I don't like this syntax :) Does `def __init__(self, log_fp, read_only=False):` fail?

asreview/logging.py
```diff
@@ -1,230 +1,35 @@
...
-
-        """
-        self._log_dict["settings"] = copy.deepcopy(vars(self.settings))
-        self._log_dict["settings"]["query_kwargs"].pop("pred_proba", None)
-        self._log_dict["settings"]["query_kwargs"].pop("query_src", None)
-        self._log_dict["settings"]["query_kwargs"].pop("current_queries", None)
-        self._log_dict.move_to_end("settings", last=False)
-        self._log_dict["time"]["end_time"] = str(datetime.now())
-        fp = Path(fp)
+    "Wrapper Logger class to easily get the right logger from file extension."
```
J535D165: docstring

asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
...
+    def add_settings(self, settings):
+        self.settings = settings
+        self.f.attrs.pop('settings', None)
+        self.f.attrs['settings'] = np.string_(json.dumps(vars(settings)))
+
+    def n_queries(self):
+        return len(self.f['results'].keys())
+
+    def review_state(self):
+        "Get internal review state variables from hdf5 file."
```
J535D165: docstring

asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
...
+            g.create_group("new_labels")
+
+        g = g['new_labels']
+
+        np_methods = np.array(list(map(np.string_, methods)))
+        _append_to_dataset('idx', idx, g, dtype=np.int)
+        _append_to_dataset('labels', labels, g, dtype=np.int)
+        _append_to_dataset('methods', np_methods, g, dtype='S20')
+
+#         bool_array = np.full(self.n_labels(), False, dtype=bool)
```
J535D165: remove

asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
+import json
+from datetime import datetime
+from pathlib import Path
+
+import h5py
+import numpy as np
+import itertools
```
J535D165: import order

asreview/hdf5_logging.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def add_settings(self, settings):
+        self.settings = settings
+        self.f.attrs.pop('settings', None)
+        self.f.attrs['settings'] = np.string_(json.dumps(vars(settings)))
+
+    def n_queries(self):
+        return len(self.f['results'].keys())
+
+    def review_state(self):
```
J535D165: I'm not sure what this is doing. A docstring would be nice. Is this a private method?

asreview/json_logging.py
```diff
@@ -0,0 +1,238 @@
...
+
+import asreview
+from asreview.settings import ASReviewSettings
+from asreview.analysis.statistics import _get_labeled_order
+from asreview.analysis.statistics import _get_last_proba_order
+
+
+class Dict_Logger(object):
+    """Class for logging the Systematic Review with no storage."""
+    def __init__(self, *_, **__):
```
J535D165: A general remark. Think about the structure of the Class. This syntax is a bit odd. 

```
class A(object)
    version = "1.0"
    read_only = True
    def __init__(self):
        pass
```
or 
```
class A(object)
    def __init__(self, version="1.0", read_only=True):
         self.version = version
         self.read_only = read_only
```

Or a combination of both

asreview/logging.py
```diff
@@ -1,230 +1,35 @@
...
-from collections import OrderedDict
-from datetime import datetime
-from pathlib import Path
+import os
 
-import asreview
-import numpy as np
-from asreview.settings import ASReviewSettings
 
+def get_logger_class(fp):
```
J535D165: `fp=None`

asreview/balance_strategies/triple_balance.py
```diff
@@ -15,26 +15,36 @@ class TripleBalanceTD(BaseTrainData):
...
         super(TripleBalanceTD, self).__init__(balance_kwargs)
-        self.balance_kwargs['pref_epochs'] = int(fit_kwargs.get('epochs', 1))
-        self.balance_kwargs['fit_kwargs'] = fit_kwargs
         self.balance_kwargs['query_kwargs'] = query_kwargs
-
-    def func_kwargs(self):
-        return triple_balance, self.balance_kwargs
+        self.function = triple_balance
 
     def default_kwargs(self):
```
J535D165: staticmethod

asreview/hdf5_logging.py
qubixes: From what I found, I can put it in the base class, and the HDF5 class will inherit its docstring, right?

asreview/hdf5_logging.py
qubixes: I thought HFD5Logger was a bit ugly, but that's okay I suppose.

asreview/hdf5_logging.py
qubixes: No it is used by the review class to return to a proper state from a log file.

asreview/logging.py
qubixes: Are you sure about this? To me it doesn't make much sense for the DictLogger to be a default choice, or at least I feel like that is not really appropriate in this case.

asreview/json_logging.py
qubixes: I don't think so, it can be useful to have as a logger that doesn't save anything.

asreview/json_logging.py
qubixes: Not yet I think? If one adds options to the HDF5 one that are not applicable, then it would start to break.

asreview/logging/utils.py
```diff
@@ -3,18 +3,18 @@
 
 def get_logger_class(fp):
     "Get logging class from file extension."
```
J535D165: docstring

asreview/logging/utils.py
```diff
@@ -3,18 +3,18 @@
...
-    from asreview.hdf5_logging import HDF5_Logger
-    from asreview.json_logging import JSON_Logger
-    from asreview.json_logging import Dict_Logger
+    from asreview.logging.hdf5 import HDF5Logger
```
J535D165: Please move them out of the function (better performance).

asreview/logging/utils.py
```diff
@@ -3,18 +3,18 @@
...
+    from asreview.logging.dict import DictLogger
 
     if fp is None:
-        return Dict_Logger
+        return DictLogger
 
     log_ext = os.path.splitext(fp)[1]
     if log_ext in ['.h5', '.hdf5', '.he5']:
-        Logger = HDF5_Logger
+        Logger = HDF5Logger
```
J535D165: `Logger` should be lowercase (and this is not the logger defined below isn't it?).

asreview/logging/utils.py
```diff
@@ -3,18 +3,18 @@
 
 def get_logger_class(fp):
```
J535D165: I think we can drop the class `Logger`. The functionality can be combined with this function. 

```
logger = get_logger_class(fp).from_file(fp)
```

This might require some work because of the class attribute `Logger.logger`. But this makes the entire API consistent.

qubixes: Well, except it wouldn't work properly as a context manager or would it? Maybe it makes sense to make get_logger_class private?

qubixes: It does look kind of ugly with two fp's, no?

J535D165: I had a more detailed look at what is going on. Hopefully, I can clarify it a bit. 

I think there should be a function `get_logger(fp)` that initializes `_get_logger_class(fp)`. This function outputs an initiated object that inherits from `BaseLogger`. The class `Logger` can get removed. Context managing is done in `BaseLogger`. 

Throughout the code library, we can replace  

``` python
with Logger.from_file(log_file) as logger:
     if not logger.is_empty():
```
 with
``` python
with get_logger(log_file) as logger:
     if not logger.is_empty():
```

This syntax is in line with other Python libraries. Please have a look at `get_logger` in tensorflow and `open` in base. 

The last one is a good example of the prefered syntax. One can call `with open(fp) as f: `. This returns an `_io.TextIOWrapper` object named `f`. What we do: We call something like `_io.TextIOWrapper.read(fp)` instead of using our util `get_logger`. 

And while we're at it, think about renaming `get_logger()` into `open_logger()`. Completely in line with `open()`

asreview/logging/base.py
```diff
@@ -0,0 +1,225 @@
...
+        self.restore(log_fp)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *_, **__):
+        self.close()
+
+    def __str__(self):
+        print(self.to_dict())
```
J535D165: This should be a return statement.

asreview/analysis/readers.py
```diff
@@ -0,0 +1,21 @@
+import os
+
+from asreview.logging.utils import Logger
+
+
+def loggers_from_dir(data_dir, prefix="result"):
```
J535D165: Move this function to `logging.utils`. Pretty useful function. Please extend docstring and sphinx reference.

asreview/analysis/analysis.py
```diff
@@ -0,0 +1,307 @@
...
+            y_coef = 1.0
+
+        for i in range(len(norm_yr)):
+            if norm_yr[i] >= val - 1e-6:
+                return (norm_yr[i] - norm_xr[i],
+                        (x_return[i], x_return[i]),
+                        (x_return[i]*y_coef, y_result[i]))
+        return (None, None, None)
+
+    def RRF(self, val=10, x_format="percentage", **kwargs):
```
J535D165: Lowercase. Can you make those names more explicit?

setup.py
```diff
@@ -42,12 +42,16 @@
...
+        'dill',
+        'PyInquirer',
+        'fuzzywuzzy',
+        'h5py',
     ],
 
     extras_require={
         'dev': ['check-manifest'],
         'test': ['coverage'],
+        'faster Levenshtein': ['python-Levenshtein'],
```
J535D165: 'faster Levenshtein' -> 'performance'?

J535D165:
The gif is nice. Please add the config file and the raw data to the folder as well. @GerbrichFerdinands can use those to make the other gifs :)

J535D165:
And please rename the gif into something less fancy. 

I think we need gif like `asreview_select_papers.gif`, `asreview_installation.gif`, `asreview_classification.gif`, `asreview_export.gif`. But let's discuss this with @GerbrichFerdinands.

J535D165:
Btw, I think we are close to merging.

asreview/logging/utils.py
qubixes: I can't; I get import errors when I do that.

asreview/analysis/analysis.py
qubixes: I updated the docstring of the rrf, I think using the abbreviations is okay, since scikit-learn does it in a similar way?

qubixes:
@J535D165 Thanks for the in depth review!

I resolved the Issues that were changed to remove some clutter from the review.

I was thinking that the gif would be a placeholder one, but I made a cleaner version with the config file so that it is at least somewhat presentable. I also updated the 10 minutes to reflect the new version.

Let me know when you're ready to merge, then I'll squash the review commits into one.

asreview/logging/utils.py
qubixes: I'm still not exactly sure what you meant, but check the new implementation, I think that should work fine, right?

J535D165:
Great work @qubixes!!

PR #129, opened 2019-11-15
Ris fix

qubixes:
- Fix being able to read RIS files with UTF-8 BOM characters.
- Add support for excel (.xlsx) files.
- Add support for PubMed XML files.
- More test sets.
- Throw error if the reader can't find titles or abstracts.


J535D165:
Thanks for the PR and welcome back ;)

Can you add a little test on this? One folder with all the exports from the different publishers/tools would be perfect. 

Please remove the debug print statement.

PR #133, opened 2019-11-19
Require Python version 3.6 or higher.

### LINKED ISSUE #131: Add minimal Python version to setup.py
We need something like 

```
import sys
if sys.version_info < (3,6):
    sys.exit('ASReview requires Python 3.6 or higher')
```
or 

```
setup([...],
      python_requires='>3.5',
      [...]
```

https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

@qubixes What do you think?

qubixes:
Don't commit to 4.0 yet.

Resolves #131

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
     ],
     keywords='systematic review',
     packages=find_packages(exclude=['contrib', 'docs', 'tests']),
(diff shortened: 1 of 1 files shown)


J535D165:
Sad that you don't want to support Py4 yet. Thanks!

PR #146, opened 2019-11-29
Several fixes

### LINKED ISSUE #136: At 'Stop', ASReview doesn't offer to export results.
I feel like the 'Export' and 'Stop' options aren't in optimal form.

When I want to stop reviewing, I expected the software to offer me the option to export. However, that is not the case:

```
----------------------------------


| 0.07% read | 2 since last inclusion | 25.00% included | total papers: 4/5782 |

? Include or Exclude?  Stop
? Are you sure you want to stop?  Yes

Closing down the automated systematic review.
```

If I want to export my results & quit, I first have to select the ` Export` button, and then after exporting I have to select ` Stop`. This might be a bit inconvenient.

### LINKED ISSUE #137: ASReview regards empty abstracts as relevant.
In my .csv export file, the most relevant (unreviewed) abstracts are the ones that are empty (see screenshot below). 

To me this seems like ASReview thinks less text = better? 
However, the papers I get presented with by ASReview are  (in general) not the empty ones.
Maybe something goes wrong in the ` Export` option.

image

qubixes:
Features:

- Remake menu structure for oracle mode.
-- You can now go back and add more papers manually.
-- Export option has been moved to the main menu.

Bugfixes:

- Fix getting loggers from a directory.
- Fix simulation modus with n_prior_inc/n_prior_exc
- Fix export in wrong order.
- Fix export fails to add labels.
- Various fixes for Analysis class.

Resolves #136 
Resolves #137


asreview/review/oracle.py
```diff
@@ -154,7 +154,10 @@ def priors_from_cli(self):
 
     def _prior_knowledge(self):
         """Create prior knowledge from arguments."""
-
+        if self.prior_included is None:
+            self.prior_included = []
```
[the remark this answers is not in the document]
qubixes: Okay, but that's going take more than a simple fix.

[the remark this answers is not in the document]
J535D165: Oke :( It's appreciated.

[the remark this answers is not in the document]
qubixes: You can check the oracle mode now, the simulation mode needs to be fixed still.

asreview/review/simulate.py
```diff
@@ -32,23 +32,24 @@ def __init__(self,
         super(ReviewSimulate, self).__init__(
             X, y, *args, **kwargs)
 
-    def _prior_knowledge(self):
+    def _prior_knowledge(self, logger):
```
J535D165: Can you clarify this one. `logger` doesn't do anything here. Why can't `classify()` access the logger?

qubixes: There is no logger in the Review class anymore, just the filename. We don't want to open/close a file every time we classify a sample.

J535D165: So you pass an open logger to classify? Why not store the open logger in the object (self)?

qubixes: Because it could lead to a bad logger? I.e. you open the file, do reviews, close the file, then access the closed file. Of course you can safeguard against that, etc. But I think that would be about (or more) painful as the current solution?

PR #149, opened 2020-01-03
Unsupervised

qubixes:
### Important Changes

- Due to significant API changes, the log file versions have been updated. As a result, log files created with older version of ASReview will not be able to be read with the new version. Keep using the old version with these old log files (for reproducibility purposes, this is generally a good idea).

- Different options for installing the package are now available. In an effort to keep the number of dependencies in check, the dependencies of some models are optional. In order to use these models, it is necessary to install these packages manually (an error will be shown giving the name of the missing package). You can also use `pip install asreview[all]` to install all optional dependencies automatically.

### New Features
- New Model: `nn-2-layer`
  - Dense Neural Network consisting of two layers. Seems to work well with the new doc2vec feature extraction.

- New Model: `rf`
  - Random Forest model (sklearn). 

- New Model: `logistic`
  - Logistic regression model.

- New Balance strategy: `double`
  - This is the same strategy as the `triple` balance strategy, except there is 

- New Query strategy: `cluster`
  - This query strategy uses K-Means clustering to divide the papers in different clusters. It then randomly selects one of these clusters and finds the one with the highest probability in that cluster.

- New Query strategy: `mixed`
  - This is a new 'class' of query strategies, where query strategies can be mixed. Previously only `rand_max` was implemented, but any two query strategies can be combined. 

- New Feature Extraction: `doc2vec`
  - Uses the doc2vec model from the `gensim` package.

- New Feature Extraction: `sbert`
  - Uses the Sentence BERT model with a pretrained (provided by sbert) dataset. This is probably not ideal, and as such I haven't had much success with it.

- New Feature Extraction: `embedding-idf`
  - Uses the average of word embeddings weighted by inverse document frequency.

### API Changes

- Create abstract 'super' model above all types of models.
- Move feature extraction out of the models. This means that one can now use different feature extractors with the same model, although some restrictions apply.
- Remove ModAL from the active learning process.
  - We were not using modAL all that intensively anymore. The main reason for removal is that modAL uses a system that requires functions/arguments to be passed around. Now we're using classes, which improves the readability and maintainability.
- Align all types of models (train, query, balance, feature extraction) with a similar class structure.
- Improve and align the hyper parameter optimization of the different types of models.
- Remove the `get_data` member function from the `ASreviewData` class. It was not a very useful structure that was often used to get one piece of data and throw away the other two. As a replacement, use `as_data.texts` to get the texts (title + abstract), and `as_data.labels` to get the labels.
- Lots of renamed classes. It is generally advised to search with a string.
- The query strategy `rand` has been renamed to `random`.

### Misc

- A lot of documentation was added/updated.
- New and improved unit tests. Query models are added to unit tests and tested for 'cheating'. Feature extraction received their own tests.


asreview/analysis/analysis.py
```diff
@@ -12,20 +12,19 @@
...
-'''
-Analysis of log files.
-'''
-
 import itertools
+import json
 import os
 import numpy as np
 from scipy import stats
+from sklearn.cluster import KMeans
```
J535D165: Import order and grouping

asreview/balance_strategies/base.py
```diff
@@ -12,35 +12,32 @@
...
-
-    @staticmethod
-    @abstractmethod
-    def function():
-        raise NotImplementedError
-
-    @staticmethod
     @abstractmethod
-    def description():
+    def sample(self, X, y, train_idx, shared):
```
J535D165: shared -> make use of kwargs?

asreview/balance_strategies/double.py
```diff
@@ -0,0 +1,135 @@
...
+    of samples in the training data.
+    """
+    name = "double-balance"
+
+    def __init__(self, a=2.155, alpha=0.94, b=0.789, beta=1.0):
+        """Initialize the double balance strategy.
+
+        Arguments
+        ---------
+        a: double
```
J535D165: Double is not a valid Python type. Do you mean float?

asreview/balance_strategies/triple.py
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(self, a=2.155, alpha=0.94, b=0.789, beta=1.0, c=0.835,
+                 gamma=2.0, shuffle=True):
+        """Initialize the triple balance strategy.
+
+        Arguments
+        ---------
+        a: double
+            Governs the weight of the 1's. Higher values mean linearly more 1's
+            in your training sample.
+        alpha: double
```
J535D165: same

setup.py
```diff
@@ -71,9 +71,12 @@ def get_long_description():
         'xlrd>=1.0.0',
     ],
     extras_require={
+        'sbert': ['sentence_transformers'],
+        'doc2vec': ['gensim'],
         'dev': ['check-manifest'],
         'test': ['coverage'],
```
J535D165: pytest

setup.py
```diff
@@ -71,9 +71,12 @@ def get_long_description():
...
         'performance': ['python-Levenshtein'],
+        'all': ['gensim', 'python-Levenshtein', 'sentence_transformers'],
```
J535D165: Can you make this into something more robust? 'all' is not incomplete, isn't it?

Maybe define some global:

```
DEPS_TEST = ['pytest', 'coverage']
DEPS_DEV = ['check-manifest']

extras_require={
   ...,
   'all': DEPS_TEST + DEPS_DEV + ....
```

J535D165:
Thanks a lot for the changelog.

asreview/balance_strategies/base.py
qubixes: I'm not sure, but I *think* I prefer it like this, since it ensures that the shared dictionary is passed on (in case the balance model needs it).

J535D165:
Thanks a lot!

PR #154, opened 2020-01-22
Make modes into dynamic entry points

qubixes:


qubixes:
Yes, I tried to build the docs locally and that does work. Maybe it is bugged because a new package was added? I hope you would know. :D

PR #155, opened 2020-01-27
Fix documentation + add documentation on extensions

qubixes:
Also improve some of the analytics for plotting.


docs/source/index.rst
```diff
@@ -33,6 +33,8 @@ The source code is freely available at
 
    faq
 
+   extensions
```
J535D165: Shall we make a second toc tree with caption Development? With the API ref and building extensions? Basic usage of extensions can be described in the other toc tree(?)

qubixes: I agree, see the update!

PR #157, opened 2020-01-29
Rename repository and organisation

J535D165:
https://github.com/msdslab/automated-systematic-review will become https://github.com/asreview/asreview.

Other repos follow soon.


J535D165:
Thanks for the detailed look. It made me find more references. I think it is better now.

PR #158, opened 2020-01-30
Allow for better handling of prefixes

qubixes:


asreview/logging/utils.py
```diff
@@ -28,8 +28,10 @@ def _get_logger_class(fp):
     log_ext = os.path.splitext(fp)[1]
```
J535D165: Note. Replace filepath handling by`pathlib` in this module.

J535D165:
Thanks!

PR #165, opened 2020-02-13
Remove Deprecation warning

J535D165:


qubixes:
We should remove the entry point as well then?

J535D165:
Good point, pushed a commit.

PR #176, opened 2020-03-11
Rename 'logging' module into 'state'

qubixes:


J535D165:
Many many thanks! @maartenhoog

PR #181, opened 2020-03-17
Keyword col

qubixes:


asreview/datasets.py
```diff
@@ -0,0 +1,142 @@
...
+    link = "https://zenodo.org/record/1162952#.XIVBE_ZFyVR"
+    authors = ["Tracy Hall", "Sarah Beecham", "David Bowes", "David Gray",
+               "Steve Counsell"]
+    year = 2012
+    license = "CC-BY Attribution 4.0 International"
+    topic = "Software Fault Prediction"
+
+
+class ExampleDataSetGroup():
+    data_sets = [
```
J535D165: self? Not sure. Especially useful to make an isolated ExampleDataSetGroup?

Do we need an append method as well?

qubixes: I'm not quite sure what you mean. I added an append method, though I don't think it needs to be used really. Maybe with the update it makes more sense?

J535D165: No problem, if this works with an extension that adds datasets, I'm fine.

J535D165:
Closes #182

PR #186, opened 2020-03-22
Browser interface for ASReview 

J535D165:


J535D165:
Developers can check https://github.com/asreview/asreview/blob/browser-interface/DEVELOPMENT.md for instructions

PR #187, opened 2020-03-22
use rapidfuzz instead of fuzzywuzzy

maxbachmann:
FuzzyWuzzy is GPLv2 licensed which would force you to licence the whole project under GPLv2. I had the same problem on one of my projects and so I wrote [rapidfuzz](https://github.com/rhasspy/rapidfuzz) which is implementing the same algorithm but is based on a version of fuzzywuzzy that was MIT Licensed and is therefor MIT Licensed aswell, so it can be used in here without forcing a License change. As a nice bonus it is fully implemented in C++ and comes with a few Algorithmic improvements making it between 5 and 100 times faster than FuzzyWuzzy.

### DIFF
--- a/asreview/data.py
+++ b/asreview/data.py
@@ -32,7 +32,7 @@ from asreview.utils import is_url
 
 with warnings.catch_warnings():
     warnings.filterwarnings("ignore")
-    from fuzzywuzzy import fuzz
+    from rapidfuzz import fuzz
 
 
 def get_fuzzy_scores(keywords, str_list):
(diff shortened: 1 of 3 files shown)


qubixes:
Thanks for the pull request! I wasn't aware that FuzzyWuzzy was GPLv2 licensed, but in that case you're right that would be a problem.

We're in the middle of trying to get out a big release, but we'll look into it as soon as we can. @J535D165

maxbachmann:
sure take your time :+1:

J535D165:
@maxbachmann Would be nice to have python wheels for all platforms. Otherwise, this might become a struggle for many of our users.

maxbachmann:
@J535D165 there are already python wheels for it https://pypi.org/project/rapidfuzz/#files or which wheels do you miss?

PR #191, opened 2020-03-27
Faster and better fuzzy matching

qubixes:


J535D165:
Looks great. It would be nice to have the functionality in a standalone python file. Are there some unit tests available?

Looking forward to merging!

qubixes:
@J535D165 Yes, unit tests were already available before, and I haven't modified them.

For now, the functionality is not that useful outside of the ASReviewData class, perhaps split them in another iteration?

J535D165:
> Yes, unit tests were already available before, and I haven't modified them.

Perfect. Merge

PR #206, opened 2020-03-31
Python exec detection + new API call

qubixes:
Whether to decide if the model has failed or not should/could be handled user side.

### DIFF
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -5,9 +5,10 @@ import json
 import logging
 import subprocess
 import shlex
+import sqlite3
 from urllib.request import urlretrieve
 import urllib.parse
-
+from time import sleep
 
 from flask.json import jsonify
 from flask_cors import CORS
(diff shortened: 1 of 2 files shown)


asreview/webapp/api.py
```diff
@@ -366,6 +367,35 @@ def api_init_model_ready(project_id):  # noqa: F401
...
+            lock_entry = db.execute(
+                'SELECT * FROM locks WHERE name = ?',
+                ("training",)).fetchone()
+            if lock_entry is None:
+                response = jsonify({"status": 0})
+            else:
+                response = jsonify({"status": 1})
+            break
+        except sqlite3.OperationalError as e:
+            print(f"Encountering operational error {e}")
```
J535D165: It would be nice to raise an exception such that we catch it and return an informative API request.

asreview/webapp/utils/project.py
```diff
@@ -221,7 +222,8 @@ def label_instance(project_id, paper_i, label, retrain_model=True):
 
     if retrain_model:
         # Update the model (if it isn't busy).
-        run_command = f"python -m asreview web_run_model '{project_id}'"
+        run_command = (f"{sys.executable} -m asreview web_run_model "
```
J535D165: Great find, `sys.executable`!

asreview/webapp/api.py
qubixes: I'm not sure what you mean?  It isn't an actual error, just means that the lock file is in use. Right now the API call should timeout when it can't somehow open de database, which is kind of okay-ish?

J535D165: Can't we use https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError when calling the subprocess?

qubixes: I don't think so, since the process asynchronous, but I could be mistaken. Does that exception get raised in the case that the model training fails?

PR #220, opened 2020-04-09
Remove all lists from the dataframe.

qubixes:


asreview/entry_points/simulate.py
```diff
@@ -70,4 +70,11 @@ def _simulate_parser(prog="simulate", description=DESCRIPTION_SIMULATE):
              "Only used when --prior_excluded is not given. "
              f"Default {DEFAULT_N_PRIOR_EXCLUDED}")
 
+    parser.add_argument(
```
J535D165, line 73: What's this? Unrelated?

qubixes: Yes, sorry, sometimes I need things and I program them in the wrong pull request. I think it should also be added to master though.

qubixes: It's a simulation only feature that I sometimes need if I want to start with specific papers. Not recommended usage, but useful for me sometimes..

PR #223, opened 2020-04-10
Add option to export Excel file

J535D165:
This PR adds an option to export results to EXCEL (xlsx). Defaults to Excel.

localhost_3000_(iPad)
localhost_3000_(iPad) (1)

### DIFF
--- a/asreview/data.py
+++ b/asreview/data.py
@@ -503,7 +503,7 @@ class ASReviewData():
     def to_file(self, fp, labels=None, ranking=None):
         """Export data object to file.
 
-        Both RIS and CSV are supported file formats at the moment.
+        RIS, CSV and Excel are supported file formats at the moment.
 
         Arguments
         ---------
@@ -518,6 +518,8 @@ class ASReviewData():
             self.to_csv(fp, labels=labels, ranking=ranking)
         elif Path(fp).suffix in [".ris", ".RIS"]:
             self.to_ris(fp, labels=labels, ranking=ranking)
+        elif Path(fp).suffix in [".xlsx", ".XLSX"]:
+            self.to_excel(fp, labels=labels, ranking=ranking)
         else:
             raise BadFileFormatError(
                 f"Unknown file extension: {Path(fp).suffix}.\n"
@@ -531,9 +533,9 @@ class ASReviewData():
         labels: list, np.ndarray
             Current labels will be overwritten by these labels
             (including unlabelled). No effect if labels is None.
-        df_order: list
+        ranking: list
             Reorder the dataframe according to these (internal) indices.
-            Deafault ordering if df_order is None.
+            Default ordering if ranking is None.
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


J535D165:
@qubixes, can you do a review of the Python code. It required more lines of code than I expected :|

asreview/data.py
```diff
@@ -556,8 +556,48 @@ def to_dataframe(self, labels=None, ranking=None):
...
+        """Export to csv.
+
+        Arguments
+        ---------
+        fp: str, NoneType
+            Filepath or None for buffer.
+        labels: list, np.ndarray
+            Current labels will be overwritten by these labels
+            (including unlabelled). No effect if labels is None.
+        df_order: list
```
qubixes, line 568: df_order -> ranking

asreview/data.py
```diff
@@ -556,8 +556,48 @@ def to_dataframe(self, labels=None, ranking=None):
...
+            Reorder the dataframe according to these (internal) indices.
+            Deafault ordering if df_order is None.
```
qubixes, line 570: Deafault -> Default, df_order -> ranking

asreview/data.py
```diff
@@ -556,8 +556,48 @@ def to_dataframe(self, labels=None, ranking=None):
...
```
qubixes, line 592: Deafault -> default, df_order -> ranking

asreview/webapp/utils/paths.py
```diff
@@ -55,6 +55,18 @@ def get_project_file_path(project_id):
     return Path(get_project_path(project_id), "project.json")
 
 
+def get_tmp_path(project_id):
```
qubixes, line 58: Perhaps we can make use of tempfile? https://docs.python.org/3.8/library/tempfile.html (tempfile.mkstemp)

asreview/webapp/src/Components/ExportDialog.js
```diff
@@ -65,17 +92,32 @@ const ExportDialog = (props) => {
         <DialogTitle id="scroll-dialog-title">Export results {props.project_id}</DialogTitle>
           <DialogContent dividers={true}>
             <Typography>
-              Download the result or your review
+              Download the result or your review to a Excel or CSV file.
```
qubixes, line 95: a Excel -> an Excel

Maybe "Export your review (Excel or CSV)."?

asreview/webapp/utils/paths.py
J535D165: Good suggestion, I will implement this. Maybe this is a good way to export the CSV as well. Easier to trace than exporting a buffer.

asreview/webapp/src/Components/ExportDialog.js
J535D165: Thanks

J535D165:
Thanks, made some changes. 

The config file suggestion is nice. Especially if there are also human-readable labels available to use in the front-end. Let's put this on our todo list. I added the option to `to_file`.

PR #218, opened 2020-04-09
Model selection in ASReview GUI

rvanelk:
start systematic review view

PR #233, opened 2020-04-15
Model selection in Step 5

J535D165:
This PR is a follow-up of PR #218 by @rvanelk. It implements the backend for the model selection.


asreview/webapp/utils/project.py
```diff
@@ -60,13 +57,6 @@ def init_project(
                 'authors': project_authors
             }, fp)
 
-        asr_kwargs = deepcopy(app.config['asr_kwargs'])  # remove config
-        with open(get_kwargs_path(project_id), "w") as fp:
-            json.dump(asr_kwargs, fp)
-
-        # make a copy of the arguments to the state file
-        asr_kwargs['state_file'] = str(get_state_path(project_id))
```
J535D165, line 68: @qubixes I'm a bit clueless on this line. Is this still required?

qubixes: It seems not, so it should be safe to delete.

asreview/webapp/api.py
```diff
@@ -392,6 +394,23 @@ def api_random_prior_papers(project_id):  # noqa: F401
...
+    ml_model = request.form.get("machine_learning_model", None)
+    if ml_model and not ml_model == "cli":
+        asr_kwargs["model"] = ml_model
+
+    # write the kwargs to a file
+    with open(get_kwargs_path(project_id), "w") as fp:
+        json.dump(asr_kwargs, fp)
+
+    # make a copy of the arguments to the state file
+    # asr_kwargs['state_file'] = str(get_state_path(project_id))  # TODO@{Raoul} obsolete?
```
qubixes, line 412: Yes, can be removed.

qubixes:
Seems alright to me. Just one question; I don't see any mention of feature extraction selection. Not all combinations of feature extraction + model work well (i.e. they might use >10GB of memory), is something done to take care of that?

J535D165:
At the moment, we do have 4 models implemented in the front-end. All with the default feature extraction method. I think the current combinations are fine, isn't it?

qubixes:
Which 4? All except the Tensorflow models? That should work fine.

asreview/webapp/api.py
```diff
@@ -392,6 +394,23 @@ def api_random_prior_papers(project_id):  # noqa: F401
...
     """Start training the model
     """
+
+    # get the CLI arguments
+    asr_kwargs = deepcopy(app.config['asr_kwargs'])
+
+    # add the machine learning model to the kwargs
+    # TODO@{Jonathan} validate model choice on server side
+    ml_model = request.form.get("machine_learning_model", None)
+    if ml_model and not ml_model == "cli":
```
qubixes, line 404: if ml_model != "cli"?

Isn't that equivalent/more concise?

asreview/webapp/utils/project.py
J535D165: Done

asreview/webapp/api.py
J535D165: Thanks. I had a different intention with this part of the code. But I will propose an implementation later on. (It's not possible to experiment with new models now. So I wanted to a "CLI" model option in the front-end.)

J535D165:
Thanks @rvanelk @qubixes @Rensvandeschoot for the work/feedback.

PR #239, opened 2020-04-15
Improve the HDF5 performance of the log file.

qubixes:
JSON + older HDF5 should be deprecated at some point.


J535D165:
Nice. Can you write a deprecation plan? And can you define the schema?

qubixes:
@J535D165 What kind of schema? Do we have it for the old one? I kind of remember something passing by...

I'm not sure about a deprecation plan. I want to improve some handling of the log-file, most notably remove "query_i" as a concept. (It took me ages to get the new format to behave like the legacy one..) So, as soon as possible would be my personal preference. But of course I understand I would break people's reviews, so maybe a read-only period? All that kind of stuff is so much work though >_>.

J535D165:
Closed in favor of https://github.com/asreview/asreview/tree/state-file-v3.

PR #240, opened 2020-04-15
Provide user feedback when model training fails

J535D165:
Give an informative error message to the user when the model training in step 5 fails.


J535D165:
Parts of #206 are now integrated in this PR. Not everything yet.

PR #243, opened 2020-04-17
Text size setting

rvanelk:
Text size setting for ArticlePanel.

test-size-normal
test-size-large


J535D165:
Merged after some modifications in commit 934a5f2.

PR #250, opened 2020-04-17
Circumvent Github Actions disk space issues

J535D165:
Get rid of disk space problems due to large dependencies. This PR splits the javascript compilation test from the Python unit tests.


J535D165:
@qubixes Please have a look. This seems to work but keep this in mind when adding an additional dependency layer.

PR #251, opened 2020-04-17
Add versioned datasets

qubixes:
Also add a dataset manager for an easier interface to the dataset.

TODO:

- Update the GUI to match the new API results.
- Test if the GUI works properly with the new version of the covid19 plugin.
- Test if the GUI works properly with the old version of the covid19 plugin.


asreview/datasets.py
```diff
@@ -41,9 +55,154 @@ def to_dict(self):
...
+
+    def list(self, latest_only=True):
+        return [self]
+
+
+class BaseVersionedDataSet():
+    def __init__(self, base_dataset_id, datasets=None, url=None):
+        if datasets is None and url is not None:
+            index_file = url+"/index.json"
+            with open(index_file, "r") as f:
```
J535D165, line 71: Not possible to open a url with `open`. This is required. The plugin should be a URL, not a relative filepath.

asreview/datasets.py
```diff
@@ -41,9 +55,154 @@ def to_dict(self):
...
+        if data_name in self.aliases:
+            return self
+        return None
+
+    def list(self, latest_only=True):
+        return [self]
+
+
+class BaseVersionedDataSet():
+    def __init__(self, base_dataset_id, datasets=None, url=None):
```
J535D165, line 68: I'm wondering why the arguments of BaseVersionedDataSet and BaseDataSet differ so much. They are similar isn't it

J535D165:
Made a change in the JS to show the first dataset in a group. But I'm running into problems with the latest covid-19 plugin.

asreview/datasets.py
qubixes: Yes, but one is a container for the other!

PR #254, opened 2020-04-18
Fix sys.executable not found

J535D165:
Windows users (with Anaconda installation?) have problems with finding the python executable. This is use to train a new model. This PR tries to find the executable with `sys.executable`. If nothing was found, the application falls back on 'python'.

### DIFF
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -5,7 +5,6 @@ import re
 import shlex
 import shutil
 import subprocess
-import sys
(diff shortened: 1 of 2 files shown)


Rensvandeschoot:
I also tried without anaconda and only python installed, and had the same issue, sorry  :-(

J535D165:
Did you test the latest pre-release? It can be a Windows specific issue and not related to anaconda or Python for Windows. Please let me know.

PR #259, opened 2020-04-22
Add option to set seed for initialisation

qubixes:
Also fixes nasty bug that makes ASReview ignore the number of inclusions/exclusions.


asreview/entry_points/simulate.py
```diff
@@ -77,4 +77,10 @@ def _simulate_parser(prog="simulate", description=DESCRIPTION_SIMULATE):
         type=int,
         help="Prior indices by id."
     )
+    parser.add_argument(
+        '--init_seed',
+        default=None,
+        type=int,
+        help="Seed for initializing the priors."
```
J535D165, line 84: Prior indices?

J535D165:
Thanks!

PR #260, opened 2020-04-23
[Feature] Set seeds of models.

qubixes:


J535D165:
Some other thoughts: 

Would it be relevant to save the seed in the state file? 
Is seeding relevant for the front-end as well?

J535D165:
> You could theoretically put them in the state files, but then resuming from said state file would become a bit messy in the sense that it would no longer be correct. Let's let it be for now? Or I could put it in and not worry about potential mishaps.

> I wouldn't worry about putting it in the front-end as long as we don't have a simulation mode there.

Thanks for your thoughts on this. I do agree. Let's skip these things for now.

tests/test_init.py
```diff
@@ -42,3 +42,17 @@ def test_no_seed():
...
             return
     raise ValueError(f"Error getting all priors in {n_test_max} iterations.")
+
+
+def test_model_seed():
+    n_test = 4
+    seed = 192874123
+    last_train_idx = None
+    for _ in range(n_test):
+        reviewer = get_reviewer(
```
J535D165, line 52: Is there anything non-deterministic in this model and query strategy?

qubixes: Yes, the balance strategy (probably). Anyway, good point, I changed it to make it random for sure.

PR #263, opened 2020-05-06
Fix one-off bugs in WSS and TTD

qubixes:


J535D165:
Thanks for fixing and @GerbrichFerdinands for figuring out.

PR #269, opened 2020-06-08
Fix appending datasets where only one has labels

qubixes:


asreview/data.py
```diff
@@ -217,6 +226,8 @@ def append(self, as_data):
             else:
                 self.column_spec[data_type] = col
 
+#         if self.
```
J535D165, line 229: Remove

J535D165:
Thanks, I will merge after the comment is fixed.

qubixes:
Thanks, I have removed the comment.

PR #276, opened 2020-06-16
added Dockerfile and instructions

behrica:
I made a minimal Dockerfile which stars the ASReview app and instructions how to use it.

Now you should be able to build the image, push it to dockerhub and replace xxxxx/asreview in the instrucstion with the real name.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,3 @@
+FROM python:3.6
+RUN pip install asreview
+ENTRYPOINT ["asreview","oracle","--ip","0.0.0.0"]
--- a/README.md
+++ b/README.md
@@ -37,11 +37,39 @@ Upgrade ASReview with the following command:
 pip install --upgrade asreview
 ```
 
+
 ## ASReview app
 
 The ASReview team developed a user-friendly user interface to replace the old command line interface. The new interface is still under development but is already available for testing and training purposes.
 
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM python:3.6
+RUN pip install asreview
```
J535D165, line 2: It might be interesting to add our plugin `asreview-covid19` there as well.

J535D165:
Hi @behrica, I made a PR on your branch. https://github.com/openefsa/asreview/pull/1 If you merge, I'm happy to merge this PR.

Dockerfile
behrica: Yes, I thought about it.

Same for other plugins.

but then I saw that they all have different versions,
so it would be hard to make a versioned Docker tag.

But maybe it makes sense to have only a single Docker image,
which includes all plugins in the latest released version.

I will have a look at that.

behrica: I think the best is to offer 2 Dockerfile and images.

One for the GUI app, where the web app is the entrypoint, like the current one.
Adding there covid19 plugin is fine.

A second Dockerfile  + image which is ment to be used from commandlie.
This could then have all plugins, but the entrypoint should be either /bin/bash or "asreview"

Mabe names could be

asreview-app
asreview-cli

I make an other issue for this

J535D165:
Closed in favor of #278

Issue #277, opened 2020-06-17
Make 2 docker files for app and cli


behrica:
see discssuion in #276 for motivation

PR #278, opened 2020-06-17
#277 two dockers

behrica:


J535D165:
This is very interesting. Thanks for this new PR. I think I prefer this one. Very useful for simulation studies.

qubixes:
@behrica Thanks for the pull request.

I think asreview-hyperopt should be removed from the installed packages. It requires some implementation of MPI (2+?) and I have tried to get OpenMPI working within a docker container, but I didn't get it to work.

If you really want the hyperparameter optimization to work, the way to do it would be to create some kind of queuing server (RQ/Celery) for jobs and also implement this in asreview-hyperopt. Wouldn't be impossible/super hard, but still quite a bit of work.

Some part of the simulation chain resides in asreview-simulation, but it shouldn't be included for the same reason (depends on MPI). In the end those scripts are mostly just convenience and nothing a user can't script themselves, so I don't think it is a problem not to include these scripts.

I forgot how entry points in docker work, so I leave that to @J535D165..

TLDR; Thanks for the PR, probably should remove `asreview-hyperopt` from the docker file.

qubixes:
Actually, never mind. I forgot I removed the MPI requirement of the asreview-hyperopt package. It should work. It will just not have multi-core/processor capabilities.

behrica:
Is the PR fine now ?
You suggested remove "simulation" as well,
but I would say that running simulations on own past reviews is  a very frequent used way to help deciding,
if ASReview can work for you.

That what I did at least as first thing, even before using the GUI...

qubixes:
@behrica Sorry for the confusion, yes, as far as I am concerned it is fine. I was talking about the `asreview-simulations` package, which includes scripts for running the simulations in parallel. The simulation mode that you're probably referring to is part of the asreview core package.

J535D165:
Thanks, everything is fine! I will merge this one soon and upload the versions to the Docker Hub.

J535D165:
Hi @behrica, I'm tweaking some stuff in this PR. Can you give me write access to this branch such that I can easily integrate your work?

behrica:
Done

J535D165:
Thanks a lot @behrica, hopefully my setup with DockerHub works...

J535D165:
Everything seems to work. At the next minor release, the `latest` and `release-x.x.x` will activate. For now, the `development` tag can be used.

PR #279, opened 2020-06-17
Add donate buttons in the software

J535D165:
Add donate buttonin left menu and export results. 

localhost_3000_(iPad)

localhost_3000_(iPad) (1)


MartijnUX:
I think for now the location is fine. If you google on "donate icon" in google you get a hand with a coin. So maybe that is better. A better solution to ask for a donation is on the website asreview.nl and show a Module-window after using it a while (while screening), with the message that we can't do this for free, help us...help you, donate...Yes/No.

Issue #275, opened 2020-06-16
Document better the "minimal" expected fields for CVS import files


behrica:
## Feature Request
I had a bit of an issue, to discover which are the  minimal fields requires to get a CSV file as import working.

I saw the docuemntation, but it is not sufficient.
https://asreview.readthedocs.io/en/latest/datasets.html#using-your-own-data

I tried for example, with only 3 columns:

- id
- abstrcat
- label 
and this failed with a non-informative error message

I tried for a while untill I figured out, to add empty columns for:

- title
- keywords

J535D165:
Thanks, we are working on updating the documentation of this page. We will implement your recommendations. 

Can you share a few lines of your file, because I do have some issues reproducing it? As far as I know, those 2 columns are not required.

This is the error you should see if there is neither a title nor an abstract.  

![localhost_5001_(iPad)](https://user-images.githubusercontent.com/12981139/84798235-fd06fd80-affa-11ea-8e33-93ae63b648b4.png)

behrica:
This happended during a "asreview simulate" run. 
I used first a csv file with these headers:

id | abstract | label
-- | -- | --
1 | Hydroxycinnamic acids and flavonoids in …. | 0
2 | This study used cell-free enzyme (CFE) extracts …. | 1

This gives me 

```ARNING:root:Unable to detect titles in dataset.
Labeled during review:

1 => 1
0 => 0

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/asreview/review/base.py", line 196, in __init__
    self.X = state.get_feature_matrix(as_data.hash())
  File "/usr/local/lib/python3.6/site-packages/asreview/state/dict.py", line 125, in get_feature_matrix
    my_data = self._state_dict["data_properties"][data_hash]
KeyError: 'data_properties'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/asreview", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/asreview/__main__.py", line 40, in main
    entry.load()().execute(sys.argv[2:])
  File "/usr/local/lib/python3.6/site-packages/asreview/entry_points/simulate.py", line 29, in execute
    review_simulate(path, **args_dict)
  File "/usr/local/lib/python3.6/site-packages/asreview/review/factory.py", line 239, in review_simulate
    review(dataset, *args, mode='simulate', **kwargs)
[...]

PR #280, opened 2020-06-18
Add tables with supported export formats/software for datasets

SagevdBrand:
Adding information on which export formats of citation managers, systematic review software and search engines work in ASReview. Also added a table containing information on which metadata should be present in .csv files.

PR #281, opened 2020-06-18
Fix issue where ASReview needs both title and abs

qubixes:
Also adds unit tests.


J535D165:
Thanks a lot.  This closes #275. Documentation follows in #280.

PR #284, opened 2020-06-18
Add documentation on simulation study results

GerbrichFerdinands:
I propose to add results on simulation studies to the documentation, in response to [this issue](https://github.com/asreview/systematic-review-datasets/issues/26) in the systematic-review-datasets repository. 

This PR contains a first draft. I've just included some key results of the simulation studies I carried out for my Master's thesis project. I would like to have some feedback, I would have time to process this somewhere next week.


docs/source/simulation_study_results.rst
```diff
@@ -0,0 +1,317 @@
...
+This document is currently a work in progress. The results documented
+here originate from Gerbrich Ferdinands’ master’s thesis which can be
+found `here
+<https://github.com/GerbrichFerdinands/asreview-thesis/blob/master/manuscript/manuscript/Ferdinands%2C-G---MSBBSS.pdf/>`_.
+
+
+To provide insight in how much screening effort ASReview could
+potentially save, seven ASReview models were simulated on six existing
+systematic review datasets. In short, for all six datasets ASReview
+could have saved at least 60% of screening effort (e.g. WSS95% was <
```
qubixes, line 13: This doesn't sound right. I would also add the '@', because I think that's how it's presented in literature. Correct me if I'm wrong. Either way, unless I'm sorely mistaken, higher values of WSS@95 denote better results. (So, should be > 60% I think).

docs/source/simulation_study_results.rst
```diff
@@ -0,0 +1,317 @@
...
+relevant publications after screening only 5% of relevant publications.
+
+Datasets
+--------
+
+Datasets were collected from the fields of medicine (Cohen et al. 2006;
+Appenzeller‐Herzog et al. 2019), virology (Kwok et al. 2020), software
+engineering (Yu, Kraft, and Menzies 2018), behavioural public
+administration (Nagtegaal et al. 2019) and psychology (van de Schoot et
+al. 2017), to assess generalizability of the models across research
```
qubixes, line 24: I would split it into two sentences or put this part first.

docs/source/simulation_study_results.rst
```diff
@@ -0,0 +1,317 @@
...
+contexts. Datasets are available in the `ASReview systematic review
+datasets
+repository <https://github.com/asreview/systematic-review-datasets>`__.
+
+Data were preprocessed from their original source into a test dataset,
```
qubixes, line 29: I think "test dataset" might be a little confusing as it might suggest cross validation, which is probably not what you mean.

docs/source/simulation_study_results.rst
```diff
@@ -0,0 +1,312 @@
...
+
+
+WSS and RRF tables
+~~~~~~~~~~~~~~~~~~
+
+**Table 2 - WSS95 values (mean, standard error) for all model-dataset
+combinations, and median (MAD) for all datasets**
+
++---------+---------+---------+---------+---------+---------+---------+
+|         | Nudging | PTSD    | Softwar | Ace     | Virus   | Wilson  |
```
J535D165, line 151: `e` is on a new line.

J535D165: Same for table below

docs/source/simulation_study_results.rst
```diff
@@ -0,0 +1,312 @@
...
+exhaustive as the ASReview backend implements various other
+configurations.
+
+Evaluation
+----------
+
+Model performance was assessed by two different measures, Work Saved
+over Sampling (WSS), and Relevant References Found (RRF). WSS indicates the reduction in publications needed to be screened, at a
+given level of recall (Cohen et al. 2006). Typically measured at a
+recall level of 0.95 (Cohen et al. 2006), WSS95 yields an estimate of
```
Rensvandeschoot, line 79: I don't think you need to cite 2x Cohen

docs/source/simulation_study_results.rst
```diff
@@ -0,0 +1,312 @@
...
+| RF +    | 62.6    | 97.1    | 99.2    | 80.8    | 67.3    | 75.5    |
+| D2V     | (5.47)  | (1.90)  | (0.34)  | (5.72)  | (3.19)  | (14.35) |
++---------+---------+---------+---------+---------+---------+---------+
+| LR +    | 67.5    | 98.6    | 99.0    | 81.7    | 70.6    | 90.6    |
+| D2V     | (2.59)  | (1.40)  | (0.00)  | (1.81)  | (2.21)  | (5.00)  |
++---------+---------+---------+---------+---------+---------+---------+
+| median  | 62.6    | 98.6    | 99.0    | 84.2    | 73.4    | 89.1    |
+| (MAD)   | (3.89)  | (1.60)  | (0.00)  | (3.71)  | (0.70)  | (2.70)  |
++---------+---------+---------+---------+---------+---------+---------+
+
```
Rensvandeschoot, line 210: Maybe add some conclusions?

docs/source/simulation_study_results.rst
```diff
@@ -0,0 +1,317 @@
...
+proportion of relevant publications that are found after screening 10%
+of all publications.
+
+Furthermore, model performance was visualized by plotting recall curves.
+Plotting recall as a function of the proportion of screened publications
+offers insight in model performance throughout the entire screening
+process (Cormack and Grossman 2014; Yu, Kraft, and Menzies 2018). The
+curves give information in two directions. On the one hand they display
+the number of publications that need to be screened to achieve a certain
+level of recall (1-WSS), but on the other hand they present how many
```
qubixes, line 92: I'm not sure I understand this part. What is the 1-WSS?

docs/source/simulation_study_results.rst
GerbrichFerdinands: No you are correct, my bad!

docs/source/simulation_study_results.rst
GerbrichFerdinands: I see I was a bit unclear. I mean to explain that you can read the recall curve in terms of WSS or RRF. In case of the WSS, the x-axis of the recall curves actually show the proportion of papers that have been screened instead of the proportion of papers that have not been screened, hence the 1-WSS. 

> On the one hand they display the proportion of publications that need to be screened (1-WSS) to achieve a certain level of recall.

qubixes: Correct me if I'm wrong, but I think you're trying to convey that you could flip the x- and y-axes and learn something interesting, right? (i.e. either ask which % of papers need to be reviewed, or what % of inclusions we find.)

Maybe the easiest way to convey this is to explain this is to describe it visually? Otherwise it might be a little hard for the non-expert to read.

I still don't understand the 1-WSS though. That would be the percentage of papers read + % of inclusions not found?

GerbrichFerdinands: Yep that's waht I'm trying to convey! Then how about:

> The x-axis represents the proportion of screened publications, the y-axis represents the proportion of relevant publications found. The curves give information in two directions. On the one hand, they display the proportion of relevant publications found for any point during the screening process (on the x-axis). On the other hand, they present the proportion of publications that need to be screened to achieve a certain level of recall (on the y-axis). The quicker the curve reaches towards the top of the y-axis, the better the performance. 

About the 1-WSS: correct me if I'm wrong, but I figured that WSS@95% is 60%, that means that after having seen 40% of publications you've found 95% of relevant publications. But I figure from your response that this is technically not 100% correct?

qubixes: Yes, I think that's much easier to understand.

No, as I understand WSS@95% == 60% means that after reading 35% of the publications we have found 95% of the publications (60 + 35 == 95). It is the amount work saved, instead of doing it randomly. So WSS@x for random search strategy should be 0 for any x. In your case if WSS@100% == 60% then you need to read 40%.

But honestly, it's quite confusing, so it wouldn't surprise me if I also got the numbers wrong somehow...

J535D165: We will close this discussion for now. Thanks for the discussion. We will address this in a new issue/PR.

J535D165:
Merge! Thank you so much!

PR #293, opened 2020-06-26
Import project function enhancement

terrymyc:
image

- Add import as a speed dial option to the add button.
- Refresh the projects after importing new projects.

### DIFF
--- a/asreview/webapp/src/App.js
+++ b/asreview/webapp/src/App.js
@@ -157,12 +157,21 @@ const App = () => {
       {appState['step'] === 'projects' &&
       <Projects
         handleAppState={handleAppState}
+        toggleImportProject={toggleImportProject}
       />
       }
 
       {appState['step'] === 'review-init' &&
       <PreReviewZone
         handleAppState={handleAppState}
+      />
+      }
+
+      {appState['step'] === 'review-import' &&
+      <ImportDialog
+        handleAppState={handleAppState}
+        toggleImportProject={toggleImportProject}
+        importProject={importProject}
       />
       }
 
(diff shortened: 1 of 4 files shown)


MartijnUX:
Wow, this is a creative way to make more out of the "+" Button, nicely done. The user question would be:"what do I import" A research (user) could think of importing a dataset or importing a project. To make intuitive we should know what happens after you press the (what's now called) the "import" icon. The other naming-options are: add, open, select, and many more. For a Floating button this option is really nice. 

Another thought: We also could choose for the Startpage option, when the researcher has no projects yet. If the researcher has one or more projects we show the floating button with this nice collapsed situation. Thank you!

terrymyc:
@MartijnUX Thanks for your suggestion. The name "Import" looks a bit vague. I changed it to "Add."

image


With regard to:
> We also could choose for the Startpage option, when the researcher has no projects yet.

Do you mean when the researcher has no projects, we could have a start page with maybe two buttons/options which explicitly saying "Create a new project" and "Import a project"?

J535D165:
Thanks @terrymyc! 

I think we should put 'new' on top (one starts to read from the top, 'new' is the most popular option I think).
 
The 'add' icon seems to be a 'publish' icon. What is a good icon? https://material.io/resources/icons/?icon=cloud_upload&style=baseline?

What do we do with the import button in the left menu? Import or add?

terrymyc:
@J535D165 On the left menu, it is "import" with the "publish" icon for now. Do you have any preference for the icon? There are just too many options from which it is hard to choose.

J535D165:
Not sure, 'create new folder', 'cloud upload', 'folder'? But I'm fine with 'publish' as well. Looks quite intuitive. 

So maybe switch the buttons and then merge :)

terrymyc:
@J535D165 "CreateNewFolder" looks suitable as we indeed add a new folder with a project. I use that "Publish" icon for "Import" on the left menu just because it looks symmetric with the "GetApp" icon for "Export"...

image

J535D165:
Thanks a lot!

PR #298, opened 2020-06-30
Save dark mode and text size preferences in localStorage

terrymyc:
With a custom React hook `useDarkMode`, the dark mode status can be saved in `localStorage`.


J535D165:
Thanks! Any thoughts on implementing the text-size selection in this solution as well?

terrymyc:
It could be the same logic. Let me try it out.

terrymyc:
@J535D165 Text-size selection has been implemented in this solution.

PR #299, opened 2020-07-01
Fix project-delete check failure in Safari

### LINKED ISSUE #291: Checks before deleting a project fail
In the project settings, you should delete a project by typing the project name. Although the delete button is disabled:
- If you don't type, you can still delete a project by directly hitting the enter button on your keyboard. 
- If you type something other than the project name, you can also delete the project by hitting the enter button.

**Version information**
 - OS: macOS 10.15.5
 - ASReview version 0.9.5+6.g094acc0

terrymyc:
Fixes #291

### DIFF
--- a/asreview/webapp/src/ProjectSettings.js
+++ b/asreview/webapp/src/ProjectSettings.js
@@ -19,6 +19,9 @@ const useStyles = makeStyles({
   root: {
     maxWidth: "100%",
   },
+  input: {
+    display: "none",
+  },
   deleteButton: {
     margin: 5,
   },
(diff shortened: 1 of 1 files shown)


J535D165:
Not sure about this solution. How did you come up with this? Is this a bug in Material UI? 

I don't know why the keyboard triggers this event/function. Resolving that might be the best solution. Second best might be to add a check in this delete function: ` if (deleteInput !== props.id){`.

terrymyc:
I referred to how an upload button is defined. https://github.com/mui-org/material-ui/blob/master/docs/src/pages/components/buttons/UploadButtons.js#L23-L34

The issue is that the keystroke triggers form submission directly if `type="submit"` is put within `<Button />`. If I wrap `<Button />` with a `<input />` that doesn't have `type="submit"`. The submission will only be triggered either by keystroke or button click when the form is filled correctly with the project id.

It seems to be an issue with Safari. Once you put `type` within `<input />`, it doesn't function normally. When there is no `<input />` wrapper, it seems that a `<Button />` with `type` doesn't function normally either. It might have something to do with `.css`. https://github.com/necolas/normalize.css/blob/master/normalize.css#L191-L200

Adding a check in this delete function can also solve the issue. Which one do you prefer?

J535D165:
A clever solution to this Safari issue. 

I think I prefer the check within the function because it might be a bit more future proof, less code, and less error-prone. I think we will run into this issue multiple times (step 1 of the pre-review).

terrymyc:
Okay, a check has been included in the function.

PR #308, opened 2020-07-15
Fix review zone scroll bar

terrymyc:
- Fix double scroll bars in the review zone
- Fix extensive blank from the article title to the top


J535D165:
For me, this was the most annoying bug. Thanks!!

PR #309, opened 2020-07-21
Add stop words to TFIDF

qubixes:
I forgot to put it as a draft; it doesn't have to be merged right now.


asreview/feature_extraction/tfidf.py
```diff
@@ -9,21 +7,26 @@ class Tfidf(BaseFeatureExtraction):
...
         ngram_max: int
             Can use up to ngrams up to ngram_max. For example in the case of
             ngram_max=2, monograms and bigrams could be used.
+        stop_words: int
+            Use *ENGLISH* stop words.
         """
         super(Tfidf, self).__init__(*args, **kwargs)
         self.ngram_max = ngram_max
-        self._model = Pipeline([
-            ('vect', CountVectorizer(ngram_range=(1, ngram_max))),
```
J535D165, line 24: What happened with the CountVectorizer?

asreview/feature_extraction/tfidf.py
```diff
@@ -9,21 +7,26 @@ class Tfidf(BaseFeatureExtraction):
...
-            ('tfidf', TfidfTransformer())]
-        )
+        self.stop_words = stop_words
+        if stop_words == 1:
```
J535D165, line 24: Why not using the arguments like in sklearn? So 'english' instead of 1?

asreview/feature_extraction/tfidf.py
qubixes: TfidfVectorizer does the same thing without needing a pipeline. Haven't run the numbers, but I'd think it is more efficient.

asreview/feature_extraction/tfidf.py
qubixes: This is easier for hyper parameter optimization.

asreview/feature_extraction/tfidf.py
J535D165: Ow, thanks, I didn't notice that one is  `TfidfVectorizer` while the other is `TfidfTransformer`.

asreview/feature_extraction/tfidf.py
J535D165: I would like to have the API as flexible as possible. Can we move this to the hyper parameter optimization scripts? That would be great.

qubixes: I'm not sure if it's that simple. What is perhaps possible is to support ["english", "none", None]. It would be practically similar to the SKLearn API, and might cause less breakage (no guarantees).

asreview/feature_extraction/tfidf.py
```diff
@@ -36,7 +40,12 @@ def full_hyper_space(self):
         from hyperopt import hp
 
         hyper_space, hyper_choices = super(Tfidf, self).full_hyper_space()
+        hyper_choices.update({
+            "fex_stop_words": ["english", "none"]
```
J535D165, line 44: What downstream code needs the `"none"`? Can't we use `None`?

qubixes: Configuration files don't have a type, so there's no real way to distinguish "none" or None. In that particular line it might not break, but just mangled somewhere down the line. I think this is the safest way to do it right now.

PR #314, opened 2020-07-31
Add line chart to statistics panel on project page

terrymyc:
A new line chart made by @SagevdBrand is implemented to the statistics panel on the project page. It compares the efficiency of finding relevant documents by ASReview against at random.

localhost_3000_(iPad) (1)
localhost_3000_(iPad)


Rensvandeschoot:
Nice! This graph will be very helpful!! 
One question:
Should the random line not reach 38 inclusions after labelling 6000 papers instead of only 300?

terrymyc:
> Should the random line not reach 38 inclusions after labeling 6000 papers instead of only 300?

I agree. It will look like this:

localhost_3000_(iPad)

Rensvandeschoot:
Yes, indeed. So, after reading 177 papers at random you would have found only 1 relevant paper, but with active learning 38... impressive!
This graph really demonstrates the added value of active learning!!!

Rensvandeschoot:
Should we add some information abouth these three graphs on rtd? And add a pop-up or hyperlink in the front--end?

J535D165:
Sounds good. Let's start with some information on RTD. 
Let's brainstorm with @MartijnUX about displaying the info in the frontend. Let's add this in a new PR.

J535D165 (approved):

J535D165:
Thank you so much @SagevdBrand and @terrymyc!

PR #315, opened 2020-08-04
Update simulation documentation

GerbrichFerdinands:
Result from the ASReview doc-a-thon: 

- add why you would want to run a simulation
- shorter and more to the point explanation  

Together with @Sasafrass


docs/source/sim_overview.rst
```diff
@@ -1,43 +1,41 @@
 Overview
 ========
 
+.. role:: strike
+
+Why run a simulation?
+---------------------
+
+Doing simulations can be a great way to assess how well ASReview performs for your particular purposes. The user can run simulations on previously labelled datasets to see - for example - how much time the user could have saved using ASReview, or how previous inclusion criteria could be improved in the future. The ASReview simulation mode works by sequentially going through the dataset like an ASReview user would, and uses the labels to learn from these inclusions and exclusions. In this way, the entire screening process is replicated.
```
qubixes, line 9: I don't think "- for example -" is necessary.

I don't understand what you mean with "or how previous inclusion criteria could be improved in the future"? How? Maybe we need to be more specific?

I would remove the "sequentially" (since it's not?). How about:
"The ASReview simulation mode works through the dataset exactly like an ASReview user would, using the inclusions and exclusions provided in the dataset for the active learning cycle."

docs/source/sim_overview.rst
```diff
@@ -1,43 +1,41 @@
...
 Doing the simulation
 --------------------
 
-Doing simulations can be a great way to assess how well *ASReview* performs for your particular
-needs.
-There are a few ways to perform simulations with the ASReview. For how to prepare your data,
-see the `dataset <datasets.html>`__ page.
-
-The first option is to use the simulation mode that is provided with the _ASReview_ package. It can
+You can use the simulation mode that is provided with the ASReview package. It can
```
qubixes, line 14: Why is the link to the dataset page removed?

docs/source/sim_overview.rst
```diff
@@ -1,43 +1,41 @@
...
-model, and storing the results in the file ``myreview.h5``. More detail can be found in the section
-on the `cli simulation <cli.html>`__ page.
+.. code-block::
 
-A second option is to use the functionality of the 
-`asreview-simulation <https://github.com/asreview/automated-systematic-review-simulations>`_
-package. It makes it easier to run multiple simulations at the same time in case that we have
-more cores/processors available. Notice that it requires a library that implements the MPI standard,
-such as OpenMPI.
+	├── my-simulation
```
qubixes, line 25: I think this level will be confusing to users not so familiar with the command line. If you want a nested directory structure, then I suggest we put it in the command as well.

docs/source/sim_overview.rst
```diff
@@ -1,43 +1,41 @@
...
+	|   ├── MY_DATASET.csv
+	|   ├── myreview.h5
 
-Another option is to use the *ASReview* `API <api.html>`__ directly. The advantage is that you'll
-have more control over the flow of the simulation, but you will need to program some code in python.
+
```
qubixes, line 29: I think some reference has to be made to the API. The explanation doesn't have to be on the same page, but in the end it should be *somewhere*. And I know a lot has to be written in that regard, it can be a stub for now.

docs/source/sim_overview.rst
```diff
@@ -56,55 +54,28 @@ give you at least a basic exploratory idea of the performance of your review:
...
+
 Simulation options
 ------------------
-The options for simulating are shown on the `CLI <cli.html>`__ page. We will highlight some of the
-more often used options here.
+The options for simulating are shown on the `CLI <cli.html>`__ page. We will highlight some of the more often used options here. When no additional arguments are specified in the ``asreview simulate`` command, default settings are used.
+
+To make your simulations reproducible you can use the ``--seed`` and ``--init_seed`` options. 'Init
+seed' controls the starting set of papers to train the model on, while the 'seed' controls the
```
qubixes, line 65: 'init_seed'

docs/source/sim_overview.rst
```diff
@@ -56,55 +54,28 @@ give you at least a basic exploratory idea of the performance of your review:
...
+random generator after initialization.
```
qubixes, line 66: controls the seed of the random number generation that is used after initialization.

docs/source/sim_overview.rst
```diff
@@ -56,55 +54,28 @@ give you at least a basic exploratory idea of the performance of your review:
...
 
-To make your simulations reproducible you can use the ``--seed`` and ``--init_seed`` options. Init
-seed controls the starting set of papers to train the model on. While the normal seed controls the
-random generator after initialization. For initialization you can also use ``--prior_idx`` to set
-the papers to start the simulation with.
+By default, the model initializes with one relevant and one irrelevant record. If you want to initialize your model with a specific set of starting papers, you can use ``--prior_idx`` to select the indices of the papers you want to start the simulation with.
```
qubixes, line 68: Would swap the order of this and the paragraph below. You can set the number of priors by ..., but if you want more control, use ...

docs/source/sim_overview.rst
```diff
@@ -56,55 +54,28 @@ give you at least a basic exploratory idea of the performance of your review:
...
 
-To limit the size of your state file and the time to simulate, you can increase ``--n_instances``,
-which reduces the number of training iterations.
+The ``--n_instances`` argument controls the number of records that have to be labelled before the model is retrained again and is set at 1 by default. If you want to reduce the number of training iterations, for example to limit the size of your state file and the time to simulate, you can increase ``--n_instances``.
 
-Use a different classifier with the ``-m`` flag. Names for implemented models are listed on the
-`models <models.html>`__ page.
+You can select a classifier with the ``-m`` flag, which is set to be Naive Bayes by default. Names for implemented classifiers are listed on the :doc:`models <models.html>`page.
```
qubixes, line 72: Fix the warning, please.

docs/source/sim_overview.rst
```diff
@@ -56,55 +54,28 @@ give you at least a basic exploratory idea of the performance of your review:
...
-	
-	[feature_param]
-	ngram_max = 2
-	split_ta = 0
-	use_keywords = 0
-	
-	[model_param]
-	alpha = 0.8267010267867003
+The last element that can be changed is the `balance strategy <balance_strategies.html>`__, and is
+changed with the ``-b`` flag. Default is triple balance.
```
qubixes, line 81: We should change that...

docs/source/sim_overview.rst
```diff
@@ -56,55 +54,28 @@ give you at least a basic exploratory idea of the performance of your review:
...
-is set with the ``-q`` option.
+can be set with the ``-q`` option.
 
-For feature feature extraction, supply the ``-e`` flag, with more details on the page for 
+For feature extraction, supply the ``-e`` flag. Default is TF-IDF, more details are on the page for
 `feature extraction <feature_extraction.html>`__.
 
-The last model that can be changed is the `balance strategy <balance_strategies.html>`__, and is
-changed with the ``-b`` flag.
-
```
qubixes, line 84: I also think this should be somewhere. Perhaps not on this page, that's true.

docs/source/sim_overview.rst
GerbrichFerdinands: The goal of the sentence is to give an example of that seeing how much time can be saved is not the only reason for using simulation mode. What is meant that we can simulate how difficult or easy it is too find certain papers and use that information to verify the screening process (maybe some inclusions shouldn't be inclusions, maybe some exclusions shouldn't be exclusions). What do you think? 

About the sentence:

"The ASReview simulation mode works through the dataset exactly like an ASReview user would, using the inclusions and exclusions in the dataset to train the model in the active learning cycle."

docs/source/sim_overview.rst
GerbrichFerdinands: I don't know. I've added it again!

docs/source/sim_overview.rst
GerbrichFerdinands: There is no need to put any directory structure in the command line, right? 
I made up a folder called 'my-simulation' to make clear that the data and the state file are in the same directory. Do you want to get rid of this folder? Do you suggest leaving that whole thing out?

docs/source/sim_overview.rst
GerbrichFerdinands: @J535D165 Where should we put it?

docs/source/sim_overview.rst
GerbrichFerdinands: Which paragraph do you mean? The next paragraph is about `n_instances`, which is about the number of records to be labeled before the model retrains again. I've updated the current paragraph by adding the arguments `--n_prior_included` and `--n_prior_excluded`: 

> By default, the model initializes with one relevant and one irrelevant record. You can set the number of priors by `--n_prior_included` and `--n_prior_excluded`. However, if you want to initialize your model with a specific set of starting papers, you can use ``--prior_idx`` to select the indices of the papers you want to start the simulation with.

docs/source/sim_overview.rst
GerbrichFerdinands: The balance strategies are still in the document, they just moved a bit

docs/source/sim_overview.rst
GerbrichFerdinands: I've created a PR #316 updating the default query strategy to max and balance strategy to double

GerbrichFerdinands:
Thanks @qubixes for the helpful feedback! And sorry for my late response.

docs/source/sim_overview.rst
qubixes: Well, I think this is the place to put the other reasons to run a simulation then. Perhaps if you make it longer, you can also split up "The ASReview simulation works .." to its own section, since it's not really the "why"? Your proposed new sentence seems fine to me.

docs/source/sim_overview.rst
qubixes: Yes, I would leave that folder out, since users might try to run the command from the wrong folder. Another solution might be to put in an explicit folder change in the documentation.

docs/source/sim_overview.rst
qubixes: Yes, that's what I meant.

docs/source/sim_overview.rst
qubixes: Yes, because I'm not sure it is anywhere else. We should at least link that page.

PR #316, opened 2020-08-13
Update default query strategy

GerbrichFerdinands:
As discussed with @Rensvandeschoot @J535D165 and @qubixes, change the default query strategy from `max_random` to `max`. From 95% max sampling & 5% random sampling to 100% max sampling. This leads to better performance. The balance strategy then also needs to be updated from `triple` to `double`.


J535D165:
Thanks! It is important to keep this one linked to https://github.com/asreview/asreview/pull/317/files#r471390624. The frontend should request these defaults, instead of defining their own.

PR #317, opened 2020-08-14
Add feature extraction & query strategy options

terrymyc:
In a project setup page, users can select a feature extraction technique and a query strategy.

localhost_3000_(iPad) (1)


asreview/webapp/src/PreReviewComponents/ProjectAlgorithms.js
```diff
@@ -48,7 +53,9 @@ const useStyles = makeStyles(theme => ({
 }));
 
 const defaultModels = {
-  model: "nb"
+  model: "nb",
+  query_strategy: "max",
+  feature_extraction: "tfidf",
 }
```
J535D165, line 59: This should be a GET request in the future: https://github.com/asreview/asreview/pull/316. This avoids defining the defaults twice.

GerbrichFerdinands: I think this relates to something @qubixes and I talked about last week. 

Is it possible to request models from a configuration file? (in the background) Similar to the option `--config-file config.ini` in the CLI. Might be convenient when we scale up to more complex configurations (hyperparameter specifications etc).

J535D165: Not completely sure that I understand you correctly. I think all models should be registered centrally. This makes it easy to pass them to the front-end. Plugins can be used to add new models.

terrymyc: > This should be a GET request in the future: #316. This avoids defining the defaults twice.

Resolved.

J535D165: I like this solution. It differs from what I was thinking of, but seems to work. Let me check for some side effects before merging.

J535D165:
How does this PR deal with non-existing combinations of classifiers and feature extraction techniques?

terrymyc:
> How does this PR deal with non-existing combinations of classifiers and feature extraction techniques?

When Naive Bayes is chosen, Doc2Vec is disabled. When Doc2Vec is chosen, Naive Bayes is disabled. Any other combination is not feasible?
By the way, do you think an alert is needed when Doc2Vec is chosen? Telling users it requires an additional package, and takes a longer time to train the model?

GerbrichFerdinands:
> > How does this PR deal with non-existing combinations of classifiers and feature extraction techniques?
> 
> When Naive Bayes is chosen, Doc2Vec is disabled. When Doc2Vec is chosen, Naive Bayes is disabled. Any other combination is not feasible?
> By the way, do you think an alert is needed when Doc2Vec is chosen? Telling users it requires an additional package, and takes a longer time to train the model?

I think the alert is a good idea. 
There are no other impossible combinations, but I think it is good to give a warning when selecting SVM + TF-IDF because it is very slow (and in a lesser sense, also SVM + D2V perhaps).

asreview/webapp/src/PreReviewComponents/ProjectAlgorithms.js
GerbrichFerdinands: > Not completely sure that I understand you correctly. I think all models should be registered centrally. This makes it easy to pass them to the front-end. Plugins can be used to add new models.

@J535D165 What do you mean with registred centrally? I am not talking about adding new models, but implementing additional features such as simulation and hyperparameter settings. Maybe it doesn't work that way, but I think if the models are requested through a configuration file, than later on, this file can also be used to specify settings simulations, and hyperparameters. 

I mean files like this one: 

```
[global_settings]
n_instances = 1
n_prior_included = 1
n_prior_excluded = 1
model = nb
balance_strategy = double
query_strategy = max
feature_extraction = tfidf

[balance_param]
a = 1249.910315374313
alpha = 0.7043792455091337
b = 0.9993171869006371

[feature_param]
ngram_max = 2
split_ta = 1
use_keywords = 1

[model_param]
alpha = 0.17584727645029802
 
```

qubixes: @J535D165 I think long term it might be good to work with the configuration files that @GerbrichFerdinands mentions. Because that should give better performance for the user than using the defaults for each of the individual model parts (classifier/balance/feature). Then allowing the user to upload these configuration files makes it so that advanced users can use any combination + set parameters as they want.

The list here could be a drop down list of "curated" combinations, where we can refer to the simulation study by @GerbrichFerdinands to indicate the performance:

NB+TFIDF ~> 100%
SVM+TFIDF ~> 97%

And perhaps some indication of running time as well, though I'm not sure how useful that is in this context:

NB+TFIDF ~> 2 seconds initialization, 0.1s running time.
Logistic+Doc2Vec ~> 10 min initialization, 0.1s running time.

Just some ideas. :)

Rensvandeschoot:
I propose to add "(default)" to the default options and a pop-up or link to more information why we have chosen these defaults. Should this be rtd or.the website? And we can add the papers with simulation results.

Also, maybe we should delete the sentence with 'skip tgis step..' because there is no option for skipping and you should always select something in this step, right?

terrymyc:
Algorithms

AlgorithmHelp

@Rensvandeschoot I agree. Maybe like this?

Rensvandeschoot:
Yes, exactly like this! 

@J535D165 do you agree??

asreview/webapp/src/PreReviewComponents/ProjectAlgorithms.js
J535D165: > The list here could be a drop down list of "curated" combinations, where we can refer to the simulation study by @GerbrichFerdinands to indicate the performance:
NB+TFIDF ~> 100%
SVM+TFIDF ~> 97%
And perhaps some indication of running time as well, though I'm not sure how useful that is in this context:
NB+TFIDF ~> 2 seconds initialization, 0.1s running time.
Logistic+Doc2Vec ~> 10 min initialization, 0.1s running time.

I like this one.

Levels of automation:
```
0. Use the defaults
1. Pick one of the combinations (like you propose)
2. Select with the tool made by @terrymyc. 
3. Upload a file with settings
```

We should think about the UX with @MartijnUX

J535D165:
For now I'm fine with this solution. However, we talked about making this an 'closed' card by default. The closed card should be the default. I think I like that solution. But, let's stick to this solution right now. It's intuitive and simple to implement.

terrymyc:
> The closed card should be the default.

Now it is. The select menu will show up once users click on the edit button.

image

J535D165 (approved):

J535D165:
Nice!!

PR #318, opened 2020-08-14
Add a quick tour for new users

terrymyc:
A quick tour designed by @JoukjeWillemsen is implemented for new users. It only appears once when users open the software in a browser for the first time.

Welcome

SetUp

Start

Benefit

DontStress

Done

Publish


terrymyc:
@J535D165 The texts in the quick tour can be edited now. The quick tour shows up once after a major or minor release. It might be an idea of adding a "Get started" card at the end of the tour with an option "Don't show this quick tour again."

J535D165 (approved):

J535D165:
Thank you so much @terrymyc and @JoukjeWillemsen for all the work in this PR. Good to see it is future proof and robust.

Rensvandeschoot:
Will our users know what the original purpose was of a floppy disk??

J535D165:
I think so. New title is: "Your projects are autosaved and stored locally". Subtitle: "Your screening decisions are automatically saved on your own device.".

PR #321, opened 2020-08-17
Add an option to mark a project as finished

terrymyc:
An option to mark a project as finished/in review is implemented in the publication zone of the project page. Thank @JoukjeWillemsen for designing the new Elas icon.

ElasHoldFinishedSign
MarkFinished
MarkInReview


terrymyc:
@J535D165 There has been an Elas holding sign "[Completed](https://github.com/JoukjeWillemsen/asreview-artwork/blob/master/ElasAdventures/5.HoldingSIGNS/SVG/Completed_200x200.svg)." Will it be replaced by "Finished"? If not, @JoukjeWillemsen may come up with a different color for "Finished."

Rensvandeschoot:
@terrymyc and @JoukjeWillemsen: this is really nice work!! 

One question, is there a difference between completed and finished? I can imagine the difference is  that the  technical log file and a paper have been published on OSF. What is your view?

asreview/webapp/api.py
```diff
@@ -797,6 +797,29 @@ def export_project(project_id):
...
+@bp.route('/project/<project_id>/finish', methods=["GET"])
+def api_finish_project(project_id):
+    """Mark a project as finished or not"""
+
+    # read the file with project info
+    with open(get_project_file_path(project_id), "r") as fp:
+        project_info = json.load(fp)
+
+    try:
+        project_info["projectFinished"] = not project_info["projectFinished"]
```
J535D165, line 809: Let's rename this into `reviewFinished`. In the future, we might want to extend the scope of the application.

asreview/webapp/src/PublicationZone.js
```diff
@@ -101,6 +101,19 @@ const PublicationZone = (props) => {
...
+          <Divider component="li" />
+          <ListItem
+            button
+            onClick={props.finishProject}
+            alignItems="flex-start"
+            disabled={!props.showExportResult}
+            key="finish-project"
+          >
+            <ListItemText
+              primary={props.projectFinished ? "Mark as in review" : "Mark as finished"}
```
J535D165, line 113: "Mark as in review", what about "Mark as finished (undo)"?

J535D165: "Mark screening as finished"?

J535D165 (changes requested):
Nice implementation. The remarks by @Rensvandeschoot make sense. The focus should be on the review process. The review process is completed or finished. A published project is something else (long term feature request).

asreview/webapp/src/PublicationZone.js
terrymyc: > "Mark screening as finished"?

How about:
"Mark reviewing as finished"
"Mark reviewing as finished (undo)"

Rensvandeschoot: I propose to use 'mark screening as complete' or '.. finished' and use 'project published' for when we implemented the link with OSF with the option to add a doi for the technical logfile plus a seperate doi for the paper.

terrymyc: MarkAsFinished
MarkAsFinishedUndo

J535D165 (approved):
Nice backward compatibility. Works like a charm.

PR #323, opened 2020-08-19
Update config.py

GerbrichFerdinands:
Add `label_abstract_screening`as a possible label for abstract inclusions.

This name makes more sense in combination with `label_included`, which is used mainly in the systematic-review-datasets-repository @terrymyc


J535D165:
Thanks, makes sense. I changed the order. The first item is the preferred wording. `label_abstract_screening` and `label_included`.

PR #325, opened 2020-08-19
Add keyboard shortcut while reviewing

terrymyc:
Three keyboard shortcuts are implemented. While reviewing, users can strike "r"/"R" to label "relevant", "i"/"I" to label "irrelevant", and "u"/"U" to undo labeling. This feature can be toggled on/off in the settings dialog.

KeyboardShortcut

### DIFF
--- a/asreview/webapp/src/SettingsDialog.js
+++ b/asreview/webapp/src/SettingsDialog.js
@@ -92,7 +92,7 @@ export default function SettingsDialog(props) {
               </ListItemSecondaryAction>
             </ListItem>
             <ListItem>
-              <ListItemText id="switch-list-label-undo" primary="Undo Enabled" />
+              <ListItemText id="switch-list-label-undo" primary="Undo" />
               <ListItemSecondaryAction>
                 <Switch
                   edge="end"
@@ -103,7 +103,7 @@ export default function SettingsDialog(props) {
               </ListItemSecondaryAction>
             </ListItem>
             <ListItem>
-              <ListItemText id="switch-list-label-key" primary="Keyboard shortcut" />
+              <ListItemText id="switch-list-label-key" primary="Keyboard shortcuts" />
               <ListItemSecondaryAction>
                 <Switch
                   edge="end"
--- a/asreview/webapp/src/SettingsDialog.js
+++ b/asreview/webapp/src/SettingsDialog.js
@@ -102,6 +102,17 @@ export default function SettingsDialog(props) {
                 />
               </ListItemSecondaryAction>
             </ListItem>
+            <ListItem>
+              <ListItemText id="switch-list-label-key" primary="Keyboard shortcut" />
+              <ListItemSecondaryAction>
+                <Switch
+                  edge="end"
+                  onChange={props.toggleKeyPressEnabled}
+                  checked={props.keyPressEnabled}
+                  inputProps={{ 'aria-labelledby': 'switch-list-label-key' }}
+                />
+              </ListItemSecondaryAction>
+            </ListItem>
           </List>
 
         </DialogContent>
--- a/asreview/webapp/src/SettingsHooks.js
+++ b/asreview/webapp/src/SettingsHooks.js
@@ -84,13 +84,13 @@ const useKeyPress = (targetKey) => {
   useEffect(() => {
 
     const downHandler = ({ key }) => {
-      if (key === targetKey) {
+      if (key.toLowerCase() === targetKey) {
         setKeyPressed(true);
       }
     };
 
     const upHandler = ({ key }) => {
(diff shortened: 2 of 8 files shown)


asreview/webapp/src/SettingsHooks.js
```diff
@@ -77,4 +77,58 @@ const useUndoEnabled = () => {
...
+    // Remove event listeners on cleanup
+    return () => {
+      window.removeEventListener('keydown', downHandler);
+      window.removeEventListener('keyup', upHandler);
+    };
+
+  }, [targetKey]);
+
+  return keyPressed;
+};
```
J535D165, line 110: Does it make sense to put this one over here? It looks like something we can move to a standalone file in the subfolder with the review components. What do you think @terrymyc?

J535D165 (commented):
This is a milestone. Very neat implementation as well. Hopefully you enjoyed implementing this one!

Can you add some documentation on this feature? Somewhere in the ReadTheDocs will do. (maybe we have to make a new section "Keyboard shortcuts")

asreview/webapp/src/SettingsDialog.js
```diff
@@ -102,6 +102,17 @@ export default function SettingsDialog(props) {
                 />
               </ListItemSecondaryAction>
             </ListItem>
+            <ListItem>
+              <ListItemText id="switch-list-label-key" primary="Keyboard shortcut" />
```
J535D165, line 106: "Enable keyboard shortcuts"?

asreview/webapp/src/SettingsHooks.js
terrymyc: It was a standalone file, but I had no idea which folder I should put it in since it is a hook. Do you have any advice?

asreview/webapp/src/SettingsDialog.js
terrymyc: I'd say "Enable" is not necessary, as it is clear to see from the switch icon when it is toggled on or off. If it is "Enable xxx", it should be changed to "Disable xxx" when it is toggled on. So, the same logic applies to that "Undo Enable" could be "Undo labeling".
Maybe we also need a help text of settings dialog to explain these features briefly.

terrymyc:
> Can you add some documentation on this feature? Somewhere in the ReadTheDocs will do. (maybe we have to make a new section "Keyboard shortcuts")

Is there any documentation on other features in the settings dialog in the ReadTheDocs? Shall we have a new section "Customize settings" at the end of the quick tour in the ReadTheDocs?

J535D165:
I'm more thinking about a small article on keyboard shortcuts only. There is nothing for the other options at the moment. 

Maybe we should work on something like this:

- Features
  - Keyboard shortcuts
  - Undo
  - Change text size

But let's start with the keyboard shortcuts.

Rensvandeschoot:
So nice!!!

I agree we carefully need to think about the text in the settings menu, maybe involve Martijn?

For the rtd I like the suggestion of the Feature structure, also the other options need to be explained and we will probably develop more of such features shortly.

asreview/webapp/src/SettingsHooks.js
J535D165: Shall we make a `hooks` folder in `src`? I like the idea of making this one into a hook btw. 

There is also another hook for the Help dialog. I think we can move that one there as well (I will do this one).

asreview/webapp/src/SettingsDialog.js
J535D165: Agree.

J535D165:
> I agree we carefully need to think about the text in the settings menu, maybe involve Martijn?

https://material.io/design/platform-guidance/android-settings.html#label-and-secondary-text

@MartijnUX can you have a look on this one?

@terrymyc Let's skip this for this PR and focus on the docs and hooks folder.

MartijnUX:
To overcome "Feature creep" we could already consider to user to implement a menu for the preferences/settings. This way we help the user make a selection of what to change. "The four-item rule applies to (short-term) memory retrieval". If we want to help the user choosing, we should try to keep it as minimal as possible. So, my advise would be to make the panel a bit bigger. 
image

J535D165:
Makes sense. I think we can make this work on small screens as well. This requires some custom made components. I think we can reuse this is step 3 (overview relevant/irrelevant items) as well.

MartijnUX:
We could also use a Tab component. https://material-ui.com/components/tabs/#vertical-tabs

Rensvandeschoot:
I agree with such menu's or tabs, since more features are comming and we will violate the '4 rule' sooner than later

J535D165 (approved):

PR #331, opened 2020-08-31
Gifs end user testing

SagevdBrand:
Adding gifs of the recall curves to the documentation of the end user testing.


J535D165:
Thanks! Live with the next release!

Issue #332, opened 2020-09-03
Importing RIS export from EndNote gives "Error" 


GerbrichFerdinands:
**Describe the bug**
I've exported a large dataset (13,000 records) from EndNote as a RIS file. 
This leads to a file with a `.txt` extension. 

Uploading this file into asreview to an error message in the GUI (screenshot 1).
Looking at the CLI output, at first it seemed there was one problematic record (screenshot 2).
However, after deleting this record from the file with a text editor, I get an encoding error (screenshot 3). 
When I upload a part of of the references, sometimes it works and sometimes it does not (screenshot 4)

I think there might be another corrupt entry in the .txt file, but the error message I'm getting is not informative. 
Do you know of a solution, for example, are there any tools how to inspect a RIS files for corrupt entries? 
Or could it be another problem? 

**To Reproduce**
Steps to reproduce the problem:

1. Export EndNote references as RIS export. 
2. Upload the resulting .txt file in ASReview (screenshot 1)
3. Thanks to an informative error message in the CLI, I can remove the corrupt reference in a text editor.
4. Upload again. (result = screenshot 2)
5. Resave the .txt file as a .ris file in a text editor.
6. Upload again (result = screenshot 3)
7. Create new document with a subset from the original text file output from endnote.
8. Upload again (result = screenshot 4) 

**Screenshots**

1. 
![image](https://user-images.githubusercontent.com/45358236/92101511-84376680-eddd-11ea-8285-7bbf713a1c40.png)

2.

![image](https://user-images.githubusercontent.com/45358236/92101523-8994b100-eddd-11ea-8f62-2859a173f209.png)

3.
![image](https://user-images.githubusercontent.com/45358236/92101997-3a9b4b80-edde-11ea-9e78-f51a663941b7.png)

4. 
![image](https://user-images.githubusercontent.com/45358236/92102371-d036db00-edde-11ea-9de4-46660c909395.png)

**Version information**
 - OS: macOS
 - ASReview version 11.1

J535D165:
Thanks for reporting. I think there are 2 potential issues. One with the encoding in `risparser` (now `rispy`) and one with inferring data types in pandas. Please send me a copy of the file to debug. 

Did a PR to clarify the documentation.

qubixes:
@J535D165 Has this been fixed/debugged?

J535D165:
Nope, unfortunately not. Only a minor update on the docs.

qubixes:
[...]

PR #333, opened 2020-09-03
Add txt to import options for RIS

J535D165:
To clarify that `.txt` import is handled as RIS import. This wasn't documented. #332


GerbrichFerdinands:
Thanks! This is helpful.

PR #335, opened 2020-09-04
Update paths.py

MathieuRietman:
When using the os.environ["ASREVIEW_PATH"] the create project failes with AttributeError: 'str' object has no attribute 'mkdir' as the environment is no path in the asreview_path.mkdir(parents=True, exist_ok=True)


J535D165 (approved):

J535D165:
Thanks for this fix @MathieuRietman. I will prepare a patch release.

J535D165:
Version 0.11.2 was released to fix this bug. Thanks again.

PR #337, opened 2020-09-16
Show a warning instead of failing on bad labels

### LINKED ISSUE #332: Importing RIS export from EndNote gives "Error"

qubixes:
Resolves #332 

Problem is that the RIS 'LB' tag can be used for things other than inclusion/exclusion labels. This pull request should instead of failing immediately give a warning that the labels could not be read.

### DIFF
--- a/asreview/io/ris_reader.py
+++ b/asreview/io/ris_reader.py
@@ -51,8 +51,10 @@ def read_ris(fp):
                 mapping = _tag_key_mapping(reverse=False)
                 entries = list(readris(bibliography_file, mapping=mapping))
                 break
-        except (UnicodeDecodeError, IOError):
+        except UnicodeDecodeError:
             pass
+        except IOError as e:
+            logging.warning(e)
 
     if entries is None:
         raise ValueError("Cannot find proper encoding for data file.")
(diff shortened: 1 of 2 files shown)


J535D165 (approved):

J535D165:
Thanks a lot, it works!

PR #338, opened 2020-09-18
Add an Alert for Exploration Mode

terrymyc:
While using the example datasets, users may see an alert during the reviewing process.
- End-user testing is renamed Exploration Mode.
- Relevant documents are displayed in green, the same color as the charts on the side statistics panel.

exploration


J535D165 (approved):

J535D165:
Yeah, nice!

Rensvandeschoot:
Well done!

PR #342, opened 2020-09-25
Fix int titles

qubixes:
Convert columns "title", "abstract", "authors", and "keywords" to string if not the case.


J535D165:
Thanks for this fix!

PR #343, opened 2020-09-25
Remove print statements and improve logging

J535D165:
This PR removes print statements in the REST api and SQlock handler. Print statements are replaced by logging statements that start with the project identifier. 

@qubixes can you have a look at the changes in the sqlock file?

### DIFF
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -95,7 +95,6 @@ def api_get_projects():  # noqa: F401
     """Get info on the article"""
 
     projects = list_asreview_project_paths()
-    logging.debug(list_asreview_project_paths)
 
     project_info = []
     for proj in projects:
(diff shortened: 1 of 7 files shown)


asreview/webapp/api.py
```diff
@@ -389,11 +389,11 @@ def api_get_project_data(project_id):  # noqa: F401
         statistics["filename"] = filename
 
     except FileNotFoundError as err:
-        print(err)
+        logging.info(err)
```
qubixes, line 392: Should be logging.error probably?

qubixes:
@J535D165 It doesn't seem like anything material changed except the inclusion of the project_id + logging? Seems all good to me.

asreview/webapp/api.py
J535D165: Thanks. This one is oke. No file found is a 200 response.

J535D165:
Yes, correct. No big changes. 

I'm facing some issues on displaying the logging messages in the stdout right now. Not completely sure how to deal with this. I would like to have all logging (asreview core, asreview lock manager, asreview REST API) statements in my stdout. Any ideas?

qubixes:
I'm sorry, I have no idea about everything in stdout. The problem is that some of it goes to stderr? Maybe it can be redirected on the command line?

J535D165:
My bad, this had to do with the log level and some other things. It's more or less resolved now. Thanks.

PR #346, opened 2020-09-26
Fix docstrings and documentation for asreview classifiers

J535D165:
This PR 

- Fixes docstrings for classifiers, balance strategies, feat extraction and query strategies. 
- Extends the reference docs https://asreview.readthedocs.io/en/latest/reference.html
- Add class members to docs
- Add models that depend on optional dependencies to docs
- Add models that depend on optional dependencies to submodules (`__init__.py`)
- Applies YAPF formatter to submodules. 
- Adds CI pipeline for bare minimum of dependencies.  
- Tests models for import errors if dependencies are not available.
- A few more things.

### DIFF
--- a/asreview/balance_strategies/double.py
+++ b/asreview/balance_strategies/double.py
@@ -20,11 +20,6 @@ from asreview.balance_strategies.base import BaseBalance
 from asreview.balance_strategies.simple import SimpleBalance
 from asreview.utils import get_random_state
 
-PARAM_A = 2.155
-PARAM_ALPHA = 0.94
-PARAM_B = 0.789
-PARAM_BETA = 1.0
-
 
 class DoubleBalance(BaseBalance):
     """Dynamic Resampling balance strategy.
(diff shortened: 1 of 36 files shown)


asreview/review/base.py
```diff
@@ -259,7 +260,13 @@ def _stop_iter(self, query_i, n_pool):
         return stop_iter
 
     def n_pool(self):
```
J535D165, line 262: property? @qubixes

asreview/review/base.py
```diff
@@ -460,7 +467,14 @@ def train(self):
             self.query_i_classified = 0
 
     def statistics(self):
```
J535D165, line 469: property? @qubixes

asreview/balance_strategies/double.py
```diff
@@ -20,6 +20,11 @@
 from asreview.balance_strategies.simple import SimpleBalance
 from asreview.utils import get_random_state
 
+PARAM_A = 2.155
```
qubixes, line 23: Why this restructuring?

J535D165: Reverted.

PR #353, opened 2020-10-01
Update help text for prior_idx in simulate.py

SagevdBrand:
Change help text for prior_idx, since it was confusing whether it referred to the column of id numbers, or rownumbers in the file.

### DIFF
--- a/asreview/entry_points/simulate.py
+++ b/asreview/entry_points/simulate.py
@@ -75,7 +75,7 @@ def _simulate_parser(prog="simulate", description=DESCRIPTION_SIMULATE):
         default=[],
         nargs="*",
         type=int,
-        help="Prior indices by id."
+        help="Prior indices by rownumber (0 is first rownumber)."
     )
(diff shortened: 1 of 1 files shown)


J535D165 (approved):

qubixes:
Hmmmm, I don't think this really makes it better. One thing to consider for example is that RIS/XML files don't really have rows. I agree that it should be updated/expanded though.

How about:

"List of record ids used as the initial inclusions/exclusions for the simulation. In the presence of a column "record_id", these will be used for this purpose. Otherwise they are sequential in the order that records occur, starting from 0. For example, in a csv/xlsx file the first row with data will have a record_id of 0, etc."

J535D165:
Hmm, this behaviour makes it more complicated. The variable name suggests an index value, not an id. I think we need documentation on RTD on this. For now, I think I will merge the PR because it improves the help message. Let's work on a more fundamental improvement in the nearby future.

Issue #352, opened 2020-10-01
Simulation - Deterministic model results in different recall curves when not seeded


J535D165:
The CLI suggests that there are 2 seeds available: one for the model and one for the prior knowledge. 

```
  --seed SEED           Seed for models. Use integer between 0 and 2^32 - 1.
  ...
  --init_seed INIT_SEED
                        Seed for setting the prior indices if the --prior_idx option is not used. If the option --prior_idx is used with one or more index, this option is ignored.
```

Running a simulation with both seeds will result in identical results. 

```
asreview simulate hall -s result_run_3.h5 --init_seed 1 --seed 1
asreview simulate hall -s result_run_4.h5 --init_seed 1 --seed 1

asreview plot result_run_3.h5 --type 'inclusion' -o plot_run_3.png
asreview plot result_run_4.h5 --type 'inclusion' -o plot_run_4.png
```

![plot_run_4](https://user-images.githubusercontent.com/12981139/94785169-7b7f8380-03cf-11eb-9070-6c82a67b0596.png)
![plot_run_3](https://user-images.githubusercontent.com/12981139/94785165-7ae6ed00-03cf-11eb-8989-c3de0d3f858b.png)

The default model is the deterministic NB model. So I thought I could skip the --seed value. However, this results in different results. Any thoughts?

```
asreview simulate hall -s result_run_1.h5 --init_seed 1
asreview simulate hall -s result_run_2.h5 --init_seed 1

asreview plot result_run_1.h5 --type 'inclusion' -o plot_run_1.png
asreview plot result_run_2.h5 --type 'inclusion' -o plot_run_2.png
```

![plot_run_1](https://user-images.githubusercontent.com/12981139/94785260-a8339b00-03cf-11eb-8b36-3e9bcadb9f93.png)
![plot_run_2](https://user-images.githubusercontent.com/12981139/94785262-a964c800-03cf-11eb-9259-b52f5d58f765.png)

GerbrichFerdinands:
My first thought: what about the Dynamic Resampling of the training data? 
There is a random component in sampling from the labeled data.

J535D165:
That was my first thought as well. But this issue was merged and released: https://github.com/asreview/asreview/pull/316 https://github.com/asreview/asreview/releases/tag/v0.12. 

This is an overview of my versions:

```
jonathans-mbp:bug_test jonathan$ asreview simulatie -h
usage: asreview [-h] [-V] [subcommand]

Automated Systematic Review (ASReview).

[...]

PR #354, opened 2020-10-01
Some clarification in the `-h` documentation.

GerbrichFerdinands:
- updated documentation for `--config_file` 
- updated documentation for `--seed`

Hopefully solving #352 

@qubixes can you review? 
I needed >1 line for the updates, not sure if that's allowed and whether my use of `"""` is correct.

### DIFF
--- a/asreview/entry_points/base.py
+++ b/asreview/entry_points/base.py
@@ -74,15 +74,15 @@ def _base_parser(prog=None, description=None):
         "--config_file",
         type=str,
         default=None,
-        help="""Configuration file with model parameters and 
-            hyperparameter values."""
+        help="Configuration file with model parameters and 
+            hyperparameter values."
     )
     parser.add_argument(
         "--seed",
         default=None,
         type=int,
-        help="""Seed for the model (classifiers, balance 
+        help="Seed for the model (classifiers, balance 
             strategies, feature extraction techniques, and query 
-            strategies). Use integer between 0 and 2^32 - 1."""
+            strategies). Use an integer between 0 and 2^32 - 1."
     )
     return parser
--- a/asreview/entry_points/base.py
+++ b/asreview/entry_points/base.py
@@ -74,7 +74,7 @@ def _base_parser(prog=None, description=None):
         "--config_file",
         type=str,
         default=None,
-        help="Configuration file with model model settings"
+        help="Configuration file with model settings"
              "and parameter values."
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


asreview/entry_points/base.py
```diff
@@ -74,12 +74,15 @@ def _base_parser(prog=None, description=None):
...
+            hyperparameter values."""
     )
     parser.add_argument(
         "--seed",
         default=None,
         type=int,
-        help="Seed for models. Use integer between 0 and 2^32 - 1."
+        help="""Seed for the model (classifiers, balance 
+            strategies, feature extraction techniques, and query 
+            strategies). Use integer between 0 and 2^32 - 1."""
```
qubixes, line 86: "Use integer" -> "Use an integer"

asreview/entry_points/base.py
```diff
@@ -74,12 +74,15 @@ def _base_parser(prog=None, description=None):
         "--config_file",
         type=str,
         default=None,
-        help="Configuration file with model, balance, and query parameters"
+        help="""Configuration file with model parameters and
```
qubixes, line 77: What is the difference between model parameters and hyperparameters?

qubixes:
You can also use single quotes on multiple lines; it should merge them into a single string (since they are enclosed).

asreview/entry_points/base.py
GerbrichFerdinands: I think of it as: 

Model parameters are the global settings, like  `balance_strategy = double` 
Hyperparameter values are nested in the model parameters, like `alpha = 1.229374068496514`

```
[global_settings]
n_instances = 1
n_prior_included = 1
n_prior_excluded = 1
model = nb
balance_strategy = double
query_strategy = max
feature_extraction = tfidf

[balance_param]
a = 1.3783791240515941
alpha = 1.229374068496514
b = 0.9153629982662178

[feature_param]
ngram_max = 3
split_ta = 0
use_keywords = 1

[model_param]
alpha = 0.06586028856339346
```

qubixes:
Ik bedoelde eigenlijk dit (voor voorbeelden):

https://github.com/qubixes/asreview-pro/blob/master/asreviewcontrib/pro/entrypoint.py

J535D165 (approved):

asreview/entry_points/base.py
qubixes: I think that might be confusing for the user. Perhaps we can call it simulation settings + model parameters?

GerbrichFerdinands: I'm not a fan of simulation settings as these settings are not confined to simulations. 
How about model settings + model parameters?

qubixes: Sure, that makes sense.

J535D165:
Thanks!!

PR #356, opened 2020-10-02
Make balance + query more modular

qubixes:
They can now be implemented through an extension under the 'asreview.balance_strategy' and 'asreview.query_strategy' entry points.


J535D165 (changes requested):
Very nice. Please add the functions to `reference.rst` (rebase first). A unit test would be nice as well.

qubixes:
Done.

J535D165 (approved):

PR #347, opened 2020-09-27
Add python linter (flake8)

J535D165:
This PR implements a Python linter. This should help enforcing the PEP8 code styling rules. 

PR #346 removes a lot of issues already, but there are many styling issues left to resolve.

PR #358, opened 2020-10-02
Remove PEP8 errors in tests

J535D165:
This PR removes the PEP8 errors in the tests module. Part of #347.


J535D165:
We're getting closer to merging #347

PR #364, opened 2020-10-06
Print warning if record ids are not unique.

qubixes:
Create new index in that case.


J535D165:
Good strategy. Can you make a unit test on this one as well?

qubixes:
I'm not sure what I would test honestly. Even before the PR, it should work. It's just that this is more neat / error prone if at some point we use the record_id's.

J535D165:
Test should show that a dataset with duplicate `record_id` values doesn't fail. Ideally, it shows somehow that it makes use of the index. It feels odd to me that it works at the moment with duplicate values btw.

qubixes:
That's the point; nothing really makes use of it at the moment. Prior_idx is kind of the only thing...

J535D165 (approved):

J535D165:
Thanks for the fix and test!

PR #366, opened 2020-10-07
Replace oracle entry point by lab entry point

J535D165:
This PR replaces the `asreview oracle` command by `asreview lab`. The first still works, but the latter is preferred. This should avoid confusion in the future with the oracle, exploration, simulation mode in the frontend. 

The PR also sorts the entry points. The subcommands at the top are 'lab', 'simulate', and 'simulate-batch'.

### DIFF
--- a/asreview/__main__.py
+++ b/asreview/__main__.py
@@ -23,9 +23,10 @@ from asreview.utils import get_entry_points
 
 PROG_DESCRIPTION = "Automated Systematic Review (ASReview)."
 
-# Entry points for internal use. These entry points are not displayed in the
-# help page of the  user interface.
+# Internal or deprecated entry points. These entry points
+# are not displayed in the help page of the  user interface.
 INTERNAL_ENTRY_POINTS = ["web_run_model"]
+DEPRECATED_ENTRY_POINTS = ["oracle"]
 
 
 def _sort_entry_points(entry_points):
@@ -54,6 +55,10 @@ def _output_available_entry_points(entry_points):
         if name in INTERNAL_ENTRY_POINTS:
             continue
 
+        # don't display the deprecated entry points
+        if name in DEPRECATED_ENTRY_POINTS:
+            continue
+
         # try to load entry points, hide when failing on loading
         try:
             description_list.append(entry.load()().format(name))
@@ -68,18 +73,10 @@ def main():
     entry_points = get_entry_points("asreview.entry_points")
 
     # Try to load the entry point if available.
-    if len(sys.argv) > 1 and (sys.argv[1] in entry_points or sys.argv[1] == "oracle"):
-
-        # first argument is the subcommand
-        subcommand = sys.argv[1]
-
-        # replace 'oracle' by 'lab'
-        if subcommand == "oracle":
-            print("Warning: subcommmand 'oracle' is replaced by 'lab'.")
-            subcommand = "lab"
+    if len(sys.argv) > 1 and sys.argv[1] in entry_points:
 
         try:
-            entry = entry_points[subcommand]
+            entry = entry_points[sys.argv[1]]
             entry.load()().execute(sys.argv[2:])
         except ModuleNotFoundError:
             raise ValueError(
--- a/asreview/__main__.py
+++ b/asreview/__main__.py
@@ -28,10 +28,27 @@ PROG_DESCRIPTION = "Automated Systematic Review (ASReview)."
 INTERNAL_ENTRY_POINTS = ["web_run_model"]
 
 
+def _sort_entry_points(entry_points):
+
+    entry_points_copy = entry_points.copy()
+
+    entry_points_sorted = {
+        "lab": entry_points_copy.pop("lab"),
+        "simulate": entry_points_copy.pop("simulate"),
+        "simulate-batch": entry_points_copy.pop("simulate-batch"),
+    }
+
+    entry_points_sorted.update(entry_points_copy)
+
+    return entry_points_sorted
+
+
 def _output_available_entry_points(entry_points):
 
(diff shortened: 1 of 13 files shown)


qubixes:
I agree with the changes in principle. I think it could be more elegant (I was thinking that oracle could be a subclass of the new lab entry point with the warning in there for example).

asreview/__main__.py
```diff
@@ -47,16 +64,22 @@ def _output_available_entry_points(entry_points):
...
     # Try to load the entry point if available.
-    if len(sys.argv) > 1 and sys.argv[1] in entry_points:
+    if len(sys.argv) > 1 and (sys.argv[1] in entry_points or sys.argv[1] == "oracle"):
+
+        # first argument is the subcommand
+        subcommand = sys.argv[1]
+
+        # replace 'oracle' by 'lab'
+        if subcommand == "oracle":
+            print("Warning: subcommmand 'oracle' is replaced by 'lab'.")
```
qubixes, line 78: logging.warning ?

J535D165:
Great feedback! Implemented all suggestions. Can you have a final check?

GerbrichFerdinands:
Looks good to me. 

Should "oracle" also be replaced by "lab" in the tests?

https://github.com/asreview/asreview/blob/2046f4340b15b7984fe17de57ea44cc24f7fa905/tests/test_simulate.py#L201-L209

qubixes:
@GerbrichFerdinands There is no more oracle mode testing... since more than a year? So, it can be removed altogether as an argument.

GerbrichFerdinands:
And even though oracle still works, I think this should also be updated to "lab" in the Docker file?

https://github.com/asreview/asreview/blob/2046f4340b15b7984fe17de57ea44cc24f7fa905/docker/asreview-lab/Dockerfile#L3

GerbrichFerdinands (changes requested):
Looks good to me! 
I am thinking, completely getting rid of 'batch' will break some of the scripts for HPC computing. Those scripts are also used for a couple of papers. Shouldn't we keep 'batch' just as we hold on to 'oracle' for now as well?

asreview/utils.py
```diff
@@ -176,15 +176,31 @@ def is_iterable(i):
 
 
 def list_model_names(entry_name="asreview.models"):
-    return [entry.name
-            for entry in pkg_resources.iter_entry_points(entry_name)]
+    return [*get_entry_points(entry_name)]
```
qubixes, line 179: Maybe list(x) is more obvious to read?

qubixes (changes requested):

qubixes:
@GerbrichFerdinands There is also an advantage in having a different name in that you can use the older version asreview-simulations along side it. The new version is different regardless, so it will break your/my HPC scripts anyway.

J535D165:
@GerbrichFerdinands Thanks for the great feedback! I removed the oracle reference in the tests and updated Docker.

asreview/entry_points/gui.py
```diff
@@ -10,17 +12,30 @@
...
-def _oracle_parser(prog="oracle", description=DESCRIPTION_ORACLE):
+# deprecated oracle class
+class OracleEntryPoint(LABEntryPoint):
+    description = "Graphical user interface for ASReview. (Deprecated)"
+
+    def _pre_execute(self):
+        logging.warning("Warning: subcommmand 'oracle' is replaced by 'lab'.")
+
+
+def _lab_parser(prog="lab", description=DESCRIPTION_ORACLE):
```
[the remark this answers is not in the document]
J535D165: Thanks

asreview/utils.py
J535D165: Hmm, I love this new syntax...

asreview/entry_points/gui.py
```diff
@@ -10,17 +12,30 @@
...
 
         main(argv)
 
 
-def _oracle_parser(prog="oracle", description=DESCRIPTION_ORACLE):
+# deprecated oracle class
+class OracleEntryPoint(LABEntryPoint):
+    description = "Graphical user interface for ASReview. (Deprecated)"
+
+    def _pre_execute(self):
```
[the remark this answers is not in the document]
J535D165: Some devs are complaining when super isn't in the first line. I will replace this with your suggestion. 

You might be right on the 'formerly oracle' thing. Let's remove that one as well.

asreview/entry_points/gui.py
```diff
@@ -1,26 +1,42 @@
...
 
 HOST_NAME = "localhost"
 PORT_NUMBER = 5000
 
 
-class GUIEntryPoint(BaseEntryPoint):
+class LABEntryPoint(BaseEntryPoint):
     description = "Graphical user interface for ASReview."
 
+    def _pre_execute(self):
```
qubixes, line 18: This can now be removed, right?

J535D165: Yes. Thanks. I was starving..

qubixes (approved):

asreview/entry_points/gui.py
qubixes: Yes, me too!

J535D165: Resolved. Successfully merged branch 'pizza' into master.

J535D165:
Thanks @qubixes and @GerbrichFerdinands for all the feedback!

Rensvandeschoot:
Great work!!

Can we have an error message whenever someone tries to start ASReview with the oracle statement? It might take some time before existing users are familiar with the new way of starting the software...

J535D165:
This change is backwards compatible. A deprecation warning is given as well. I'm not planning to remove this subcommand in the upcoming 6 months. The release of version 1 might be a good moment to remove it.

PR #369, opened 2020-10-09
Implemented basic simulation flow

emielvdveen:


asreview/webapp/src/PreReviewComponents/ProjectPageOracle .js
```diff
@@ -0,0 +1,303 @@
...
+                showExportResult={state.info.projectInitReady && !state.setup && !state.training}
+                toggleExportResult={props.toggleExportResult}
+                reviewFinished={state.finished}
+                finishProject={finishProject}
+              />
+              <DangerZone
+                project_id={props.project_id}
+                handleAppState={props.handleAppState}
+              />
+            </Box>
```
J535D165, line 287: This should also be available on the project page for simulation. They have to see the statistics and they should be able to export the project as well as the results.

asreview/webapp/src/PreReviewComponents/ProjectPageOracle .js
```diff
@@ -0,0 +1,303 @@
...
+  cardBox : {
+    paddingBottom: "24px",
+  },
+  stateElas : {
+    width: "100%",
+    maxWidth: "200px",
+    display: "block",
+    margin: "auto",
+  },
+}));
```
J535D165, line 78: Reuse with CSS file?

asreview/webapp/src/PreReviewComponents/PreReviewZone.js
```diff
@@ -69,6 +68,11 @@ const PreReviewZone = (props) => {
     ready: false
   });
 
+  const [project, setProject] = React.useState({
+    "id": null,
+    "mode": null
+  })  
+
```
J535D165, line 75: I'm a bit concerned about this. I like the idea of a project object. It might be a bit confusing to have this redefined here. It might be more intuitive to extend the Redux store.

J535D165 (changes requested):
Thanks for the PR. This is my preliminary review.

docs/source/cli.rst
```diff
@@ -78,3 +79,5 @@ The available parameters are:
 	                        Seed for setting the prior indices if the --prior_idx option is not used. If the option --prior_idx is used with one or more index, this option is ignored.
 	  --verbose VERBOSE, -v VERBOSE
 	                        Verbosity
+	  --completion_file COMPLETION_FILE
```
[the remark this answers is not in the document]
emielvdveen: My idea was keep the the responsibility to resolve paths in the `api.py`. I just changed the logic to save a file for every simulation run and store this in a folder `simulations` in the the project folder. For every simulation a UUID is generated and returned to the frontend. The folder structure will look something like this:

```
.asreview/
   <project_id>/
        data/
        simulations/
           <simulation_id>.json
           <simulation_id>.json
           <simulation_id>.json
```

Will push this code soon.

asreview/webapp/src/PreReviewComponents/ProjectPageOracle .js
emielvdveen: One question: is finishing a project needed in case of a simulation? And what would it mean?

Rensvandeschoot: I think a project is finished if the results are known... there is nothing more you can do once the simulation results are in. Maybe in the future we can add post-processing options of the data (for other graphics), but not in this stage and maybe I would prefer to do post-processing via R or Python scripts and not within ASReview.

emielvdveen: Oke, in that case I will remove the option to finish the project for a simulation.

docs/source/cli.rst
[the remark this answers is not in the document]
J535D165: Makes sense.

asreview/webapp/src/PreReviewComponents/ProjectPageOracle .js
emielvdveen: Fixed by reusing in ProjectPage.js

asreview/webapp/src/PreReviewComponents/PreReviewZone.js
emielvdveen: Removed the idea of a project state 'object'

J535D165 (approved):

PR #374, opened 2020-10-12
Remove abstract-only option

J535D165:
This PR proposes to remove the `abstract_only` option in ASReview. There is limited use and users can easily rename the columns to get things work. 

There is no deprecation plan, because the use of this function is very limited.


qubixes:
Personally, I think removing at this point might cause more trouble than it's worth for now. Perhaps just remove the command line option, and leave the rest in?

J535D165:
Any thoughts what is about to cause troubles. I can't find any references in the software at the moment. But I'm fine with removing the command line option first.

qubixes:
It might break loading of old review files, because of the ASReviewSettings. There might be other problems as well, but nothing I know from the top of my head.

qubixes:
Although it does seem to not give pytest errors, so it might just work?

J535D165:
Oke makes sense. I decided to revert the changes to the settings file. Other things are oke to remove I guess.

PR #379, opened 2020-10-14
Replace column definitions into more flexible structure

J535D165:
This PR replaces the stacked list for column specs into a dict. This makes it easier to extend.


J535D165:
All tests pass, but there is a problematic issue with the temporary pickle files in the front-end. This needs to be fixed before merging.

PR #386, opened 2020-10-20
New package structure for models

J535D165:
This PR proposes to changes the structure of the 'models' submodule. The 'models' submodule will become the main place for all model components. Former models are now 'classifiers'. 

All deprecated submodules will raise FutureWarnings.


J535D165:
I decided not to make the entry points backward compatible because this introduces quite some boilerplate code. I think there is limited use. It is important to update the template after the release https://github.com/asreview/template-extension-new-model.

PR #388, opened 2020-10-22
Add build dependencies error to troubleshooting

GerbrichFerdinands:
Since Python 3.9 was released, I've been getting quite some messages from new users that run into build dependency errors.  
This also happened when 3.8 was just out and will probably happen again. 

The troubleshooting section now: 

- explains what causes the `installing build depencies ... error`
- guides users to the Python installation guide on the ASReview website to either up- or downgrade their Python.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -129,3 +129,29 @@ First check whether the package is correctly installed. Do this with the command
 .. code-block:: bash
 
   python -m asreview oracle
+  
+
+Build dependencies error
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The command line returns the following message:
+
+.. code:: bash
+
+  "Installing build dependencies ... error"
+
+Some of the dependencies of ASReview are not compatible with your Python 
+installation. This error typically happens when the version of your Python 
+installation is either too old or too new. Then, the dependencies do not support 
+your version anymore, or yet.
+
+If the version of your Python installation has been released very recently, the 
+best thing to do is to install the second most recent version of Python
+instead. If your Python installation is outdated, you should install a Python 
+version that is 3.6 or higher. 
+
+Detailed step-by-step instructions to install 
+Python (and ASReview) are available for
+`Windows <https://asreview.nl/installation-guide-windows/>`__ and
+`MacOS <https://asreview.nl/installation-guide-mac/>`__ users.
+
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -140,18 +140,11 @@ The command line returns the following message:
 
   "Installing build dependencies ... error"
 
-Some of the dependencies of ASReview are not compatible with your Python 
-installation. This error typically happens when the version of your Python 
-installation is either too old or too new. Then, the dependencies do not support 
-your version anymore, or yet.
-
-If the version of your Python installation has been released very recently, the 
-best thing to do is to install the second most recent version of Python
-instead. If your Python installation is outdated, you should install a Python 
-version that is 3.6 or higher. 
-
-Detailed step-by-step instructions to install 
-Python (and ASReview) are available for
+This error typically happens when the version of your Python installation has been 
+released very recently. Because of this, of the dependencies of ASReview are not 
+compatible with your Python installation yet. The  best thing to do is to install 
+the second most recent version of Python instead. Detailed step-by-step instructions 
+to install Python (and ASReview) are available for
 `Windows <https://asreview.nl/installation-guide-windows/>`__ and
 `MacOS <https://asreview.nl/installation-guide-mac/>`__ users.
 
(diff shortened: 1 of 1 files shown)


qubixes:
Perhaps it is wise to recommend a version of Python in the instructions of the installation of it? Or is that already there? Since then users following the manual will not encounter the error at all.

GerbrichFerdinands:
@qubixes At the moment we recommend anything that's Python 3.6+. 
Would you rather recommend 3.8 then?

qubixes:
@GerbrichFerdinands I think that makes sense. We could say it is compatible with Python 3.6+, and recommend 3.8. In this way we don't have the problem again when 3.10 comes out. We can then later update the recommendation when the dependencies are packaged without the need for compiling it.

docs/source/installation.rst
```diff
@@ -129,3 +129,29 @@ First check whether the package is correctly installed. Do this with the command
...
+
+The command line returns the following message:
+
+.. code:: bash
+
+  "Installing build dependencies ... error"
+
+Some of the dependencies of ASReview are not compatible with your Python 
+installation. This error typically happens when the version of your Python 
+installation is either too old or too new. Then, the dependencies do not support
```
J535D165, line 145: If the Python executable is not supported because it is too old, you will install an older version of ASReview automatically. So no error will pop up. This happens:

```
jonathan$ pip install asreview
Collecting asreview
  Downloading https://files.pythonhosted.org/packages/32/4e/808054f480f4ebab50b7657d02af49b77d46726adf569c71431bafca9350/asreview-0.3.0.tar.gz (66kB)
    100% |████████████████████████████████| 71kB 882kB/s 
Requirement already satisfied: numpy in /Users/jonathan/.pyenv/versions/anaconda2-2018.12/lib/python2.7/site-packages (from asreview) (1.15.4)
```

I think this might be confusing, I will yank those very old versions and see what kind of error pops up in that case.

J535D165: The behavior is slightly different as expected (and described in https://www.python.org/dev/peps/pep-0592/ ?). 

```
jonathan$ pip install asreview
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support pip 21.0 will remove support for this functionality.
WARNING: The candidate selected for download or install is a yanked version: 'asreview' candidate (version 0.3.0 at https://files.pythonhosted.org/packages/32/4e/808054f480f4ebab50b7657d02af49b77d46726adf569c71431bafca9350/asreview-0.3.0.tar.gz#sha256=ddd5cce2313faed924835eb1cca1a7b84805f2024ce89aefa7b3cc73c5078b9c (from https://pypi.org/simple/asreview/))
Reason for being yanked: <none given>
Collecting asreview
  Using cached asreview-0.3.0.tar.gz (66 kB)
```

The warning is raised in yellow btw. Maybe we should delete those very old versions of ASReview from Pypi (<=0.3). From that moment on, ASReview requires at least Python 3.6.

J535D165: https://github.com/asreview/asreview/pull/133

J535D165: We decided to delete the <0.4 releases. These releases are very unstable and users with a Pytohn <3.6 install might not notice that there try to install asreview==0.3.0 instead of the latest version. The error message will be:
```
jonathan$ pip install asreview
ERROR: Could not find a version that satisfies the requirement asreview (from versions: none)
ERROR: No matching distribution found for asreview
```

J535D165: Happy to merge this PR after removing the part on outdated Python distributions.

J535D165 (changes requested):
Thanks a lot for this PR. Happy to merge after the remark is resolved. 

For reference. Numpy had a statement on this btw https://numpy.org/news/. 

> PYTHON 3.9 IS COMING, WHEN WILL NUMPY RELEASE BINARY WHEELS?
Sept 14, 2020 – Python 3.9 will be released in a few weeks. If you are an early adopter of Python versions, you may be dissapointed to find that NumPy (and other binary packages like SciPy) will not have binary wheels ready on the day of the release. It is a major effort to adapt the build infrastructure to a new Python version and it typically takes a few weeks for the packages to appear on PyPI and conda-forge. In preparation for this event, please make sure to update your pip to version 20.1 at least to support manylinux2010 and manylinux2014
use --only-binary=numpy or --only-binary=:all: to prevent pip from trying to build from source.
```

docs/source/installation.rst
GerbrichFerdinands: Thanks! Deleting the <0.4 releases sounds like a good approach.

J535D165 (approved):

J535D165:
Thanks!!

PR #389, opened 2020-10-23
Don't fail on repeated project imports

J535D165:
An exported project file can only be imported once. It will raise the following error when imported a second time. This PR gets rid of this error and imports the project with the suffix "-copy" (multiple times).

image

### DIFF
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -635,65 +635,77 @@ def api_init_model_ready(project_id):  # noqa: F401
 def api_import_project():
     """Import uploaded project"""
 
-    import_project_id = None
-
-    if 'file' in request.files:
-
-        project_file = request.files['file']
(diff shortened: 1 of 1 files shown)


terrymyc (approved):
Hi @J535D165 , thanks! This looks good to me. 
- One thing: suppose I upload the same file multiple times, then I may have a project named `ptsd-copy-copy-copy......`. There may be corresponding features that allow users to edit the project name, description, and user's name.
- Another thing: maybe users want to update the project file modified by collaborators. In this case, an option to replace the existed file while uploading may be straightforward. If users want to keep them both, then a copy stays.

J535D165:
Thanks for this feedback. I think you are right on both remarks. Maybe we have to start with an option to edit project metadata on the project page. Thereafter, we can have more advanced import options.

PR #392, opened 2020-10-28
Add option to read state file from ASReview project file

PeterLombaers:
Added function and changed existing functions to read .asreview files,.


asreview/state/utils.py
```diff
@@ -122,11 +127,43 @@ def state_from_file(data_fp):
...
-    if not Path(data_fp).suffix in STATE_EXTENSIONS:
-        logging.error(f"file {data_fp} does not end with {STATE_EXTENSIONS}.")
-        return None
+    if Path(data_fp).suffix == ".asreview":
+        base_state = state_from_dot_asreview(data_fp)
+    else:
+        if not Path(data_fp).suffix in STATE_EXTENSIONS:
+            logging.error(f"file {data_fp} does not end with {STATE_EXTENSIONS}.")
+            return None
+        base_state = _get_state_class(data_fp)(state_fp=data_fp, read_only=True)
```
J535D165, line 136: Not tested, but what about this? And raise ValueError instead of logging?

```
    if Path(data_fp).suffix == ".asreview":
        base_state = state_from_dot_asreview(data_fp)
    else if Path(data_fp).suffix in STATE_EXTENSIONS:
        base_state = _get_state_class(data_fp)(state_fp=data_fp, read_only=True)
    else:
        raise ValueError(f"Expected ASReview file or file {data_fp} with extension {STATE_EXTENSIONS}.")
```

asreview/state/utils.py
```diff
@@ -122,11 +127,43 @@ def state_from_file(data_fp):
...
     state = {
         os.path.basename(os.path.normpath(data_fp)):
-        _get_state_class(data_fp)(state_fp=data_fp, read_only=True)
+        base_state
     }
     return state
+
+
+def state_from_dot_asreview(data_fp):
```
J535D165, line 144: `state_from_asreview_file`? I think we should prefix this with `read_` for consistency. But that requires changes to the other functions in this file as well. 

Please add this one to `asreview.state.__init__.py` (can you add `state_from_file ` and `states_from_dir` as well?)

asreview/state/utils.py
```diff
@@ -122,11 +127,43 @@ def state_from_file(data_fp):
...
+        Path to .asreview file.
+
+    Returns
+    -------
+    BaseState:
+        The same type of state file as in the .asreview file, which at the moment
+        is JSONState.
+    """
+    if not Path(data_fp).suffix == '.asreview':
+        logging.error(f"file {data_fp} does not end with '.asreview'.")
```
J535D165, line 159: Raise ValueError instead of logging.

asreview/state/utils.py
J535D165: Please add the function to the documentation https://raw.githubusercontent.com/asreview/asreview/master/docs/source/reference.rst.

J535D165 (changes requested):
Great work! Thanks for this first contribution. 

Please see the remarks. Can you add a unit test on this?

asreview/state/utils.py
PeterLombaers: I could change it to ```read_asreview_file``` and the others to ```read_state_from_file```, ```read_states_from_dir```, but then they will have to be changed everywhere I guess. Also in visualization for example. Or do you only want to change the name of ```state_from_asreview_file```? That would be easier.

J535D165: Oké. Clear. We will work on the deprecation in another PR.

asreview/state/__init__.py
```diff
@@ -17,3 +17,6 @@
 from asreview.state.hdf5 import HDF5State
 from asreview.state.json import JSONState
 from asreview.state.utils import open_state
+from asreview.state.utils import states_from_dir
```
qubixes, line 20: These are generally only used internally I think, but it's fine either way with me.

qubixes (approved):
This part of the code still feels a little awkward to me (with the duplication), but that's more with the design. Either way, nice addition. :)

asreview/state/__init__.py
J535D165: I think extensions are easier to maintain if they make use of the public interface. That's why I requested this.

tests/test_state.py
```diff
@@ -27,6 +27,12 @@ def test_read_dict_state():
         assert isinstance(state, DictState)
 
 
+def test_read_asreview_file():
+    state_fp = Path("tests", "state_files", "test_1_inst.asreview")
+    with state_from_asreview_file(str(state_fp)) as state:
+        assert isinstance(state, JSONState)
```
J535D165, line 33: I think the test syntax should be (there is no context manager involved in `state_from_asreview_file`):

```
    with open_state(state_fp) as state:
        assert isinstance(state, JSONState)
```

J535D165:
> This part of the code still feels a little awkward to me (with the duplication), but that's more with the design. Either way, nice addition. :)

I think you are right. It makes more sense to have `open_state` as the (only?) public function in this submodule and deprecate(?) `state_from_file`. I think the return value of `state_from_file` should be the state, not the file path and state. 

Another thing: 

```
with open_state(state_fp) as state:
        assert isinstance(state, JSONState)
```

This one fails now. I think it should open in `read_only` mode. 

Maybe we should merge this PR first and make another PR to resolve some of these issues. Interested to work on this?

tests/test_state.py
J535D165: Hmm, maybe this is oke. Ideally, both situations should work. See the other comment as well. 

For example https://github.com/asreview/asreview/blob/cc7d15e99d00abc511880b9fb78a09d80633c8b2/asreview/analysis/analysis.py#L107 can make use of open_state, isn't it? Closing the state file will be safer.

qubixes:
I think the obvious thing is to call states_from_file from state_from_dir. I don't think there is a good reason not to. Apart from that I think it's "fine". Personally I would keep it as is.

PeterLombaers:
It seems logical to do everything with open_state and deprecate state_from_file. I wasn't really sure why there where two functions to access a state file, so I just left it as it was. In the analysis module (and in visualization as well) I think the filepath is used as an identifier, so that's why it is included in the output of state_from_file. 

This was the first time I wrote a unit test, so if there's something wrong there, just let me know.

The following from open_state seems a little bit strange as well. If the try part raises an exception, state will never be defined, so there will be an error in the finally section?

```
   try:
        state = state_class(state_fp=fp, *args, read_only=read_only, **kwargs)
        yield state
    finally:
        state.close()
```

J535D165:
Thanks for your thoughts on this. I'm postponing this PR to version 0.15 such that we can have a look at the duplicate functions (in another PR). 

The `try-finally` syntax seems to be correct. See context managers for more information on error 
 handling: https://book.pythontips.com/en/latest/context_managers.html.

PeterLombaers:
Stop me if I'm being too nitpicky, but I think there's a difference :)
```
@contextmanager
def open_file(name):
    f = open(name, 'w')
    try:
        yield f
    finally:
        f.close()
```
In this example ```f``` is defined before the try part. In our case, ```state``` is not defined before the try part, only ```state_class``` is. Not that it will ever matter, I guess, because the way ```state_class``` is defined will catch all errors already.

J535D165:
> Stop me if I'm being too nitpicky

I love it :) I think you are right. Thanks. At the moment, this might not be a problem. But this can become a problem when a new (database-based) state gets developed. So it would be nice to apply your suggestions. Can you open a new PR on this?

J535D165 (approved):

J535D165:
Thanks @PeterLombaers, really nice contribution. 

We are facing some issues with the visualization package and asreview project files https://github.com/asreview/asreview-visualization/issues/15. So we can't make beautiful plots with it right now, but hopefully, someone can resolve this :)

PR #401, opened 2020-11-06
Nan values should be converted to empty strings

qubixes:
@J535D165 You were trying to find something? :)


J535D165:
Haha, thanks :D I just pushed a similar solution in the other PR. I will merge this one first!

PR #403, opened 2020-11-09
Make project card fully clickable

### LINKED ISSUE #397: Make projects on project page fully clickable
**Describe the bug**
Hi J, user expect that you can open the project by clicking the project-tile-title. Is it possible to make this clickable with the link from “Open”
Some users where clicking on the Title, expecting that the project would open.

**To Reproduce**
1. Go to terminal and open ASR
2. When the browser started, en alas loaded the page
3. Go with you mouse to the project tile you already created
4. The cursor will change to a hand (what indicates it is clickable)
5. The user clicks, but has to go to the button "Open"

**Screenshots**
The title here has an "ahref"  but is not clickable. especially for mobile this would be very useful.
image

**Version information**
n.a.

**Additional context**
From an Usability Test we had some user who did this. It can be frustrating to click on something without a reaction. Do you have that as well?

J535D165:
Resolves #397

### DIFF
--- a/asreview/webapp/src/ProjectCard.js
+++ b/asreview/webapp/src/ProjectCard.js
@@ -41,12 +41,9 @@ const ProjectCard = (props) => {
 
   return (
     <Card className={classes.root}>
-      <CardActionArea>
-        {/*<CardMedia
-          className={classes.media}
-          image="/static/images/cards/contemplative-reptile.jpg"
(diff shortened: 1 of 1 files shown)


MartijnUX:
Love it! So simple, that the user will understand how it works. Great work @J535D165

PR #405, opened 2020-11-13
change name max sampling

Rensvandeschoot:
I propose to change the name 'max' in the front-end to 'certainty-based' so that it matches the paper and documentation. I did not change any of the labels, only the the wording in the dropdown menu.

### DIFF
--- a/asreview/webapp/src/PreReviewComponents/ProjectAlgorithms.js
+++ b/asreview/webapp/src/PreReviewComponents/ProjectAlgorithms.js
@@ -281,7 +281,7 @@ const ProjectAlgorithms = ({project_id, scrollToBottom}) => {
                             value="max"
                             color="default"
                           >
-                            {"Max (default)"}
+                            {"Certainty-based (default)"}
                           </MenuItem>
 
                           <MenuItem


Rensvandeschoot:
@terrymyc Is uncertainty-based also available?

Rensvandeschoot:
@terrymyc If you agree we should also update the text in the information pop-up (and we should delete the balancing strategy, because it cannot be changed in the front-end, right?).

J535D165 (approved):

Rensvandeschoot:
@terrymyc  / @J535D165  We should not provide the option for mixed sampling in the front-end, more research is needed what the optimal balance is for this option. So I would prefer three options: certainty-based, uncertainty-based and random.

terrymyc:
> @terrymyc If you agree we should also update the text in the information pop-up (and we should delete the balancing strategy, because it cannot be changed in the front-end, right?).

No, we should remove the balancing strategy mentioned in the help information.
@J535D165 Do we have an uncertainty-based strategy?

terrymyc (approved):

J535D165:
Internal functions still make use of the word Max. We can change this, but it doesn't look urgent to me. 

Thanks, and let's forward the other discussions to new PRs.

PR #410, opened 2020-11-16
Force record ids to be integers.

qubixes:
Right now floating point numbers as record id's or missing record id's might cause trouble.

### DIFF
--- a/asreview/io/utils.py
+++ b/asreview/io/utils.py
@@ -121,10 +121,11 @@ def standardize_dataframe(df, column_spec={}):
         else:
             try:
                 df['record_id'] = pd.to_numeric(df['record_id'])
+                df['record_id'] = df['record_id'].astype(int)
                 df.set_index('record_id', inplace=True)
             except ValueError:
-                logging.warning("Column 'record_id' has non-integer values. "
-                                "Continuing with new index.")
+                logging.warning("Column 'record_id' has non-integer or missing"
+                                " values. Continuing with new index.")
 
     # Create a new index if we haven't found it in the data.
     if df.index.name != "record_id":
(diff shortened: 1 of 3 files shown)


J535D165 (approved):

asreview/io/utils.py
```diff
@@ -121,10 +121,11 @@ def standardize_dataframe(df, column_spec={}):
         else:
             try:
                 df['record_id'] = pd.to_numeric(df['record_id'])
+                df['record_id'] = df['record_id'].astype(int)
```
J535D165, line 124: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_numeric.html#pandas.to_numeric 

Use downcast argument?

qubixes: I tried that, but it didn't work...

J535D165: Oke.

J535D165:
Can you explain why record_ids are converted to integers? Why not accept any unique, complete list of record_ids?

qubixes:
It seems more convenient than strings to me. But accepting them as strings is possible (but requires modification of the code).

J535D165:
For example, the CORD-19 database has MDF5 identifiers. I think we should not try to convert record ids into integers.

qubixes:
They will be ignored. I think there is a case to be made to use asreview_record_id, though that should be adjusted after this PR. (I don't think it matters all that much internally, but it might save some confusion.)

J535D165:
Oke, fine for me now. Will merge after a unit test is available (for both negative and missing record_ids in the `ASReviewData` class).

qubixes:
I have added the two missing unit tests. There is some commented out code that might be useful if we change the warning logs to actual warnings.

PR #411, opened 2020-11-16
More consistent record keeping.

qubixes:
Changes the behavior of ASReviewData.record() such that the record_id's will confirm to the value of by_index. This might resolve some issues in the backend.

### DIFF
--- a/asreview/data.py
+++ b/asreview/data.py
@@ -119,7 +119,7 @@ class ASReviewData():
         self.df = df
         self.data_name = data_name
         self.prior_idx = np.array([], dtype=int)
-        self._reverse_index = []
+        self._reverse_index = {}
         if df is None:
             self.column_spec = {}
             return
--- a/asreview/data.py
+++ b/asreview/data.py
@@ -119,6 +119,7 @@ class ASReviewData():
         self.df = df
         self.data_name = data_name
         self.prior_idx = np.array([], dtype=int)
+        self._reverse_index = []
         if df is None:
             self.column_spec = {}
             return
@@ -315,7 +316,8 @@ class ASReviewData():
                     **self.df.loc[j, :],
                     column_spec=self.column_spec,
                     record_id=j,
-                    by_index=False)
+                    record_by_id=j,
+                    record_by_index=self.reverse_index[j])
                 for j in index_list
             ]
         else:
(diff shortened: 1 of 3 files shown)


qubixes:
@J535D165 Can you see if this would resolve the issues that were seen?

asreview/io/paper_record.py
```diff
@@ -169,16 +169,16 @@ def body(self):
...
+        if self.included == LABEL_NA:
+            included = None
         paper_dict = {
             "title": self.title,
             "abstract": self.abstract,
             "authors": self.authors,
             "keywords": self.keywords,
             "record_id": self.record_id,
-            "label": label,
+            "included": included,
```
J535D165, line 181: This looks problematic. Do we know where this method is used?

asreview/data.py
```diff
@@ -311,15 +312,21 @@ def record(self, i, by_index=True):
...
+                for j in index_list
             ]
         else:
             records = [
                 PaperRecord(**self.df.iloc[j],
                             column_spec=self.column_spec,
-                            record_id=self.df.index.values[j])
+                            record_id=j,
+                            record_by_id=self.df.index.values[j],
+                            record_by_index=j)
```
J535D165, line 329: What is the aim of passing these values?

asreview/io/paper_record.py
qubixes: Apparently nowhere, since it was completely broken..

asreview/data.py
qubixes: The point it that the user of the function can choose which kind of record_id it wants to use. Sometimes you need the internal one (record_by_index), sometimes the external one (record_by_id). And I can imagine you might not really can and you can use the default one (record_id).

GerbrichFerdinands:
I believe this PR possibly solves the mix-up between record_id's and row indices when using `--prior_idx`. 
I will run a test asap to see if this is the case.

asreview/io/paper_record.py
J535D165: Confirmed. Please remove the method.

asreview/data.py
```diff
@@ -466,6 +473,13 @@ def get(self, name):
         except KeyError:
             return self.df[name].values
 
+    @property
+    def reverse_index(self):
+        if len(self._reverse_index) != len(self):
+            self._reverse_index = {
+                self.df.index.values[i]: i for i in range(len(self))}
+        return self._reverse_index
+
```
J535D165, line 482: Why do you cache this function if it is single-use?

qubixes: Otherwise multiple calls to ASReviewData.record(by_index=False) will be very slow.

asreview/io/paper_record.py
qubixes: Done

Issue #360, opened 2020-10-05
Error while training model after re-uploading dataset


laurienmeijer:
Hi ASReview team,

After using ASReview for a while I realized there were still many duplicate records in my dataset that weren't recognized by my reference management software. I didn't want to lose my progress so I downloaded the Excel spreadsheet of results, manually removed the duplicates, and uploaded it again (in a new project). I named the column that indicates which records I had previously labeled (ir)relevant 'included', as recommended in the FAQ.

Uploading the file works, and ASReview correctly recognizes the 'included' column, displaying the correct numbers under 'prior knowledge'. I set the active learning model to the default settings (Classifier:
Naïve Bayes; Query strategy: Max; Feature extraction: tf-idf).

Clicking 'Finish' results in an endless loading icon on 'training model' and the following error message:

_An error occured. Please send an email to asreview@uu.nl or file an issue on GitHub.
Error message
index 6323 is out of bounds for axis 0 with size 6077_

(I have tried it a couple of times and the 'index' number sometimes changes, but is always greater than 6077, which is the number of records in my dataset).

If anyone knows how to solve this, it would be much appreciated!

Thank you in advance,

Kind regards,
Laurien Meijer

Rensvandeschoot:
Dear @laurienmeijer 

Many thanks for reporting this issue! 

Could you share your file with us via asreview@uu.nl so we can run some tests to see what went wrong? 

We will come back to you as soon as possible (since you are probably in the middle of your research)

best
Rens

laurienmeijer:
Sent, thank you in advance!

Rensvandeschoot:
Great!
It appears that if you delete the column titled 'record_id' all works fine. If you do so, you can continue with screening your papers. We will leave this issue open till we found a better solution.

thanks again for letting us know!
best
rens

J535D165:
For the sake of completeness, the issue was resolved by deleting the `record_id` column. It seems like there is a conflict between the use of `record_id`'s and row numbers. This seems to be a bug. 

I'm linking issue https://github.com/asreview/asreview/pull/353.

GerbrichFerdinands:
This might be related to a feature/bug in simulation mode: 

 @qubixes mentioned in #353:

> "List of record ids used as the initial inclusions/exclusions for the simulation. In the presence of a column "record_id", these will be used for this purpose. Otherwise they are sequential in the order that records occur, starting from 0. For example, in a csv/xlsx file the first row with data will have a record_id of 0, etc."

[...]

PR #415, opened 2020-11-18
Internal indices should be all index based.

qubixes:
There is in issue where record_ids are passed around, but some functions assume that they are by row number.

My suggestion is to have everything "internal" passed around as row numbers, mainly because I find that surface the easiest to remember/explain.

Either way, this is quite a big issue, since in particular the export function assumes row numbers instead of record_ids. (I guess it's a bit more complicated than that, but that's the easiest way to explain it now.) The problem is that the labels are mixed up if the record_id's are not lined up.

I think this is part of the solution, but there could be more that has to be changed. This issue probably has been there since the first time ASReview had a GUI.

My estimation right now:

Unaffected datasets:
- No column with record ids
- RIS files
- Record id column, containing non-numerical values (i.e. have non-numeric characters).
- Record id column, with duplicates values (or empty)
- Record id column, with values ranging from 0...N_row - 1 (in any order)

Affected datasets:
- Record id column with unique values that can be converted to numerical values and do not range from 0...N_row - 1

What will the user notice:
- If the values are far from 0...N_row-1, the user will almost certainly quickly notice a problem like #360 
- If values are close to 0...N_row-1, say i_row + 1, the user might not notice anything, except that when stopping the export will have the labels assigned to the wrong record_ids.
- If record ids fall between -N_row-1 and N_row-1, there might not be a crash at all even after reviewing all records. This case seems very rare though.

What can happen during the usage of ASReview that makes an unaffected dataset into an affected dataset:
- The user starts a review, exports to csv, deletes some records, and uploads it again. Depending on the number of deleted records, the user might sooner or later run into #360.
- Instead of deleting records, the user adds records with valid record_ids, but not exactly sequential to the ones generated by ASReview. This seems rather unlikely.

### DIFF
--- a/asreview/webapp/utils/project.py
+++ b/asreview/webapp/utils/project.py
@@ -111,7 +111,7 @@ def add_dataset_to_project(project_id, file_name):
 
         if as_data.labels is not None:
             unlabeled = np.where(as_data.labels == LABEL_NA)[0]
-            pool_indices = as_data.record_ids[unlabeled]
+            pool_indices = unlabeled
 
             label_indices_included = \
                 [[int(x), 1] for x in np.where(as_data.labels == 1)[0]]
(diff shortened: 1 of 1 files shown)


J535D165:
Resolved in #418. This PR provides a nice overview of affected projects.

PR #418, opened 2020-11-26
Resolve bugs with record_ids

J535D165:
This PR resolves a bug with record_ids, adds inline documentation, and add tests.


asreview/io/utils.py
```diff
@@ -130,5 +130,5 @@ def standardize_dataframe(df, column_spec={}):
     if df.index.name != "record_id":
         df["record_id"] = np.arange(len(df.index))
         df.set_index('record_id', inplace=True)
-    df.sort_index(inplace=True)
+    # df.sort_index(inplace=True)
```
J535D165, line 133: This might be problematic for existing reviews with an unordered record_id column.

J535D165:
See #415 for more details on affected projects.

asreview/io/utils.py
Rensvandeschoot: Can we create a check to find out if an existing record_id is correctly ordered and has the same length as row_numbers? If this is the case, then it is not problematic, if the check fails we can provide a warning?

J535D165: This is already implemented. However, it will create a new record_id based on the row number. I will replace this with an error in another PR (explicit over implicit, zen op Python).

PR #427, opened 2020-12-02
info added about the ordering of the export file

Rensvandeschoot:
@J535D165 is record_id also added to the export file? if so, we need to add this to the documentation as well


J535D165:
Yes, it is.

Rensvandeschoot:
@J535D165 and just to be sure: record_id is row number of the data as uploaded and starting at zero

J535D165:
If present, the variable `record_id` contains the values of the `record_id` column. Otherwise, ASReview generates a new record_id, based on the row number and starting at 0.

J535D165 (approved):

J535D165:
Thanks, nice contribution!

PR #430, opened 2020-12-03
update text quicktour to match vocabular

Rensvandeschoot:


asreview/webapp/src/Components/QuickTourDialog.js
```diff
@@ -39,38 +39,38 @@ const mapStateToProps = state => {
...
     imgPath: SetUp,
     textTitle: "Easy project setup",
-    text: "Supply a dataset, select prior knowledge, and choose a machine learning model (optional).",
+    text: "Supply a clean dataset, select prior knowledge, and choose a machine learning model (optional).",
   },
   {
     imgPath: Start,
-    textTitle: "Screen for relevant publications",
-    text: "Read the displayed publication and decide whether it is relevant or not.",
+    textTitle: "Mark text as relevant",
```
GerbrichFerdinands, line 52: Shouldn't this be
relevant --> (ir)relevant 
?

GerbrichFerdinands (commented):
I like it! 

According to our vocabulary, both records and texts are correct words here, but maybe for consistency, we should pick one of them?

J535D165:
Great changes! Thanks @Rensvandeschoot

PR #431, opened 2020-12-04
Poorly defined record_ids now raise an error

J535D165:
This PR remove the implicit record_id handling on datasets with poorly defined record_id values. 

In case the user has a poorly defined record_id column it should be renamed or dropped.

### DIFF
--- a/asreview/io/utils.py
+++ b/asreview/io/utils.py
@@ -63,8 +63,28 @@ def type_from_column_spec(col_name, column_spec):
     return None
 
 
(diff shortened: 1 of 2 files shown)


J535D165:
We have to study the behavior of the Datasets card in the frontend. This card might return a "provide a title and abstract error instead of the informative record_id error". Will solve this in another PR.

J535D165:
Turns out not to be a problem. 

image

PR #434, opened 2020-12-09
Only count labels '1' in statistics.py.

PeterLombaers:
Only count labels '1' in statistics.py. This should fix https://github.com/asreview/asreview-visualization/issues/15


J535D165:
Very nice fix. No changes for the extension.

PR #439, opened 2020-12-14
improve text for searching prior knowledge

Rensvandeschoot:
This PR changes the text when searching for prior knowledge so that it is more clear this is not yet the screening phase.


asreview/webapp/src/PreReviewComponents/ListItemPaper.js
```diff
@@ -85,7 +85,7 @@ const ListItemPaper = (props) => {
         fullWidth={true}
       >
         <DialogTitleWithClose
-          title={"Is this article relevant?"}
+          title={"If this document can be used as prior knowledge, is it relevant or irrelevant?"}
```
Rensvandeschoot, line 88: @J535D165 What is the maximum of characters I can use?

terrymyc: image

It looks too long as a title...

Rensvandeschoot: @MartijnUX Do you have suggestions?

J535D165:
Thanks, hopefully, this resolves the confusion for some users!

PR #446, opened 2020-12-30
Benchmark platform integration

J535D165:
This PR integrates the datasets of https://github.com/asreview/systematic-review-datasets in the ASReview software. Currently only available for simulation mode. 

Usage:

- Select dataset by name. This is the name of the metadata file (the JSON file). For example [Bos_2018](https://github.com/asreview/systematic-review-datasets/tree/metadata/datasets/Bos_2018).
- Prefix with  benchmark: `benchmark:Bos_2018`
- run simulation:

```
$ python -m asreview simulate benchmark:Bos_2018

            _____ _____            _
     /\    / ____|  __ \          (_)
    /  \  | (___ | |__) |_____   ___  _____      __
   / /\ \  \___ \|  _  // _ \ \ / / |/ _ \ \ /\ / /
  / ____ \ ____) | | \ \  __/\ V /| |  __/\ V  V /
 /_/    \_\_____/|_|  \_\___| \_/ |_|\___| \_/\_/

---------------------------------------------------------------------------------
|                                                                                |
|  Welcome to the ASReview Automated Systematic Review software.                 |
|  In this mode the computer will simulate how well the ASReview software        |
|  could have accelerate the systematic review of your dataset.                  |
|  You can sit back and relax while the computer runs this simulation.           |
|                                                                                |
|  GitHub page:        https://github.com/asreview/asreview                      |
|  Questions/remarks:  asreview@uu.nl                                            |
|                                                                                |
---------------------------------------------------------------------------------

Labeled during review:

3519 => 1
2096 => 0

The following records are prior knowledge:

3519 - Long-Term Clinical Impact of Vascular Brain Lesions on Magnetic Resonance Imag..   Kaffashian, S.;Soumare, A.;Zhu, Y. C.;..
2096 - A SPECT imaging study of MRI white matter hyperintensity in patients with dege..   Ott, B. R.;Faberman, R. S.;Noto, R. B...

```

## TODO
- [x] Update URL after https://github.com/asreview/systematic-review-datasets/pull/36 is merged
- [ ] Documentation (maybe we can wait with his one because more updates on this are upcoming)

## Question

@Rensvandeschoot what is a good name for this integration? `benchmark` or should we use something else? (In the end, this might replace the example datasets tab in LAB, so the name should fit that tab as well)

### DIFF
--- a/asreview/datasets.py
+++ b/asreview/datasets.py
@@ -413,5 +413,23 @@ class BuiltinDataGroup(BaseDataGroup):
         )
 
 
+class BenchmarkDataGroup(BaseDataGroup):
+    group_id = "benchmark"
+    description = "A collections of fully labeled datasets for benchmarking."
+
+    def __init__(self):
+        meta_file = "https://raw.githubusercontent.com/asreview/systematic-review-datasets/master/config/index.json"  # noqa
+        with urlopen(meta_file) as f:
+            meta_data = json.loads(f.read().decode())
+
+        datasets = []
+        for data in meta_data.values():
+            datasets.append(_create_dataset_from_meta(data))
+
+        super(BenchmarkDataGroup, self).__init__(
+            *datasets
+        )
+
+
 def find_data(project_id):
     return DatasetManager().find(project_id).get()
--- a/asreview/datasets.py
+++ b/asreview/datasets.py
@@ -35,7 +35,7 @@ def _create_dataset_from_meta(data):
             datasets.append(BaseDataSet.from_config(config))
         return BaseVersionedDataSet(base_dataset_id, datasets, title)
     elif data["type"] == "base":
-        return BaseVersionedDataSet.from_config(data["configs"][0])
+        return BaseDataSet.from_config(data)
     else:
         raise ValueError(f"Dataset type {data['type']} unknown.")
 
@@ -418,7 +418,7 @@ class BenchmarkDataGroup(BaseDataGroup):
     description = "A collections of fully labeled datasets for benchmarking."
 
     def __init__(self):
-        meta_file = "https://raw.githubusercontent.com/asreview/systematic-review-datasets/master/config/index.json"  # noqa
+        meta_file = "https://raw.githubusercontent.com/asreview/systematic-review-datasets/metadata/index.json"  # noqa
         with urlopen(meta_file) as f:
             meta_data = json.loads(f.read().decode())
 
--- a/asreview/datasets.py
+++ b/asreview/datasets.py
@@ -415,7 +415,7 @@ class BuiltinDataGroup(BaseDataGroup):
 
 class BenchmarkDataGroup(BaseDataGroup):
     group_id = "benchmark"
-    description = "A collections of fully labeled datasets for benchmarking."
+    description = "A collections of labeled datasets for benchmarking."
 
     def __init__(self):
         meta_file = "https://raw.githubusercontent.com/asreview/systematic-review-datasets/metadata/index.json"  # noqa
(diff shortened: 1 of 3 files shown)


asreview/datasets.py
```diff
@@ -413,5 +413,23 @@ def __init__(self):
         )
 
 
+class BenchmarkDataGroup(BaseDataGroup):
+    group_id = "benchmark"
+    description = "A collections of fully labeled datasets for benchmarking."
```
Rensvandeschoot, line 418: Rephrase: "A collection of labelled datasets for benchmark testing" 

Do the data need to be fully labelled? The simulation also runs for partly labelled data and the unlabelled data are then just ignored. So, is the "fully labelled" needed for the new benchmark integration?

setup.py
```diff
@@ -143,6 +143,7 @@ def get_cmdclass():
         ],
         'asreview.datasets': [
             'builtin = asreview.datasets:BuiltinDataGroup',
+            'benchmark = asreview.datasets:BenchmarkDataGroup',
```
Rensvandeschoot, line 146: Will there be a difference between the built-in datasets and the benchmarkDataGroup? I can image we keep some datasets available in the front-end for workshops, but I can also image we simplify the code and keep only the latter group.

tests/test_data.py
```diff
@@ -34,6 +34,8 @@ def test_fuzzy_finder(keywords, paper_id):
     "ptsd",
     "ace",
     "hall",
+    "benchmark:Bos_2018",
+    "benchmark:van_de_Schoot_2017"
```
Rensvandeschoot, line 38: Do we need to add more of such data for testing purposes?

Rensvandeschoot (changes requested):

asreview/datasets.py
J535D165: Good question. It depends on what you put in here: https://github.com/asreview/systematic-review-datasets. At the moment, we do have fully labeled datasets only. We should have a good reason when mixing partly and fully labeled datasets over there (however can be a good idea). We need a metadata standard for those datasets and make this part of that metadata. The current metadata version is only a draft: https://github.com/asreview/systematic-review-datasets/blob/master/datasets/van_de_Schoot_2017/van_de_Schoot_2017.json. I know that @bmkramer and Otto do have some ideas/experiences on this.

setup.py
J535D165: Good one. I'm deprecating the built-in group. I was thinking about having a featured subset in LAB with the datasets we currently use a lot. In this way, we resolve both.

tests/test_data.py
J535D165: Maybe we should test all the featured datasets.

setup.py
Rensvandeschoot: O yes, that is a perfect idea!

asreview/datasets.py
Rensvandeschoot: Ok, thanks for explaining! So, in the text we can drop the word 'fully' because it is not a technical requirement. 

I am working on some partly labelled datasets we can add in a later stage.

J535D165: Oké. I will update this. Please keep in mind that for Exploration mode this is a requirement. So, please remember this when adding the partly labeled datasets. That will require a implemented heuristic that's not implemented yet.

Rensvandeschoot: Ok! Is there a test for the exploration mode to test if the column with labels contains no missing values, and if so, it should provide an error message

J535D165: This is to be implemented. Let's forward this check to the next release (the release that implements the new exploration mode and simulation mode).

PR #448, opened 2020-12-30
Add benchmark datasets to LAB

J535D165:
This is the draft of the benchmark dataset integration in the frontend. Feel free to post some first feedback (functionality had priority so far). 

localhost_3000_ (8)


Rensvandeschoot:
Nice!!! 

How do you select which art-work to display for new data?

@MartijnUX do you have ideas for this?

asreview/datasets.py
```diff
@@ -413,5 +419,36 @@ def __init__(self):
         )
 
 
+class BenchmarkDataGroup(BaseDataGroup):
+    group_id = "benchmark"
+    description = "A collections of fully labeled datasets for benchmarking."
```
Rensvandeschoot, line 424: "A collection of labelled datasets for benchmark testing"

Rensvandeschoot:
@J535D165 in what order are the datasets displayed?

J535D165:
Order: [index.csv](https://github.com/asreview/systematic-review-datasets/blob/master/index.csv). The screenshot is outdated.

The next step should be a redesign of the cards I think. The variables in [index.csv](https://github.com/asreview/systematic-review-datasets/blob/master/index.csv) are now available for these cards and some of them are interesting for the user (number of records, link to the publication, topic, year, license, ...). 
Besides that, the card-in-card design is a violation of Material I think. However it's hard to find a good alternative. Maybe we should start with the expandable card https://material.io/components/cards#behavior for adding more info.

Artwork selection: https://github.com/asreview/systematic-review-datasets/blob/150dc12eb0f829ebd6f5f02709a06f914a84e4fc/datasets/van_de_Schoot_2017/van_de_Schoot_2017.json#L14. I'm working on a representation for the empty ones.

Rensvandeschoot:
Thanks for you elaborate answer! 

I agree with using expandable cards, it would help to see some meta-data, but only if you are interested in seeing it.

Also, we should think how users can search for specific datasets and how to structure the datasets if we have many more in the repository, let's say 200...

J535D165:
Please drop your thoughts on interesting variables to display. See the index.html for variables currently available. I will leave this up to you (except from including the license ;)). Hopefully @terrymyc and @MartijnUX can implement the expandable card.

J535D165:
Implemented the Featured datasets. See the new screenshot. 

I'm not happy with the cards, hard to fit the title in there. The images are a lot of work to maintain. Maybe we should use them for featured datasets only. Maybe we should consider the accordion component https://material-ui.com/components/accordion/. The last example with buttons is nice imo https://material-ui.com/components/accordion/#secondary-heading-and-columns. We can use "Select dataset" over there.

localhost_3000_ (9)

MartijnUX:
I totally agree that the cards are not easy to scan for the user. With a big list of options I prefer the List pattern. Even search/sort/filter would be a nice feature for a bit amount of data(sets). Let me know if you need a prototype!

J535D165:
I think we need a prototype. I don't have a clue how to implement this in a List component (the expandable part). I'm also fine with a popup with the information. @LHofstee and @Rensvandeschoot can provide you the variables to display. It will be a subset of the variables you will find here: https://github.com/asreview/systematic-review-datasets/blob/master/index.csv. 

I would like to ask @terrymyc to implement the prototype.

asreview/webapp/api.py
```diff
@@ -262,9 +262,28 @@ def api_demo_data_project():  # noqa: F401
...
+            include="benchmark"
+        )
+
+        # mark the featured datasets
+        featured_dataset_ids = [
+            "van_de_Schoot_2017",
+            "Hall_2012",
+            "Cohen_2006_ACEInhibitors",
+            "Bos_2018"
+        ]
```
J535D165, line 281: Featured datasets can be specified here @Rensvandeschoot

J535D165:
With the help of @Rensvandeschoot and @LHofstee, we updated the metadata and the representation in the frontend. Thanks

localhost_3000_ (13)

Rensvandeschoot:
Nice! This is much cleaner!!

To do: 
- list instead of pictures (except featured datasets) 
- option to order datasets on authors or topic -> @MartijnUX / @terrymyc 

- more details metadata -> @LHofstee / @Rensvandeschoot

PR #453, opened 2021-01-06
improve test compatibility and added error in Distillor

Rensvandeschoot:
This PR fixes some typo's and clarifies how to import data from search engines and other software packages.


docs/source/intro/datasets.rst
```diff
@@ -126,27 +126,27 @@ Citation Managers
...
+| **Refworks**    | Supported     |              |              |
+|                 |               |              |              |
++-----------------+---------------+--------------+--------------+
+| **Zotero**      | Supported     |  Supported   |              |
+|                 |               |              |              |
++-----------------+---------------+--------------+--------------+
+
+- Supported: The data can be exported from the citation manager and imported
+in ASReview.  
+- Empty: This format does not exist.
```
J535D165, line 149: Are you fine with changing this into `This export format is not available. `

docs/source/intro/datasets.rst
```diff
@@ -224,15 +227,16 @@ compatibility with ASReview.
 
 - Supported: The data can be exported from the software and imported in ASReview using this extension.
 - Not supported: The exported data can not be imported in ASReview using this extension.
-- (empty): The data cannot be exported from the software using this extension.
+- Empty: The extension format cannot be exported from the search engine.
```
J535D165, line 230: These aren't search engines. Maybe use the same phrase as above.

docs/source/intro/datasets.rst
```diff
@@ -224,15 +227,16 @@ compatibility with ASReview.
...
 
-\* When using Covidence it is possible to export articles in .ris formats for different citation managers,
+\* When using Covidence it is possible to export articles in ``.ris`` format for different citation managers,
```
J535D165, line 232: According to the experts https://developers.google.com/style/filenames?hl=en#file-type-names, it is preferred to say `RIS file format`. I did apply this rule in several places.

docs/source/intro/datasets.rst
```diff
@@ -224,15 +227,16 @@ compatibility with ASReview.
...
 such as Endnote, Mendeley, Refworks and Zotero. All of these are compatible with ASReview.
 
-\** When exporting from Distiller set the ``sort references by`` to ``Authors``. Then the data can be
-imported in ASReview.
+\** When exporting from Distiller and if the following error occurs ``Unable
+to parse string "Yes (include)" at position 0`` set the ``sort references by``
+to ``Authors``. Then the data can be imported in ASReview.
 
-\*** Robotreviewer does not provide exports suitable for asreview, since it supports evidence synthesis.
+\*** Robotreviewer does not provide exports suitable for asreview.
```
J535D165, line 239: Remove this line. The table says it all.

docs/source/intro/datasets.rst
```diff
@@ -126,27 +126,27 @@ Citation Managers
...
+|                 |               |              |              |
++-----------------+---------------+--------------+--------------+
+| **Refworks**    | Supported     |              |              |
+|                 |               |              |              |
++-----------------+---------------+--------------+--------------+
+| **Zotero**      | Supported     |  Supported   |              |
+|                 |               |              |              |
++-----------------+---------------+--------------+--------------+
+
+- Supported: The data can be exported from the citation manager and imported
```
J535D165, line 147: Nice.

docs/source/intro/datasets.rst
Rensvandeschoot: no, because then it seems it is not available in ASReview, but the formats simply do not exist in the citation managers

Rensvandeschoot:
@J535D165 thank you for your feedback! I will update the files after @SagevdBrand is done with her review.

docs/source/intro/datasets.rst
J535D165: This is the main reason why there was a Not Supported option. The confusing thing is, we support them all in this table ;) 

With the proposed changes, the empty option has different descriptions/meanings(?) across the page. I'm fine with changing the text somehow, but I prefer to have identical descriptions.

Rensvandeschoot: I have added  the following to the tables:

-  ✅ = The data can be exported from the citation manager and imported in ASReview. 
-  N/A = This format does not exist.
-  X = not suported.

PR #460, opened 2021-01-09
Update icons for relevant and irrelevant records

Rensvandeschoot:
I think the 'X' might be confusing because it is typically used for closing something. Therefore, I propose to use a thumb-down icon.

### DIFF
--- a/asreview/webapp/src/Components/HistoryDialog.js
+++ b/asreview/webapp/src/Components/HistoryDialog.js
@@ -209,7 +209,7 @@ const HistoryDialog = (props) => {
               <ArrowBackIcon />
             </IconButton>
             <div className={classes.detailTitle}>
-              Overview
+              Details
             </div>
           </DialogTitle>
         }
--- a/asreview/webapp/src/Components/HistoryDialog.js
+++ b/asreview/webapp/src/Components/HistoryDialog.js
@@ -18,7 +18,7 @@ import { makeStyles } from '@material-ui/core/styles';
 
 import ArrowBackIcon from '@material-ui/icons/ArrowBack';
 import FavoriteIcon from '@material-ui/icons/Favorite';
-import CloseIcon from '@material-ui/icons/Close';
+import ThumbDownIcon from '@material-ui/icons/ThumbDown';
 
 import {
   HistoryListCard,
@@ -209,7 +209,7 @@ const HistoryDialog = (props) => {
               <ArrowBackIcon />
             </IconButton>
             <div className={classes.detailTitle}>
-              Details
+              Overview
             </div>
           </DialogTitle>
         }
@@ -313,7 +313,7 @@ const HistoryDialog = (props) => {
               <Typography className={classes.recordLabelActionText}>
                 {openIndex.record.included === 1 ? "Relevant" : "Irrelevant"}
               </Typography>
-              {openIndex.record.included === 1 ? <FavoriteIcon color="secondary" /> : <CloseIcon/>}
+              {openIndex.record.included === 1 ? <FavoriteIcon color="secondary" /> : <ThumbDownIcon/>}
             </Box>
           }
           {openIndex["index"] === null &&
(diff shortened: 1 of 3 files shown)


asreview/webapp/src/Components/HistoryDialog.js
```diff
@@ -18,7 +18,7 @@ import { makeStyles } from '@material-ui/core/styles';
 
 import ArrowBackIcon from '@material-ui/icons/ArrowBack';
 import FavoriteIcon from '@material-ui/icons/Favorite';
-import CloseIcon from '@material-ui/icons/Close';
+import ThumbDownIcon from '@material-ui/icons/ThumbDown';
```
J535D165, line 21: Interesting idea! 

If we are changing it here, then we should also change it in the review screen (and Prior Knowledge screen?).

Rensvandeschoot: I think it makes sense...

terrymyc: I'll make a screenshot for preview!

Rensvandeschoot:
I changed the heading of the pop-up screen back to Details and i will open a separate PR for it.

terrymyc:
image

image

Rensvandeschoot:
o nice! looking good!

@J535D165 I propose to integrate this in a future release (v0.17 / 18?) because we have to also adjust the documentation

J535D165:
This requires quite some updates on the documentation. Please use this PR for documentation updates.

Rensvandeschoot:
@J535D165 I changed the icon in the decision bar. It is not used in the prior knowledge phase.
@terrymyc I also update the icon in `HistoryListCard.js`

Rensvandeschoot:
New screenshots are need on lines 33/34 + 111/112: https://github.com/asreview/asreview/blob/master/docs/source/features/screening.rst

and lines 88/89
https://github.com/asreview/asreview/blob/master/docs/source/lab/oracle.rst

MartijnUX:
We could consider to change the hart as well to a finger up icon. In that way it is clear that it is a negative/positive. See example.
image

MartijnUX:
@terrymyc Is it possible to put the icon before the text? See example 
image

Rensvandeschoot:
I am fine with changing the hart in a thumbs-up. I think the colors should indeed be green and gray, like above.

terrymyc:
> Is it possible to put the icon before the text? See example

@MartijnUX Yes, the icon can be put before.

Rensvandeschoot:
@terrymyc I have added the 'thumbsup' icon, can you see if you can make it green? and update the order as requested by @MartijnUX? many thanks!

J535D165:
After seeing the updates and screenshots, I'm not directly convinced about this change. On the other hand, this PR convinces me that we should consider changes. It's a great discussion and team effort. 

The thumbs-up and thumbs-down approach gives it a bit of a like-dislike feeling. As with social media. That's not in line with the things we do. It also doesn't translate to relevant or irrelevant for me. Not a big problem when labels (relevant and irrelevant) are available. However, it might be confusing when used standalone. Do you like the record, or do you want to mark the item as relevant. 

My main concern is the extensibility. We are planning/ thinking about options like skipping, postponing, and multilabel decisions. The thumbs-up and thumbs-down gestures are hard to extend with something like skipping. I think we should also think about the colors. It makes sense to make relevant green, however, it makes more sense to make irrelevant records red. This leaves grey records for postponing/unlabeled/skipping for example. 

Material turns out to have a component that might fit our purpose. It is named Chip https://material-ui.com/components/chips/#chip. You find this component in your email box for tags, or when filling in keywords when submitting a paper for example.

The following example shows how a chip can look like. I decided to use letters instead of icons. 

Screenshot 2021-01-11 at 19 55 48
Screenshot 2021-01-11 at 19 54 22

Bonus

image

I'm still a bit struggling with the overview page. The following page is NOT good imo, but I'm sharing to facilitate the discussion. 

image

terrymyc:
@J535D165 I like the outlined chips! Maybe we went too far with the `thumb` idea. The concern in the first place was a cross may be confusing on the details (overview) page. I suggest replacing the cross and heart with chips only on the left bottom of the overview page.

Rensvandeschoot:
ok! let's work on this for V1.0

PR #462, opened 2021-01-10
title pop-up history overview

Rensvandeschoot:
I propose to change the title of the pop-up screen from 'details' into 'overview'. This is because clicking the button brings you back to the overview page and not not more details about the record.


terrymyc (approved):

J535D165:
Thanks!

PR #463, opened 2021-01-10
Access history dialog only in review state

terrymyc:
While opening a project waiting for setup or initialing a new project, the history dialog is unavailable. Currently, it tries to access the history dialog, which throws an internal error. The history dialog should be accessible only after the start of the review.

### DIFF
--- a/asreview/webapp/src/App.js
+++ b/asreview/webapp/src/App.js
@@ -185,12 +185,14 @@ const App = (props) => {
         toggleExportResult={toggleExportResult}
         exportResult={exportResult}
       />
+      {(props.app_state === 'review') &&
       <HistoryDialog
         recordState={recordState}
         setRecordState={setRecordState}
         toggleHistory={toggleHistory}
         history={history}
       />
+      }
     </ThemeProvider>
   );
 }
--- a/asreview/webapp/src/App.js
+++ b/asreview/webapp/src/App.js
@@ -67,6 +67,7 @@ const App = (props) => {
     'isloaded': false,
     'record': null,
     'selection': null,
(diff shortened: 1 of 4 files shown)


J535D165:
Thanks. This resolves the internal server error in the beginning, however I think we need to resolve a couple more things. I think the history dialog should not update at every decision. It should request the history when opening the dialog. Not update it in advance. 

A loader and error handler would be nice to have. See example datasets.

terrymyc:
> I think the history dialog should not update at every decision. It should request the history when opening the dialog. Not update it in advance.

Now making decisions in the review screen will not update the history dialog, whereas changing a decision in the history dialog will update both the dialog and the review screen.

J535D165:
You are of great help today. I will test, review, merge it asap.

J535D165:
I'm facing an issue. All requests are double.

This was caused by moving the `recordState` to `App.js`. This results in a lot of state refreshes as well. I do understand why this is implemented (updating the record in the background). This is when Redux can be very useful. However, this takes quite some time to implement/debug. Therefore, I think we should revert the update-in-the-background function for now and implement this later on in redux. 

0.15
Screenshot 2021-01-11 at 22 10 31

PR  #463
Screenshot 2021-01-11 at 22 11 30

terrymyc:
> I think we should revert the update-in-the-background function for now and implement this later on in redux.

Oops, thanks for the heads up! I reverted the change. Now the review screen doesn't update when a decision is changed in the history dialog.

J535D165:
Thanks @terrymyc, seems to work well now!

PR #465, opened 2021-01-10
Drop broken pickle files due to change in submodule data 

J535D165:
Pickle files can't be read anymore because of refactor of data submodule. This PR can catch all errors and creates a new temporary pickle file.


J535D165:
I will add this PR to the maintenance branches of 0.14 and 0.15. This ensures backward compatibility.

PR #471, opened 2021-01-17
Avoid file changes after running tests

J535D165:
Running the test suite results in unintentionally changed files (4 state files used for testing). This PR proposes a solution to avoid file changes.

### DIFF
--- a/tests/test_simulate.py
+++ b/tests/test_simulate.py
@@ -91,11 +91,8 @@ def test_state_continue_h5(tmpdir):
 
 def test_nb(tmpdir):
 
-    # copy state file to tmp dir for changes
-    tmp_h5_state_fp = Path(tmpdir, "tmp_state_nb.h5")
-
     check_model(model="nb",
-                state_file=tmp_h5_state_fp,
+                state_file=None,
                 use_granular=True,
                 n_instances=1,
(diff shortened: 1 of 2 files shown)


tests/test_simulate.py
```diff
@@ -107,33 +127,52 @@ def test_rf():
...
 
 
 @pytest.mark.xfail(not ADVANCED_DEPS["tensorflow"],
                    raises=ImportError,
                    reason="requires tensorflow")
-def test_lstm_base():
+def test_lstm_base(tmpdir):
+
+    # copy state file to tmp dir for changes
+    tmp_h5_state_fp = Path(tmpdir, "tmp_state.h5")
```
PeterLombaers, line 148: Is there a reason some have h5 state files and some json? Just for variety?

PeterLombaers (approved):
Indeed seems like an improvement to use a copy and a temporary directory!

tests/test_simulate.py
J535D165: Yes, it isn't consistent. I think this is for variety. Might be good to test the complete array (in another PR). However, we will replace both state files with a new state model. So, it might not be very useful to prioritize this.

PR #483, opened 2021-01-28
add description new plugin wordcloud

Rensvandeschoot:


J535D165:
There is an issue with the line endings I think https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#global-settings-for-line-endings. Merged via c74a0dd6, thanks.

PR #456, opened 2021-01-06
Move statistics from asreview-stats into asreview

J535D165:
This PR moves the first part of https://github.com/asreview/asreview-statistics into asreview. Starting with the most urgent part: the dataset statistics.

PR #519, opened 2021-03-19
Move statistics from asreview-stats into asreview

J535D165:
Similar to #456, but without CLI argument.


J535D165:
This PR replaces PR #456. I think it is better to add those command-line utils to an extension. I think this extension should contain tools like `data-convert`, `data-dedup`, and `data-describe`.  The functionality in this PR, is relevant to the core as well. The wrapper and entry-point in the extension make use of the functions in the PR.

PR #528, opened 2021-03-24
Remove FAQ (and use Discussions instead)

J535D165:
The following questions have to be migrated or marked as outdated. 

- [x] How to check which version of ASReview I have installed?
- [x] What parts of my Systematic Review project does ASReview support?
- [x] Can we use ASReview LAB for things other than Systematic Reviews?
- [x] Can we use ASReview LAB as a stand-alone screener?
- [x] Has the use of ASReview in systematic reviews been validated?
- [x] Has ASReview been used in any peer-reviewed article?
- [x] How can I refer to ASReview in my paper?
- [x] Can the tool be used in combination with any (academic) database?
- [x] How to deal with records that do not have abstracts?
- [x] How does the tool handle quality and standardisation of abstracts? Is this accounted for in any way in training the machine learning model?
- [x] What happens if I have records from different languages?
- [x] Why did you choose a license that allows commercial reuse for the software?
- [x] What do you mean with a dataset?
- [x] What do you mean with a model?
- [x] How do we decide when to stop?
- [x] Can we use ASReview LAB with multiple screeners?
- [x] Can we use ASReview LAB also to screen full text?
- [x] Which classifier should I choose in ASReview LAB?
- [x] I already started labeling before I came across ASReview. How can I keep my former screening decisions when starting a new project in ASReview?
- [x] Related question: How can I add more publications while I are already started screening in ASReview?
- [x] Is it possible to get the inclusion likelihood for unlabeled papers?
- [x] How can I make my previously labeled records green, like in the benchmark datasets?
- [x] How do I remove duplicate publications?


terrymyc (approved):

ottomattas (approved):
Changes look good.

PS. Some toctree issues for unchanged files, should we check them out?
- docs/source/index.rst#L61
- docs/source/plugins/plugin_wordcloud.rst#L1

J535D165:
Thanks @ottomattas, toctree issue resolved in #535.

Issue #362, opened 2020-10-06
Better contrast in dark mode


Sybren-UU:
## Feature Request

**Is your feature request related to a problem? Please describe.**
In dark mode the contrast is sometimes hard to read, especially when starting a project the text is brown with a black background, the doi is also brown with a black background

**Describe the solution you'd like**
It would be nice if the contrast would get better. So no brown text with a black/grey background in dark mode. The overal contrast is also low, some users in the usability tests thought this could be higher.

**Why would you like to have this feature**
Especially in dark mode brown text with a grey/black is hard to read. The overall contrast is fine, but also low.

J535D165:
Closed via #531

PR #531, opened 2021-03-28
Improve contrast ratio

terrymyc:
In response to #362, this PR improves the contrast ratio according to [WCAG guidelines](https://webaim.org/blog/wcag-2-0-and-link-colors/) on several pages either in light or dark mode.

- [x] Links
- [x] Document texts in the Exploration Mode
- [x] Other places have brown texts with black background

image

image

image

### DIFF
--- a/asreview/webapp/src/Components/ArticlePanel.js
+++ b/asreview/webapp/src/Components/ArticlePanel.js
@@ -20,10 +20,10 @@ const useStyles = makeStyles(theme => ({
   },
   titleDebug: {
     lineHeight: 1.2,
-    color: "#00C49F",
+    color: theme.overrides.debug.color,
   },
   debug: {
-    color: "#00C49F",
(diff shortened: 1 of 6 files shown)


J535D165:
https://github.com/asreview/asreview/blob/71bd23606e5f36c8be96dbc5945d2734526145be/asreview/webapp/src/Components/ArticlePanel.js#L94

Let's set this to primary (delete it). Both for title and abstract. Don't know why we used secondary in the past.

terrymyc:
> Let's set this to primary (delete it). Both for title and abstract. Don't know why we used secondary in the past.

Then the texts in the article panel will be purely black. You sure? The current contrast ratio of the grey texts against the white background is acceptable.

image

J535D165:
I think it is 0.87 black isn't it? I really like it. It looks way better on my new laptop.

J535D165:
Thanks!

PR #533, opened 2021-03-29
Add project mode to REST API and project files

J535D165:


asreview/webapp/src/PreReviewComponents/ProjectInit.js
```diff
@@ -0,0 +1,237 @@
+import React from "react";
```
J535D165, line 1: @terrymyc was this file moved on master?

terrymyc: This file was renamed as `ProjectInfo.js`.

asreview/webapp/src/PreReviewComponents/ProjectInfo.js
```diff
@@ -96,26 +110,17 @@ const ProjectInfo = (props) => {
...
     var bodyFormData = new FormData();
+    bodyFormData.set("mode", info.mode);
     bodyFormData.set("name", info.name);
     bodyFormData.set("authors", info.authors);
     bodyFormData.set("description", info.description);
 
-    (props.edit
-      ? ProjectAPI.info(props.project_id, true, bodyFormData)
-      : ProjectAPI.init(bodyFormData)
-    )
```
terrymyc, line 106: All lines involving the removal of `props.edit` will affect the functionality of editing the project info. Currently, this initiating a new project function and editing the project info function share the same dialog. For now, I would propose to keep `props.edit` as an indicator of whether this dialog is used as a function to initiate or edit a project. And when the dialog plays the function of editing the project info, no selection menu for the mode is displayed.
In the future, I expect these two functions to be separated. The function to initiate a project should be integrated into the setup stage.

terrymyc (changes requested):

asreview/webapp/api.py
```diff
@@ -714,49 +703,112 @@ def api_set_algorithms(project_id):  # noqa: F401
...
+            logging.error(err)
+            return jsonify(message="Failed to train the model."), 500
 
     response = jsonify({'success': True})
     response.headers.add('Access-Control-Allow-Origin', '*')
     return response
 
 
-@bp.route('/project/<project_id>/model/init_ready', methods=["GET"])
+@bp.route('/project/<project_id>/ready', methods=["GET"])
```
terrymyc, line 762: The corresponding test for this API in `test_api.py` should also be changed.

terrymyc (changes requested):

asreview/webapp/src/PreReviewComponents/ProjectInfo.js
J535D165: Thanks, great remark. I will try to resolve these conflicts and make the edit function work again.

J535D165: I think I resolved this.

asreview/webapp/api.py
J535D165: Thanks

J535D165:
I'm merging this even though the tests fail. These tests are designed to test old project files for forward compatibility. They fail for various reasons (not sure what the cause is). However, I want to avoid more merge conflicts and merge this PR. 

Another remark: the simulation mode doesn't work properly yet (seems to be random). Working on this in a new PR.

PR #566, opened 2021-04-10
Automated Test for Project Created in Old Versions

terrymyc:
This PR implements automated testing for REST API in processing project files created in previous versions. The project files will be automatically retrieved and tested for every commit.

### DIFF
--- a/asreview/webapp/utils/project.py
+++ b/asreview/webapp/utils/project.py
@@ -479,7 +479,7 @@ def retrieve_project_url_github():
             version_tags.append(file["path"])
 
     for tag in version_tags:
-        file_url = f"/{tag}/asreview-project-{tag.replace('.', '-')}-"
-        project_urls.append(repo_url + file_url + file_type)
+        file_version = f"/{tag}/asreview-project-{tag.replace('.', '-')}-"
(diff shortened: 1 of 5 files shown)


asreview/webapp/utils/project.py
```diff
@@ -457,3 +458,28 @@ def move_label_from_labeled_to_pool(project_id, paper_i):
...
+
+    for file in json_file:
+        if file["type"] == "tree":
+            version_tags.append(file["path"])
+
+    for tag in version_tags:
+        file_version = f"/{tag}/asreview-project-{tag.replace('.', '-')}-"
+        project_urls.append(repo_url + file_version + file_type)
+
+    return project_urls
```
J535D165, line 485: I think it makes more sense to keep this in the tests folder. Maybe create a utils file in there.

asreview/webapp/utils/project.py
```diff
@@ -19,6 +19,7 @@
 import sys
 import time
 from pathlib import Path
+from urllib.request import urlopen
```
J535D165, line 22: Remove?

J535D165:
So happy with this one.

PR #577, opened 2021-04-15
Fix boot api error handler

terrymyc:
This PR fixes the `boot` API error handler when the server is down.


J535D165:
Good catch!

PR #581, opened 2021-04-16
Fix line chart read data before loaded

terrymyc:
This PR fixes the possible failure of the progress line chart on the project page. The custom tooltip might try to read data before it is loaded.


J535D165:
Thanks!

PR #591, opened 2021-04-27
Rewrote some text and fixed typos

govertv:


J535D165 (approved):
Thanks, we had to write the docs in very limited time. Thanks for improving it!

PR #600, opened 2021-05-12
Added quick link to Docker instructions

behrica:
I just learned from Rens, that the Docker version of asreview is very successful, but I did not find a "quick link" to the Docker instructions in the main README.


J535D165 (approved):

J535D165:
Thanks @behrica

PR #606, opened 2021-05-17
Implementation of basic HTTP auth and project locking functionality

valmelnikov:


asreview/webapp/auth.py
```diff
@@ -0,0 +1,151 @@
...
+        super().__init__(scheme=scheme, realm=realm)
+
+        self.reinit(auth_file)
+
+
+    def reinit(self, auth_file):
+        self.auth_file = auth_file
+
+        if not auth_file:
+            self.auth_file = get_auth_file_path()
```
leonardovida, line 26: Here it would be nice to inform the user that we are retrieving the default "auth.txt" file

asreview/webapp/auth.py
```diff
@@ -0,0 +1,151 @@
...
+        if delimiter is None:
+            delimiter = ASReviewAuth.USERPWD_DELIMITER
+
+        lines = ['%s%s%s\n' % (user, delimiter, pwd) for user, pwd in users.items()]
+        with open(fp, 'w') as f:
+            f.writelines(lines)
+
+
+    @staticmethod
+    def cmd_tool(argv):
```
leonardovida, line 112: I guess it's personal, but I think the `cmd_tool()` logic might be put in `start_flask.py` so that what is passed to `auth.py` is a working auth_file.
Doing this would also allow us to move the `entry_point/auth` to the `lab` and have the main parser argument in one place without the need of an external entry point called by cmd. But this needs to be discussed with @J535D165 !

asreview/webapp/sqlock.py
```diff
@@ -123,3 +123,114 @@ def release(self):
         logging.debug(
             _log_msg(f"Released lock {self.lock_name}", self.project_id))
         self.lock_acquired = False
+
+
+class SQLiteUserLock():
```
leonardovida, line 128: This is great!

leonardovida (commented):
I really like this! What I am very slowly developing is the user management with the frontend. You are doing user creation (to which I can adapt what I began doing) and project locking (which I didn't touch). 

An obvious addition, which I guess you are already working on is comprehensive tests for the project locking (and potentially user creation).

I would like to discuss in our next call is what you see next for this PR and then understand how to better merge my previous work with this great PR.

asreview/webapp/auth.py
valmelnikov: Done

asreview/webapp/auth.py
valmelnikov: I see your point, but would really discuss if that is a better way. Even if we make auth file manipulation a sub-command of lab cmd command, I would keep this method within dedicated auth module.

J535D165:
Before we merge into `asreview:auth-layer`, it might be good to fix the test-frontend. Pep8 linting would also be nice.

J535D165:
Thanks a lot for the implementation. Looking forward to the PR of @leonardovida extending this.

PR #610, opened 2021-05-17
REST API to Fetch List of Algorithms

terrymyc:
This PR develops a REST API to fetch a list of available classifiers, balance strategies, query strategies, and feature extraction methods in the backend and implements them in the frontend.
This PR also develops a REST API to clear the error state file while retrying setup.
This PR also adds labels to all algorithms.

image

image

image


J535D165:
Nice! Might be interesting to consider updating this at the same time https://github.com/asreview/asreview/blob/a296205f7fe1f2e4a09e1856bd5be01438d9223a/asreview/entry_points/algorithms.py#L37-L72. Labels can be interesting to have there as well (`asreview algorithms`)

terrymyc:
> . Labels can be interesting to have there as well (`asreview algorithms`)

@J535D165 So like replace "logistic" with "Logistic regressions", right?

J535D165:
I think both (such that you know the name for the simulation CLI).

terrymyc:
@J535D165 There is an internal bug when model training fails. For example, if I choose Doc2Vec without `gensim` package installed, the model training will throw `ImportError` with `result.json` and `error.json` created. The problem is that if I refresh and choose another viable option other than `Doc2Vec`, the `ImportError` of `gensim` still appears. The moment I change the option, `kwargs.json` will change, but `result.json` and `error.json` are not cleared.

J535D165:
Very interesting feature. Do we need to update parts of the documentation as well?

terrymyc:
> Very interesting feature. Do we need to update parts of the documentation as well?

@J535D165 I think `list_classifiers` and other similar functions need new descriptions. I guess in the documentation those are updated automatically?

asreview/models/classifiers/lstm_pool.py
```diff
@@ -90,6 +90,7 @@ class LSTMPoolClassifier(BaseTrainClassifier):
     """
 
     name = "lstm-pool"
+    label = "LSTM-pool"
```
J535D165, line 93: LSTM with max-pooling layer?

terrymyc: Meaning...?

terrymyc: Then what about the label for `lstm-base`?

asreview/models/query/max_random.py
```diff
@@ -0,0 +1,33 @@
...
+    At each query 95% of the instances would be sampled with the maximum
+    query strategy after which the remaining 5% would be sampled with
+    the random query strategy.
+    """
+
+    name = "max_random"
+    label = "Mixed (95% Maximum and 5% Random)"
+
+    def _query(self, X):
+        return MixedQuery
```
terrymyc, line 33: @J535D165 This subclass seems to work. Did I miss something?

J535D165:
For readers of the https://www.nature.com/articles/s42256-020-00287-7 publication. This PR breaks the naming of "Mixed sampling". 

> The available query strategies for the active learning part are (1) random selection, ignoring model-assigned probabilities; (2) uncertainty-based sampling, which chooses the most uncertain record according to the model (that is, closest to 0.5 probability); (3) certainty-based sampling (max in ASReview), which chooses the record most likely to be included according to the model; and (4) mixed sampling, which uses a combination of random and certainty-based sampling.

Mixed sampling is a contraction of two sampling methods. This is made explicit from now on and replaced by `max_uncertainty` or `max_random`. We decided to not make this partly backward compatible. Users of ASReview LAB/BUI can continue with projects making use of this method. For sake of maintainability, the simulation CLI doesn't have this option anymore. If there are questions about the deprecation, please leave a message in this thread.

asreview/models/balance/double.py
```diff
@@ -49,6 +49,7 @@ class DoubleBalance(BaseBalance):
     """
 
     name = "double"
+    label = "Double"
```
Rensvandeschoot, line 52: Can you add a bit more explanation like you do with the other labels? for example `Two way rebalancing function`

asreview/models/balance/triple.py
```diff
@@ -62,6 +62,7 @@ class TripleBalance(BaseBalance):
     """
 
     name = "triple"
+    label = "Triple"
```
Rensvandeschoot, line 65: Can a bit more information be inlcluded? `This divides the training data into three sets: included papers, excluded    papers found with random sampling and papers found with max sampling`

asreview/models/balance/undersample.py
```diff
@@ -34,6 +34,7 @@ class UndersampleBalance(BaseBalance):
     """
 
     name = "undersample"
+    label = "Undersampling"
```
Rensvandeschoot, line 37: For the label description: `undersampling the irrelevant records`

asreview/models/classifiers/lstm_base.py
```diff
@@ -86,6 +86,7 @@ class LSTMBaseClassifier(BaseTrainClassifier):
     """
 
     name = "lstm-base"
+    label = "LSTM classic"
```
Rensvandeschoot, line 89: what is the difference between `base` and `classic`? If there is none, I would not suggest to use both terms. Maybe a more informative label can be provided? `LSTM layer with one output, a dense layer, and a single sigmoid output node`

asreview/models/feature_extraction/doc2vec.py
```diff
@@ -90,6 +90,7 @@ class Doc2Vec(BaseFeatureExtraction):
     """
 
     name = "doc2vec"
+    label = "Doc2Vec"
```
Rensvandeschoot, line 93: maybe change to something like `Feature extraction technique provided by the `gensim    &lt;https://radimrehurek.com/gensim/&gt;`__ package`

asreview/models/feature_extraction/embedding_idf.py
```diff
@@ -63,6 +63,7 @@ class EmbeddingIdf(BaseFeatureExtraction):
     """
 
     name = "embedding-idf"
+    label = "Embedding IDF"
```
Rensvandeschoot, line 66: maybe add a bit more info?

asreview/models/feature_extraction/embedding_lstm.py
```diff
@@ -75,6 +75,7 @@ class EmbeddingLSTM(BaseFeatureExtraction):
     """
 
     name = "embedding-lstm"
+    label = "Embedding LSTM"
```
Rensvandeschoot, line 78: same.. needs to be a bit more informative

asreview/models/query/max.py
```diff
@@ -19,12 +19,13 @@
...
 
 class MaxQuery(ProbaQueryStrategy):
-    """Maximum sampling query strategy.
+    """Maximum query strategy.
 
     Choose the most likely samples to be included according to the model.
     """
 
     name = "max"
+    label = "Maximum"
```
Rensvandeschoot, line 28: `Max refers to the most likely records to be shown to classify`

asreview/models/query/max.py
```diff
@@ -19,12 +19,13 @@
 
 
 class MaxQuery(ProbaQueryStrategy):
-    """Maximum sampling query strategy.
+    """Maximum query strategy.
 
     Choose the most likely samples to be included according to the model.
```
Rensvandeschoot, line 24: samples=>records?
included=>shown?

asreview/models/query/random.py
```diff
@@ -20,7 +20,7 @@
 
 
 class RandomQuery(NotProbaQueryStrategy):
-    """Random sampling query strategy.
+    """Random query strategy.
 
     Randomly select samples with no regard to model assigned probabilities.
```
Rensvandeschoot, line 25: samples=>records?

asreview/models/query/random.py
```diff
@@ -30,6 +30,7 @@ class RandomQuery(NotProbaQueryStrategy):
 
     """
     name = "random"
+    label = "Random"
```
Rensvandeschoot, line 33: `Random selection of records`?

asreview/models/query/uncertainty.py
```diff
@@ -19,7 +19,7 @@
 
 
 class UncertaintyQuery(ProbaQueryStrategy):
-    """Maximum uncertainty query strategy.
+    """Uncertainty query strategy.
 
     Choose the most uncertain samples according to the model (i.e. closest to
```
Rensvandeschoot, line 24: samples=>records

asreview/models/query/uncertainty.py
```diff
@@ -28,6 +28,7 @@ class UncertaintyQuery(ProbaQueryStrategy):
     """
 
     name = "uncertainty"
+    label = "Uncertainty"
```
Rensvandeschoot, line 31: `Shows the most uncertain records to be screened`

asreview/models/query/uncertainty.py
J535D165: Out of the scope of this PR.

asreview/models/query/random.py
J535D165: Out of the scope of this PR.

asreview/models/query/max.py
J535D165: Outside the scope of this PR.

J535D165:
@Rensvandeschoot this many changes compared to https://www.nature.com/articles/s42256-020-00287-7/tables/2 will be very confusing and requires major issues with our documentation, isn't it?

I would suggest replacing the labels of double and triple with Dynamic resampling (Double) and Dynamic resampling (Triple) (see https://www.nature.com/articles/s42256-020-00287-7/tables/2). I would suggest not making major changes to the other labels.

asreview/models/balance/double.py
J535D165, line 52:
```suggestion
    label = "Dynamic resampling (Double)"
```

asreview/models/balance/triple.py
J535D165, line 65:
```suggestion
    label = "Dynamic resampling (Triple)"
```

J535D165:
YESSSS. Thank you so much @terrymyc!

PR #625, opened 2021-06-09
Readthedocs- changed step 4

JTeijema:
feature extraction and selecting the classifier is only done once, it used to say back to step 1. Also maybe there should be a step 5 which ends the process.


J535D165 (approved):

J535D165:
Thanks! Happy to have a step 5 as well.

PR #631, opened 2021-06-11
Add doi to Configuration File

terrymyc:
This PR adds `doi` to the configuration file of the software. The column name `doi` of a dataset file is then not case-sensitive.


J535D165 (approved):

J535D165:
Very nice implementation!

PR #635, opened 2021-06-16
Deprecated lr. use learning_rate

JTeijema:


J535D165:
Thanks for the PR. Can you add the minimal version (of tf?) to `setup.py`? See:

https://github.com/asreview/asreview/blob/c03eb4ba380cc817819abab78d8ee96bee340f4d/setup.py#L111-L124

JTeijema:
I.. am not sure how I should be figuring out the minimal version hahaha

J535D165:
It seems like this holds for the entire 2-series of TensorFlow. Maybe restrict ASReview to TensorFlow 2.

J535D165:
Thanks! Fixed remark via PR #672.

PR #636, opened 2021-06-16
Transfer Learning setup for neural networks

govertv:
This is a demo of how transfer learning can be implemented for the neural networks in ASReview. Right now its a bit hacky/lazy programming, which I plan to fix later... 

Explanation of how it works right now:
After the first model is trained, the full model gets saved to 'model.h5'. When training the subsequent models, the previous model is loaded first, and then its trained for 1 epoch (technically I implemented a function that lowers the amount of epochs from 35 to 1, but it goes down to 1 very quickly) on the new data. This saves a lot of computation time on training the models. In the tests I ran with the nudging dataset, it lowered the duration of the full simulation by about 50%.

There seems to be some loss in performance right now (sample size n=1), but in theory this could be regained by using more complex models that would otherwise be unfeasible due to the computation time required to train them.

Current neural network implementation simulated on the Nudging dataset:
regular nn nudging

Transfer learning neural network implementation simulated on the Nudging dataset:
transfer learning nn nudging

I think this test case is proof that loading the weights works correctly, here is what happens when I let it use the model from the previous simulation run from the start:
only transfer learning nudging

### DIFF
--- a/asreview/models/classifiers/nn_2_layer.py
+++ b/asreview/models/classifiers/nn_2_layer.py
@@ -14,13 +14,13 @@
 
 import logging
 
-
 try:
     import tensorflow as tf
     from tensorflow.keras.layers import Dense
     from tensorflow.keras.models import Sequential
     from tensorflow.keras.wrappers.scikit_learn import KerasClassifier
     from tensorflow.keras import regularizers
+    from tensorflow.keras.models import load_model
 except ImportError:
     TF_AVAILABLE = False
 else:
@@ -31,6 +31,8 @@ else:
         logging.getLogger("tensorflow").setLevel(logging.ERROR)
 
 import scipy
+from os.path import isfile
+import numpy as np
 
 
 from asreview.models.classifiers.base import BaseTrainClassifier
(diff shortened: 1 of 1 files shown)


J535D165:
Very interesting to see the differences. Sounds like a very promising feature. I think it is important to give the user of the model the option to use this feature or not (add it to the constructor of the class). Looking forward to the next steps.

asreview/models/classifiers/nn_2_layer.py
```diff
@@ -183,6 +191,11 @@ def _create_dense_nn_model(vector_size=40,
     _check_tensorflow()
 
     def model_wrapper():
+
+        if isfile('model.h5'):
+            model = load_model("model.h5")
+            return model
```
J535D165, line 197: This is risky because multiple simulations can be performed from the same folder. That might result in conflicts.

J535D165:
Very interesting idea. Can be really useful. 

I think we should develop this as an extension first https://asreview.readthedocs.io/en/latest/API/extensions_dev.html#add-model. 

I would suggest adding this to ASReview when:

- performance is at least as good as the current implementation
- it has a new name such that the old model remains backward compatible
- it is technically stable (temp store of the model is an open challenge)
- fully reproducible with the state file

It might be easier to implement this as an extension and work on the items mentioned above. This makes development a lot faster (less testing and quality checks). 

After the model was developed, we are welcoming a PR to add it to the list of model extensions https://asreview.readthedocs.io/en/latest/extensions/overview_extensions.html. After this was considered to be mature, we can integrate it in ASReview (or not).

PR #640, opened 2021-06-18
Not really clear how to set the axis to absolute in the visualization part.

JTeijema:
API section needs a better explanation, its also not in quick.py.

### DIFF
--- a/docs/source/plugins/plugin_visualization.rst
+++ b/docs/source/plugins/plugin_visualization.rst
@@ -158,11 +158,13 @@ Those are the scripts that are used for the command line interface.
     from asreviewcontrib.visualization.plot import Plot
 
     with Plot.from_paths(["PATH_1", "PATH_2"]) as plot:
-        inc_plot = plot.new("inclusion")
+        inc_plot = plot.new("inclusion", result_format="number")
         inc_plot.set_grid()
         inc_plot.set_xlim(0, 30)
         inc_plot.set_ylim(0, 101)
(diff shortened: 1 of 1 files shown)


docs/source/plugins/plugin_visualization.rst
```diff
@@ -158,11 +158,13 @@ Those are the scripts that are used for the command line interface.
     from asreviewcontrib.visualization.plot import Plot
 
     with Plot.from_paths(["PATH_1", "PATH_2"]) as plot:
-        inc_plot = plot.new("inclusion")
+        inc_plot = plot.new("inclusion", result_format="number")
```
J535D165, line 161: Maybe better to make a second example for this one?

docs/source/plugins/plugin_visualization.rst
```diff
@@ -158,11 +158,13 @@ Those are the scripts that are used for the command line interface.
...
         inc_plot.set_grid()
         inc_plot.set_xlim(0, 30)
         inc_plot.set_ylim(0, 101)
         inc_plot.set_legend()
+        inc_plot.add_wss('filename.h5', value=100)
```
J535D165, line 166: The file name is slightly confusing. Do you know what it is? Is it a position argument or keyword?

J535D165 (commented):
Thanks for the PR. It might be better to add a second API example with these changes. This makes it clear that the default is not that complicated. The second example shows that there is a lot of flexibility.

PR #654, opened 2021-07-06
Fix Missing doi Bug

terrymyc:
This PR fixes a bug introduced by #631. Missing doi in a dataset caused failure in starting reviewing.

### DIFF
--- a/asreview/io/paper_record.py
+++ b/asreview/io/paper_record.py
@@ -139,7 +139,11 @@ class PaperRecord():
                 col = attr
             else:
                 col = None
-            setattr(self, attr, kwargs.pop(col, None))
+
+            attr_val = kwargs.pop(col, None)
+            if attr_val is not None and pd.isna(attr_val):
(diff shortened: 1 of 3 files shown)


J535D165 (changes requested):
Thanks. It would be really useful to have a unit test for this such that we avoid this issue in the future.

J535D165:
Can you also implement `publish_time` is the same way as `doi` is implemented?

terrymyc:
> Can you also implement `publish_time` in the same way as `doi` is implemented?

Should we add `publish_time` to the [documentation](https://asreview.readthedocs.io/en/latest/intro/datasets.html#data-format)? This column name is ambiguous. I think we usually see `year` and `date` in a reference manager. And is it necessary to have `date` in a dataset?

J535D165 (approved):

J535D165:
Thanks for the fix @terrymyc.

PR #664, opened 2021-07-14
Add one-click-deploy buttons

leonardovida:
Add one-click-deploy buttons:
- Azure
- Heroku
- Google Cloud

Given that Heroku and GC could not be tested on a private repository, this PR will be used to test them (thus more commits will follow).


J535D165:
Thanks for this contribution. Looking forward to the user feedback.

PR #666, opened 2021-07-15
Add Styled OpenInNewIcon Component

terrymyc:
This PR adds a styled OpenInNewIcon component. It is used in different cases when a clickable button directs users to a new window.


J535D165 (approved):

J535D165:
Thanks @terrymyc!

PR #671, opened 2021-07-20
Implementation transfer learning + early stopping for neural network

BartJanBoverhof:
Build upon [earlier work ](https://github.com/asreview/asreview/pull/636) by @govertv

### DIFF
--- a/asreview/models/classifiers/nn_2_layer.py
+++ b/asreview/models/classifiers/nn_2_layer.py
@@ -14,13 +14,15 @@
 
 import logging
 
-
 try:
     import tensorflow as tf
     from tensorflow.keras.layers import Dense
     from tensorflow.keras.models import Sequential
     from tensorflow.keras.wrappers.scikit_learn import KerasClassifier
     from tensorflow.keras import regularizers
+    from tensorflow.keras.models import load_model
+    from tensorflow.keras.callbacks import EarlyStopping 
+
 except ImportError:
     TF_AVAILABLE = False
 else:
@@ -31,6 +33,8 @@ else:
         logging.getLogger("tensorflow").setLevel(logging.ERROR)
 
 import scipy
+from os.path import isfile
+import numpy as np
 
 
 from asreview.models.classifiers.base import BaseTrainClassifier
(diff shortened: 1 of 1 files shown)


BartJanBoverhof:
This PR is an implementation of transfer learning + early stopping for the 2 layer neural network, as implemented already.

# I. Why? 
Training the `nn-2-layer` neural network takes quite a bit of time. By implementing transfer learning + early stopping, training times can be shortened considerably. This code builds upon an earlier made [PR](https://github.com/asreview/asreview/pull/636) by @govertv. In this PR transfer learning was implemented by saving and re-using the model at each iteration. Each iteration consisted of a single epoch. 

# II. What does this PR add?
In addition to transfer learning as implemented by @govertv, early-stopping is added. The reason for doing so is to try and minimise the loss in performance, whilst still trying to reduce training time. Instead of solely running the a single epoch within each iteration, a maximum of the specified number of epochs (with current default: `epochs=35`) may be run. The early stopping rule stops training breaks off the training when the training loss has not been improved by the value `delta`, and moves to the next iteration. It does so with a "patience" of `patience`. 

# III. Simulation results
A small simulation is done, comparing both results and training times of the regular `nn-2-layer` with the modified `nn-2-layer`. The utilised parameters (specified as defaults) for early stopping are set at `delta=0.03` and `patience=2` respectively.  Simulation have been ran in simulation mode on the `Kwok_2020` dataset. 

### **Statistics + training time comparison**
|               | Vanilla nn | Transfer-learning + Early-stopping nn |
|---------------|---------|------------------------------------|
| Training time |         | 1h 13m                             |
| WSS@95        |         | 47.04 %                            |
| WSS@100       |         | 20.89 %                            |
| RRF@5         |         | 57.26 %                            |
| RRF@10        |         | 72.65 %                            |


### **Recall curve: Transfer-learning + Early-stopping model**
transfer

### **Recall curve: vanilla nn**

J535D165:
Great contribution. I like the Early Stopping and model saving. 

There are several technical concerns regarding the implementation in my opinion. I'm describing two of them:

- backward compatibility. Changing a model affects existing projects. Not a big deal at the moment, because this model is not available in the frontend (<=0.17).  Simulations are less problematic because they are conducted in one run usually. Should we rename this model (such that the legacy model can live next to the new one), create a plugin, of replace it?
- saving the model in this way causes potential issues when running multiple simulations in a row/multicore. Where do you delete the model when the simulation is finished?

J535D165:
I'm closing this PR. See PR #636 for the reasoning. Feel free to discuss this in the thread. 

I do have plenty of questions about the performance btw. It looks very promising. But maybe discuss this in the model extension repo.

PR #684, opened 2021-08-03
Fix the Toobal to the Top in Settings Dialog

terrymyc:
This PR fixes the toolbar to the top in the settings dialog by wrapping dialog content with the `DialogContent` component.


J535D165 (approved):

J535D165:
Thanks for the fix

PR #694, opened 2021-08-11
Implement new data model for state file

PeterLombaers:
This pull request implements a working version of a new data format. The old state files in .json or .h5 format will be deprecated, and in its place will come a new .asreview file. There is a converter utility that converts the old .json format to the new .asreview format. 

## Frontend

For ongoing projects, users will be prompted to automatically convert their project to the new format. After that everything is the same. 

## Simulate

This PR changes the state files into ASReview project files. Basically, the state files are now added to the project file container. 

Old:
```
$ asreview simulate benchmark:Hall_2012 -s test.h5
```

New:
```
$ asreview simulate benchmark:Hall_2012 -s test.asreview
```


asreview/state/__init__.py
```diff
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 from asreview.state.base import BaseState
-from asreview.state.dict import DictState
+from asreview.state.legacy.dict import DictState
+from asreview.state.legacy.hdf5 import HDF5StateLegacy
 from asreview.state.hdf5 import HDF5State
```
J535D165, line 18: What would be a good name for the new model?

asreview/state/__init__.py
```diff
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 from asreview.state.base import BaseState
-from asreview.state.dict import DictState
+from asreview.state.legacy.dict import DictState
+from asreview.state.legacy.hdf5 import HDF5StateLegacy
```
J535D165, line 17: I think we can remove all legacy state classes. This makes reviewing more straightforward.

J535D165 (changes requested):
Thanks for this PR. Very promising.

I posted some first feedback. Can you also remove the lint errors (use yapf for formatting, `yapf -i ...`)?

asreview/state/__init__.py
PeterLombaers: Maybe SqlState? It's not entirely accurate, because there are also other files in there, but it fits in line with jsonstate and hdf5state.

PeterLombaers: Let's do SqlStateV1.

asreview/state/sqlstate.py
```diff
@@ -550,6 +558,40 @@ def add_labeling_data(self, record_ids, labels, classifiers,
...
+        cur.execute(
+            "UPDATE results SET labels = ? WHERE record_ids = ?",
+            (new_label, record_id)
+        )
+
+        # Add the change to the decision changes table.
+        cur.execute(
+            "INSERT INTO decision_changes VALUES (?,?, ?)",
+            (record_id, new_label, current_time)
+        )
```
J535D165, line 582: I like this.

J535D165 (approved):

PR #695, opened 2021-08-11
Add Mobile Screen State to Top Level

terrymyc:
This PR adds a state that indicates the screen size to the top level. This state has already been used in the settings, help, and review dialog and will be used in the history dialog.


J535D165:
Thanks, do you also need this for the 0.x series? (history?) Or do we move this forward to the v1 release?

terrymyc:
I think it is easier to move forward to the v1 since this involves other v1 changes.

PR #605, opened 2021-05-12
Add min stopping simulation option

govertv:
The suggested changes allow using 'min' as an option for --n_queries. When using 'min' the simulation run will stop when all relevant papers have been includes. My testing indicates it works correctly. However, the error message for an incorrect --n_queries value (not an int and not 'min') is now a default python error which might be not be readable for users.

Another change will also be needed to enable ASReview-visualization to draw the full recall curve (see: https://github.com/asreview/asreview/issues/598)

Issue #598, opened 2021-05-07
Stop simulation mode when all relevant documents are found


govertv:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Simulation runs can take quite long for models with high computation times. The current default behavior is to stop the simulation when every single document has been labelled. However, after all relevant documents have been found, there is no reason to continue the simulation for most of the metrics currently used. A lot of computation time can be saved by stopping the simulation when all relevant documents are found.

![first_long_simulation](https://user-images.githubusercontent.com/42623130/117465718-42555600-af52-11eb-8602-1f9b490e9899.png)

**Describe the solution you'd like**
Add a CLI option that allows stopping the simulation when all relevant documents are found. This could probably be a default setting, though I'm sure there are some cases where it is useful to finish the full simulation run.

**Describe alternatives you've considered**
An alternative solution could be printing a message in the terminal when all relevant documents have been found. The user can then choose to stop the run by issuing a keyboard interrupt.

**Teachability, Documentation, Adoption, Migration Strategy**
Adding it to the current documentation about the CLI should be sufficient.

govertv:
Integrating it with n_queries sounds fine to me. Implementing it myself sounds like a nice thing to try. One more thing to consider is that the output of the recall plot should probably look the same as when the run is completed fully.  I think this could be done in two different ways: either in the package that handles the visualisations, or when creating the statefile. I don't know which of these is the preferred solution.

PeterLombaers:
Sounds like a smart idea to me! You'd save on some computing time, but you would lose the order in which the records are presented and the probabilities/ranking of the model for the last records. Since that ranking should be pretty stable after all relevant records have been labelled, you don't lose out on a lot I would think. At the very least it's a nice option to have.

The implementation should be the main asreview part, because that's where the simulations run, not in visualizations. I'm not completely sure where it should be. In the main reviewing loop for the simulation mode there should be some check if all relevant records have been labelled and then the reviewing loop stops?

J535D165:
I think we should do two things:

1. Update ASReview and extend `n_queries` with `min` or `auto`. 
2. Update ASReview-visualization such that it can complete the recall curve. 

The latter one requires some interesting challenges I guess. For example, you need to check whether the simulation was not aborted because of anything else. Might be interesting to have this check in the new state file API @PeterLombaers (something like `StateFile.is_simulation_complete`). 

The relevant code for the first todo (afaik): 

[...]

PR #697, opened 2021-08-15
Add min stopping plots

govertv:
Continuation of [#605](https://github.com/asreview/asreview/pull/605) (see also [#598](https://github.com/asreview/asreview/issues/598)). Currently when using the min stopping option to run simulations, the visualization plugin will create a recall plot that looks like this:

image

The code in this pull request extends the recall plot (when all relevant records have been found), so it looks this instead:

By making the changes in analysis.py as suggested here, no changes are needed to the state file or to the visualization plugin. One downside of the current implementation is that it would become default behavior and cannot be turned off by the user through the CLI. I'm not sure how to best address that.

### DIFF
--- a/asreview/analysis/analysis.py
+++ b/asreview/analysis/analysis.py
@@ -165,8 +165,18 @@ class Analysis():
             x_norm /= len(labels)
             y_norm /= self.inc_found[fl]["inc_after_init"]
 
-        norm_xr = (np.arange(1,
-                             len(self.inc_found[fl]["avg"]) + 1) - dx) / x_norm
+        # extend the arrays if all relevant records have been found
+        total_inc_found = int(self.inc_found[fl]["avg"][-1])
+        if total_inc_found == self.inc_found[fl]["inc_after_init"]:
+            norm_xr = np.arange(1, 1 + len(labels) -
+                                self.inc_found[fl]["n_initial"]) / x_norm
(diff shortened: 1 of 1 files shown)


asreview/analysis/analysis.py
```diff
@@ -165,8 +165,18 @@ def inclusions_found(self,
...
+        total_inc_found = int(self.inc_found[fl]["avg"][-1])
+        if total_inc_found == self.inc_found[fl]["inc_after_init"]:
+            norm_xr = np.arange(1, 1 + len(labels) -
+                                self.inc_found[fl]["n_initial"]) / x_norm
+            missing_y = len(labels) - self.inc_found[fl]["n_initial"] - \
+                len(self.inc_found[fl]["avg"])
+            self.inc_found[fl]["avg"].extend([total_inc_found] * missing_y)
+        else:
+            norm_xr = (np.arange(1, len(self.inc_found[fl]["avg"]) + 1) - dx)\
+                      / x_norm
```
J535D165, line 178:
```suggestion
            norm_xr = (np.arange(1, len(self.inc_found[fl]["avg"]) + 1) - dx) \
                / x_norm
```

J535D165: Hopefully, this will resolve the lint error.

J535D165 (commented):
Nice implementation. If we remove the lint error, we can merge :)

PR #698, opened 2021-08-16
Introduce React Query for Server State Management

terrymyc:
This PR introduces the [React Query](https://react-query.tanstack.com/) library to fetch, cache, synchronize and update server state.
This will likely
> Help you remove many lines of complicated and misunderstood code from your application and replace with just a handful of lines of React Query logic.
Make your application more maintainable and easier to build new features without worrying about wiring up new server state data sources.
Have a direct impact on your end-users by making your application feel faster and more responsive than ever before.
Potentially help you save on bandwidth and increase memory performance.

 The code refactoring will follow in other PRs.

### DIFF
--- a/asreview/webapp/package-lock.json
+++ b/asreview/webapp/package-lock.json
@@ -21,6 +21,7 @@
         "react-confetti": "^6.0.1",
         "react-dom": "^16.14.0",
         "react-dropzone": "^10.2.2",
+        "react-query": "^3.19.6",
         "react-rotation": "^1.3.0",
         "react-scripts": "^3.4.4",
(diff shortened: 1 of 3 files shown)


J535D165:
Thanks, @terrymyc, looks very interesting. This will definitely simplify lots of components. Shall we implement this in V1 only? Merging this also in *master* will be a lot of effort.

terrymyc:
Sure. Let's take this change to v1.

J535D165 (approved):

J535D165:
An interesting module that can be of great benefit for ASReview.

PR #709, opened 2021-08-24
Revise Tests for API & Old Projects

terrymyc:
To accommodate the new state file, this PR revises the testing of API and project files created in previous versions.

### DIFF
--- a/asreview/state/sql_converter.py
+++ b/asreview/state/sql_converter.py
@@ -130,7 +130,7 @@ def convert_project_json(project_fp, review_id, start_time,
         {
             'id': review_id,
             'start_time': start_time,
-            'review_finished': project_info['reviewFinished']
+            'review_finished': project_info.get('reviewFinished', False)
         }
     ]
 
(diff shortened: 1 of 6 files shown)


J535D165 (approved):

asreview/state/sql_converter.py
```diff
@@ -62,7 +62,7 @@ def convert_asreview(fp):
 
     # Create the reviews folder and the paths for the results and settings.
     Path(fp, 'reviews', review_id).mkdir(parents=True)
-    sql_fp = Path(fp, 'reviews', review_id, 'results.sql')
+    sql_fp = str(Path(fp, 'reviews', review_id, 'results.sql'))
```
PeterLombaers, line 65: Any particular reason for turning this into a string?

PeterLombaers (approved):

asreview/state/sql_converter.py
terrymyc: `sqlite3.connect()` does not accept a `PosixPath` argument.

PeterLombaers: Ah okay! This seems to be solved in python 3.7 (https://bugs.python.org/issue33496), but then it's probably smart to keep it a string.

terrymyc: Hmm... then I guess our test workflow needs an update. @J535D165 
https://github.com/asreview/asreview/blob/master/.github/workflows/ci-frontend.yml#L10

PR #711, opened 2021-08-26
Deprecate the use of Python 3.6

J535D165:
`pandas` requires 3.7 nowadays.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ def get_long_description():
 
 DEPS = {
     "sbert": ['sentence_transformers'],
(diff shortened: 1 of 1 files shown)


terrymyc:
@J535D165 This PR caused `test-core` failure on `master` branch because of the conflict between `tensorflow` and `numpy` packages. See https://github.com/tensorflow/models/issues/9706#issuecomment-774319552.
It is an ongoing issue and has not been resolved yet (https://github.com/tensorflow/tensorflow/pull/48935).

bnavigator:
You can (and should) use numpy 1.19 with python 3.8, too. No need for python 3.6.

PR #720, opened 2021-09-02
Fix creating project with name in non-alphabetic language

terrymyc:
This PR fixes an issue with project id generation failing for non-alphabetic languages.

### DIFF
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -165,7 +165,7 @@ def api_init_project():  # noqa: F401
     project_description = request.form['description']
     project_authors = request.form['authors']
 
-    project_id = re.sub(r'\s+', '-', project_name).lower()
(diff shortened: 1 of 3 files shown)


J535D165:
Can you provide an example to test with. I think your change now breaks for project names with characters that are not allowed for Windows or POSIX file names. For example, what happens with `my?project`? 

https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

terrymyc:
@J535D165 Thanks for the reminder. You are right. I changed the regular expression to detect invalid file name characters.

J535D165:
I think using https://tedboy.github.io/flask/generated/werkzeug.secure_filename.html might be a good solution. This is already used by asreview. 

Double question mark is interesting, but maybe not a priority. Only dashes is rare I think.

J535D165 (approved):

J535D165:
Thanks @terrymyc, nice teamwork on this one.

PR #728, opened 2021-09-08
Fix unit tests due to breaking gensim release

J535D165:
See https://github.com/RaRe-Technologies/gensim/issues/3226.


terrymyc (approved):
Perfect! Thanks!

PR #731, opened 2021-09-08
Add ivar documentation for data object and render it for autodoc

JTeijema:
This PR adds the variables for the data base.py file to the documentation, and sets ivar to True in the conf.py file so they render correctly. It would render under https://asreview.readthedocs.io/en/latest/API/reference.html#asreview.ASReviewData.


J535D165:
Like this PR. Important to document attributes 👍

PR #743, opened 2021-09-20
Add new API reference style and extension plugin guide

JTeijema:
So, this in a new implementation of the API reference, formatted in a tree structure as opposed to a list. It also contains an updates extension created guide.

Yet to do is a hunt for dead links, but I wanted to start the pull request for review.

Note: the API reference can be automated using the :recursive: trigger in the autosummary. I've tested this and it looks very promissing, but I wanted to get this implemented first.


JTeijema:
and also to fix the test errors apparently

ottomattas:
Good work!

I love the tree structure while I am not convinced the extension documentation should be a branch from API.
Just a question to get some more insight on your thought process - have you considered splitting extensions away from the API logically? I mean, instead of `docs/source/API/extensions_dev.rst` + `docs/source/API/reference.rst`, you would instead do something like `docs/source/extensions/development.rst` + `docs/source/API/reference.rst`.

(Why I think this is worth considering: we can then simply add additional docs without worrying about logical confusion. Think FAQs, for example. `docs/source/API/extensions_dev/FAQ.rst` + `docs/source/API/FAQ.rst` vs `docs/source/extensions/FAQ.rst` + `docs/source/API/FAQ.rst`)

JTeijema:
@ottomattas Are we looking at the same version? I don't think docs/source/API/extensions_dev.rs is part of this PR anymore hahaha. I came to the same conclusion as you :)

ottomattas:
You are right @JTeijema!
I saw there was a huge diff for those deprecated files which I did not open. Actually the diff was that they are deleted.

JTeijema:
And yeah, I totally agree with the extension being out of place in the API hahaha.

docs/source/extensions/extension_development.rst
```diff
@@ -0,0 +1,212 @@
...
+
+
+
+
+
+
+
+
+
+
```
J535D165, line 97:
```suggestion
```

Please remove the additional linebreaks.

docs/source/extensions/extension_development.rst
```diff
@@ -0,0 +1,212 @@
...
+Dataset extensions
+------------------
+An extension of the :class:`asreview.datasets.BaseDataSet` class.
+
+Dataset extensions integrate new datasets for use in ASReview. Adding datasets
+via extension provides quick access to the dataset via commandline or other user
+interface.
```
J535D165, line 186:
```suggestion
Dataset extensions integrate new datasets for use in ASReview. Adding datasets
via extension provides quick access to the dataset via Command Line Interface or in
ASReview LAB.
```

Let's add cross-references to the relevant parts of CLI and LAB.

docs/source/extensions/extension_development.rst
```diff
@@ -0,0 +1,212 @@
...
+
+.. code:: bash
+
+    ├── README.md
+    ├── asreviewcontrib
+    │   └── dataset_name
+    │       ├── __init__.py
+    │       └── your_dataset.py
+    ├── config
+    │   └── your_dataset.json
```
J535D165, line 199: I would suggest not use the config option for this example. This is pretty advanced usage. This was used in ASReview-covid to make automated updates available. Most use cases don't need this I think.

docs/source/extensions/extension_development.rst
```diff
@@ -0,0 +1,212 @@
...
+    ├── datasets
+    │   └── dataset_name
+    │       ├── __init__.py
```
J535D165, line 202: No need to have an `__init__.py` file in a data folder.

docs/source/extensions/extension_development.rst
```diff
@@ -0,0 +1,212 @@
...
+
+    ├── README.md
+    ├── asreviewcontrib
+    │   └── dataset_name
+    │       ├── __init__.py
+    │       └── your_dataset.py
+    ├── config
+    │   └── your_dataset.json
+    ├── datasets
+    │   └── dataset_name
```
J535D165, line 201:
```suggestion
    ├── data
```

docs/source/extensions/extension_development.rst
```diff
@@ -0,0 +1,212 @@
...
+    │       ├── __init__.py
+    │       └── your_dataset.csv
+    ├── setup.py
+    └── tests
+
```
J535D165, line 206: I think we need an example of `your_dataset.py` here. Covid plugin is relatively complicated.

I think a minimal example would look like this:

```
from asreview.datasets import BaseDataSet
from asreview.datasets import BaseDataGroup

class YourDataGroup(BaseDataGroup):
    group_id = "your_data_group"
    description = "A new data group with my awesome datasets."

    def __init__(self):
        
        dataset = BaseDataSet.from_config({
            "dataset_id": "your_data_id",
            "url": "",
            "reference": "",
            "link": "",
            "license": "",
            "title": "Your Data",
            "authors": [
              "Jane Doe",
              "John Doe"
            ],
            "year": 2021,
            "topic": "Your topic",
            "final_inclusions": true,
            "title_abstract_inclusions": false
          }
        )

        super(YourDataGroup, self).__init__(dataset)
        # pass multiple datasets to init if there are more datasets
```

docs/source/extensions/overview_extensions.rst
```diff
@@ -72,3 +69,28 @@ The following extensions are developed and maintained by the ASReview community:
 If an extension is not on this list, or you made one and you would like it to 
 be added to this list, please initiate an issue on `Github
 <https://github.com/asreview/asreview/issues>`__.
+
+
+Usage
+-----
```
J535D165, line 75: "Installation" would fit the context.

docs/source/index.rst
```diff
@@ -66,6 +66,8 @@ A (citable) PDF of the ASReview documentation can be found on `Zenodo <https://d
 
    extensions/overview_extensions
 
+   extensions/extension_development
```
J535D165, line 69: @Rensvandeschoot I need your help here. I think this belongs to the development section (see below).

docs/source/extensions/extension_development.rst
```diff
@@ -0,0 +1,212 @@
...
+For minimal functionality, ``your_dataset.py`` should extent
+:class:`asreview.datasets.BaseDataSet`. Further functionality can be extensions
+of any other class in :mod:`asreview.datasets`.
```
J535D165, line 209: I would suggest creating a template repo here. Like https://github.com/asreview/template-extension-new-model

This should be straightforward with the example above.

docs/source/extensions/extension_development.rst
```diff
@@ -0,0 +1,212 @@
...
+    │   └── your_dataset.json
+    ├── datasets
+    │   └── dataset_name
+    │       ├── __init__.py
+    │       └── your_dataset.csv
+    ├── setup.py
+    └── tests
+
+For minimal functionality, ``your_dataset.py`` should extent
+:class:`asreview.datasets.BaseDataSet`. Further functionality can be extensions
```
J535D165, line 208: Maybe also explain what `asreview.datasets.BaseDataGroup` does?

J535D165 (changes requested):
Thanks a lot for this PR. I love the new reference style. Relatively minor feedback to address.

J535D165:
Any idea why the tests fail?

docs/source/index.rst
JTeijema: I've moved it to development. (Also to manage broken links)

JTeijema:
As there is only one documented exception, I am removing the exception block from the template:
```
   {% block exceptions %}
   {% if exceptions %}
   .. rubric:: {{ _('Exceptions') }}

   .. autosummary::
      :toctree:
   {% for item in exceptions %}
      {{ item }}
   {%- endfor %}
   {% endif %}
   {% endblock %}
```

The only exception listed by the API reference was: `asreview.datasets.DataSetNotFoundError`

JTeijema:
Do we want with or without signatures?
afbeelding

afbeelding


I'm thinking without. 

Also, the error is within the template, so I guess I'm learning Jinja

JTeijema:
See: 
afbeelding

JTeijema:
Yes ofcourse because module calls class. ^

JTeijema:
Okay so both classes and func triggers the problem

JTeijema:
Oh c'mon. It cant be both those things. Those are totally unrelated things.

JTeijema:
THIS IS THE SAME AS BEFORE WHY ARE YOU BUILDING?

Anyway, test's fixed :)

J535D165:
Thanks!!

docs/source/API/extension_dev.rst
```diff
@@ -0,0 +1,193 @@
...
+example`` subcommand. See ``asreview -h`` for this option.
+
+An example of a subcommand extension is the `Visualization Extension
+<https://github.com/asreview/asreview-visualization>`_
+
+Dataset extensions
+------------------
+An extension of the :class:`asreview.datasets.BaseDataSet` class.
+
+Dataset extensions integrate new datasets for use in ASReview. Adding datasets
```
Rensvandeschoot, line 164: Can you maybe add an explanation how the 'dataset' as used here, is different from a dataset you can import or a benchmark dataset you can see in the exploration mode?

PR #746, opened 2021-09-22
Fix frontend components location changes

terrymyc:
This PR fixed error brought by the change of file location.


J535D165:
Thanks, sorry for messing up :|

PR #748, opened 2021-09-22
Remove use of labeled.json for stats and rename functions

J535D165:


terrymyc (commented):
Thanks. The corresponding tests in `test_api.py` also need to be modified.

asreview/webapp/api.py
```diff
@@ -1046,23 +1046,22 @@ def api_get_progress_history(project_id):
...
+def api_get_progress_recall(project_id):
     """Get cumulative number of inclusions by ASReview/at random"""
 
     try:
-        statistics = get_data_statistics(project_id)
-        labeled = read_label_history(project_id)
-        data = []
-        for [key, value] in labeled:
-            data.append(value)
+
```
PeterLombaers, line 1054: This needs `project_path = get_project_path(project_id)`.

PeterLombaers (commented):
One small issue, but looks good for the rest.

asreview/webapp/api.py
J535D165: Thanks

PR #753, opened 2021-09-24
Update overview_extensions.rst

JTeijema:
As requested, added the extension from #650 to the list.


J535D165:
Thanks! Interesting project.

PR #758, opened 2021-09-29
Fix breaking frontend when stats wont load

J535D165:


terrymyc (changes requested):
The problem is not in the front end. Only the `api_get_projects_stats` needs to be changed.
Change the line 188 into:
`elif "reviewFinished" not in res or res["reviewFinished"] is not True:`
https://github.com/asreview/asreview/blob/7b6b6f222f60dccb7cb32bd32a0b33ce8608bd7f/asreview/webapp/api.py#L188

terrymyc:
Adding `isError` to detect the error instead of replacing `isFetched` with `isSuccess`. `isSuccess` does not guarantee the data has been fetched, which will break the frontend as well.
I will make another PR for this:

> The problem is not in the front end. Only the `api_get_projects_stats` needs to be changed. Change the line 188 into: `elif "reviewFinished" not in res or res["reviewFinished"] is not True:`
> 
> https://github.com/asreview/asreview/blob/7b6b6f222f60dccb7cb32bd32a0b33ce8608bd7f/asreview/webapp/api.py#L188

J535D165:
Replaced by #759

PR #770, opened 2021-10-04
State column names singular. pool_labeled_function.

PeterLombaers:
I turned the state column names to singular and added a function to get the pool and labeled record_ids in one query. I did not connect the function to the rest of the code yet.


J535D165:
Thanks for this PR. I think this will simplify some work on the middleware and top layers of ASReview (and maybe break some things for now). I will test this after the lint errors are resolved.

J535D165 (approved):

J535D165:
Thanks for this effort. Makes ASReview more future-proof!

PR #773, opened 2021-10-04
Adjust test for old asreview files

terrymyc:


J535D165 (approved):

J535D165:
Thanks. This works fine now.

PR #775, opened 2021-10-05
update cli model list with working links

JTeijema:
Did not double check the requires column.


JTeijema:
afbeelding

J535D165:
Let's fix the source of the broken links. (Should be the public API, over the internal API)

PR #785, opened 2021-10-07
Improve API reference structure for public ASReview API

jteijema:
I've adjusted the previous API settings from showing internal API to using the public API where possible, as requested in #743. Note that the public API is only usable when initialised, meaning that uninitialised objects still need to be refered to by module:

afbeelding

One final decision; should the methods include a signature like it does in the SKLearn API? It can be adjusted with the       :nosignatures: feature in the class template.

afbeelding
 
 vs
 
 
afbeelding

### DIFF
--- a/docs/source/API/reference.rst
+++ b/docs/source/API/reference.rst
@@ -37,6 +37,14 @@ Functions
    :toctree: generated/
    :template: custom-module-template.rst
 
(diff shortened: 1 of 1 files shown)


JTeijema:
@J535D165 I think this is still relevant for v0.18

Issue #499, opened 2021-02-21
ASReview Analysis submodule struggles with simulations with exactly 1 inclusion. 


J535D165:
More specific: If the number of relevenant items in the dataset matches the number of relevant items in the prior knowledge, the submodule ASReview Analysis struggles with computing metrics. This pops up when using ASReview-statistics. 

Example: Dataset with 1 relevant record which is used ats prior knowledge. Simulate run and then call `asreview stats`. 

```
jonathan$ asreview stat test.h5 
/Users/jonathan/asreview/asreview/analysis/analysis.py:170: RuntimeWarning: invalid value encountered in true_divide
  norm_yr = (np.array(self.inc_found[fl]["avg"]) - dy) / y_norm
/Users/jonathan/asreview/asreview/analysis/analysis.py:171: RuntimeWarning: invalid value encountered in true_divide
  norm_y_err = np.array(self.inc_found[fl]["err"]) / y_norm
Traceback (most recent call last):
  File "/Users/jonathan/.pyenv/versions/asreview/bin/asreview", line 11, in <module>
    load_entry_point('asreview', 'console_scripts', 'asreview')()
  File "/Users/jonathan/asreview/asreview/__main__.py", line 80, in main
    entry.load()().execute(sys.argv[2:])
  File "/Users/jonathan/asreview-statistics/asreviewcontrib/statistics/entrypoint.py", line 71, in execute
    print(stat)
  File "/Users/jonathan/asreview-statistics/asreviewcontrib/statistics/statistics.py", line 99, in __str__
    results = self.to_dict()
  File "/Users/jonathan/asreview-statistics/asreviewcontrib/statistics/statistics.py", line 89, in to_dict
    "loss": self.loss(),
  File "/Users/jonathan/asreview-statistics/asreviewcontrib/statistics/statistics.py", line 61, in loss
    return loss/len(avg_discovery_time)
ZeroDivisionError: division by zero

```

Rensvandeschoot:
Indeed, it should result in a warning

J535D165:
Resolved in #688

PR #787, opened 2021-10-08
Delete analysis API

J535D165:
This PR deprecates `asreview.analysis`. The functionality is moved to asreview-insights (previously asreview-visualization)

TODO: update external packages like asreview-statistics and asreview-visualization.


JTeijema:
Do you think that this will fix issue #499?

PR #795, opened 2021-10-13
Remove labeled.json and pool.json from project files

J535D165:


J535D165:
Thanks @PeterLombaers for all the work. This is a major milestone and hopefully the last big step in the ASReview state file update.

PR #799, opened 2021-10-17
Sort import of dependencies and add import linter

J535D165:
Adds import linter with flake8-import-order. Ignoring I201. 

Applies `isort  -m 7 --sl` to all files.


JTeijema:
Much needed :). Thank you.

PR #801, opened 2021-10-17
Deprecate data_name and data_type in ASReviewData

J535D165:
Remove data_name and data_type because there is no clear use for them. 

This PR might break some (external) extensions. asreview-statistics might be one of the them. This package usages the data_name for output in the CLI. My point of view is that this should be handled in the extensions, and not in the API.


JTeijema:
It's found in [settings ](https://asreview.readthedocs.io/en/latest/API/generated/asreview.settings.ASReviewSettings.html)and in [data](https://asreview.readthedocs.io/en/latest/API/generated/asreview.data.ASReviewData.html#asreview.data.ASReviewData). I am having some trouble with `pip install .` after removal so can't say that it works now.

PR #814, opened 2021-11-04
Remove icon on dashboard number card

terrymyc:
This PR removed the icons on the dashboard number cards. The icons were more decorative than functional.

image


J535D165:
Thanks :)

PR #828, opened 2021-11-10
Fix inconsistent color on tooltip of progress recall chart

terrymyc:
This PR fixed a minor color inconsistency on the tooltip of the progress recall chart.


J535D165:
Good catch.

PR #846, opened 2021-11-23
Extend REST API /labeled with option to subset and order

J535D165:
Examples:

```
http://localhost:5000/api/project/test-performance/labeled?subset=relevant&page=1&latest_first=1
```

```
http://localhost:5000/api/project/test-performance/labeled?subset=excluded&page=1&latest_first=0
```


terrymyc (changes requested):
`http://localhost:5000/api/project/test-performance/labeled?subset=relevant&page=2&latest_first=1`

If your project has `next_page == 2`, meaning there are more than 20 labeled records, it seems that you cannot go to the next page.

terrymyc (approved):

PR #847, opened 2021-11-24
Add RIS file export option

ottomattas:


ottomattas:
Instead of merging only the RIS export feature, we are merging the current state of master branch which already includes the RIS export feature.

Follow it here:
https://github.com/asreview/asreview/pull/848

PR #853, opened 2021-11-25
Add error handler for exporting review results

terrymyc:
This PR added an error handler for exporting review results.

image


J535D165:
Nice! Code with the Blob looks more complex than I expected.

Issue #840, opened 2021-11-19
Return error when dataset is successfully edited from file in setup


terrymyc:
When a dataset has been added from file, adding another dataset from the file succeeds but returns the error `Labeling priors, but not all record_ids were found in the pool.`
This was introduced from #795.

J535D165:
Thanks for reporting. I think I can resolve this for you soon.

terrymyc:
Closed by #854

PR #854, opened 2021-11-26
Delete state from project when changing dataset

J535D165:
Not tested yet.

### DIFF
--- a/asreview/state/utils.py
+++ b/asreview/state/utils.py
@@ -200,6 +200,40 @@ def open_state(working_dir, review_id=None, read_only=True):
             pass
 
 
+def delete_state_from_project(project_path):
+
+    project_file_path = get_project_file_path(project_path)
+
+    try:
+        path_fm = get_feature_matrices_path(project_path)
+        shutil.rmtree(path_fm)
+    except Exception:
+        print("Failed to remove feature matrices.")
+
+    try:
+        path_review = get_reviews_path(project_path)
+        shutil.rmtree(path_review)
+    except Exception:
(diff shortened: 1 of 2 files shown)


asreview/state/utils.py
```diff
@@ -200,6 +200,39 @@ def open_state(working_dir, review_id=None, read_only=True):
...
+    with open(project_file_path, "r") as f_read:
+        project_dict = json.load(f_read)
+
+    # add path to dict (overwrite if already exists)
+    reset_config = {
+        'projectInitReady': False,
+        'reviewFinished': False,
+        'reviews': [],
+        'feature_matrices': []
+    }
```
terrymyc, line 229: The key `projectHasDataset` has been deprecated? Now when the value of `reviews` is an empty array, it means no dataset?

terrymyc (approved):

asreview/state/utils.py
J535D165: I don't think so. `projectHasDataset` is generated on the fly in 

https://github.com/asreview/asreview/blob/6af53348b67c16b0c6579a4871ff0b6137e351b2/asreview/webapp/api.py#L198-L203

`reviews` doesn't contain datasets. We decided to put only one dataset in a project, but with the option to have multiple reviews (useful for simulation in the future).

J535D165:
Not clear to me why the tests fail :(

J535D165:
Fixed the tests. There was a boolean wrong.

terrymyc:
Close #840

Issue #822, opened 2021-11-09
Integrate project setup with the new V1 layout


J535D165:

terrymyc:
Closed by #852

PR #857, opened 2021-11-27
Remove dataset when uploading new dataset

J535D165:


J535D165:
@terrymyc FYI. You can merge with PR #822.

PR #866, opened 2021-12-02
Fix adding extension dataset when setting up a project

terrymyc:
Closes #830

Issue #863, opened 2021-12-01
Error giving a BaseDataSet object a filepath


JTeijema:
**Describe the bug**
BaseDataSet objects have a filepath (fp) argument during construction (as can be seen [in the source](https://asreview.readthedocs.io/en/latest/_modules/asreview/datasets.html#BaseDataSet)). However, setting the filepath and not an URL, results in a failure to load as urls are mandatory in the front-end. The error message results in a failure to find `dataset.url`.

**To Reproduce**
Steps to reproduce the problem:
Install ASReview
Install a custom plugin, with the following configuration:
```python
"dataset_id": "test",
"fp": "../file.csv"
```
Start the front-end using asreview lab
Select the plugin as dataset

**Version information**
 - OS: Windows
 - ASReview version 0.17, 0.18

Issue #862, opened 2021-12-01
Request for more documentation on BaseDataSet.from_config(config_file)


JTeijema:
## Feature Request
When creating a dataset plugin, there is no documentation available on the [from_config](https://asreview.readthedocs.io/en/latest/API/generated/asreview.datasets.BaseDataSet.html#asreview.datasets.BaseDataSet.from_config) function, even though it is needed in creating a dataset extension. This request is for an explanation on how to use from_config.

As from_config's only function is to create a BaseDataSet with the config file as attributes, the real missing documentation is on what attributes can be given to such an object.

**Describe the solution you'd like**
The solution would be documentation found on either the template repository and/or the readthedocs.

Issue #861, opened 2021-12-01
Extending BaseDataSet class results in front-end errors in 0.17 and 0.18


JTeijema:
**Describe the bug**
When implementing a BaseDataSet in an extension for ASReview, the following happens:

![image](https://user-images.githubusercontent.com/28191416/144274304-9e73a1b3-75b3-4049-9131-00015b2cb2cb.png)

The following message appears in the front-end POST request:
![image](https://user-images.githubusercontent.com/28191416/144274525-f5c13aea-6f1a-4512-84d3-d5cec4f6149f.png)

Using a BaseVersionedDataSet however, does work. Using the extension dataset from the command line also works.

**To Reproduce**
Steps to reproduce the problem:
Install ASReview
Install the following test repository: https://github.com/JTeijema/Test-Dataset-Extension
Launch ASReview Lab
Create a new project
Select the new dataset as input data

**Version information**
 - OS: Windows
 - ASReview version 0.17

**Additional information**
The plugin datasets are available from the url location:

![image](https://user-images.githubusercontent.com/28191416/144275150-c49bfd79-9cc5-4e51-a31c-d4286ac6b61f.png)

PR #867, opened 2021-12-02
Remove VersionedBaseDataset and refacter datasets submodule

J535D165:
This PR removes the `VersionedBaseDataset` and simplifies/pythonizes the `datasets` API. This will break the current template format. 

```
from asreview.datasets import BaseDataSet
from asreview.datasets import BaseDataGroup

class YourDataGroup(BaseDataGroup):
    group_id = "your_data_group"
    description = "A new data group with my awesome datasets."

    def __init__(self):

        dataset = BaseDataSet(
            dataset_id= "your_data_id",
            url= "",
            title= "Your Data",
            # ... other param 
        )

        super(YourDataGroup, self).__init__(dataset)
        # pass multiple datasets to init if there are more datasets
```

@JTeijema will ask you for review soon.


J535D165:
Relates to #867 #866 #861 #862 #863 (maybe more? please link)

tests/test_datasets.py
```diff
@@ -0,0 +1,73 @@
...
+    class TemplateDataGroup(BaseDataGroup):
+        group_id = "template"
+        description = "Template group"
+
+        def __init__(self):
+
+            super(TemplateDataGroup, self).__init__(
+                my_dataset1, my_dataset2
+            )
+    # END
```
J535D165, line 68: This test tests the pipeline of https://github.com/asreview/template-extension-new-dataset. Let's make a V1 branch in this repo as well to prepare for the V1 release.

asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/BenchmarkDataset.js
```diff
@@ -88,12 +88,12 @@ const BenchmarkDataset = (props) => {
...
             DOI:{" "}
             <Link
-              href={DOILink(props.doi)}
+              href={props.doi && DOILink(props.doi)}
               underline="none"
               target="_blank"
               rel="noreferrer"
             >
-              {props.doi}
+              {props.doi && props.doi}
```
J535D165, line 96:
```suggestion
              {props.doi}
```

J535D165:
Our official datasets extension on COVID19 works with the new API if the following PR is merged https://github.com/asreview/asreview-covid19/pull/41

J535D165:
@terrymyc I really struggled with some merge conflicts. Can you properly test? I think there is still an issue with non-unique `keys` somewhere, but can't figure out why. You might have to install the PR of the covid19 plugin to test it properly. 

@JTeijema Can you do a review and test as well. This PR should resolve many issues with the plugin datasets.

asreview/datasets.py
```diff
@@ -300,77 +292,103 @@ def find(self, dataset_name):
...
+                        raise err
+
+            return dataset_list_ser
+
+        return group_list
 
 
 class BenchmarkDataGroup(BaseDataGroup):
     group_id = "benchmark"
     description = "A collections of labeled datasets for benchmarking."
```
J535D165, line 373: What is a good description (displayed in the GUI)?

asreview/datasets.py
```diff
@@ -300,77 +292,103 @@ def find(self, dataset_name):
...
-    try:
-        with urlopen(url, timeout=10) as f:
-            meta_data = json.loads(f.read().decode())
-    except URLError as e:
-        if isinstance(e.reason, socket.timeout):
-            raise Exception("Connection time out.")
-        raise e
+class NaturePublicationDataGroup(BaseDataGroup):
+    group_id = "benchmark-nature"
+    description = "Featured benchmarking datasets from the Nature publication."
```
J535D165, line 386: What is a good description (displayed in the GUI)?

terrymyc: Featured datasets from Nature

asreview/datasets.py
terrymyc: It might not be necessary to display the description of non-featured datasets. Such description is repeated with the text in the GUI.

terrymyc:
@J535D165 Can we also use `extension` rather than `plugin` in the backend?

terrymyc:
> I think there is still an issue with non-unique `keys` somewhere, but can't figure out why. You might have to install the PR of the covid19 plugin to test it properly.

Fixed and tested.

asreview/datasets.py
```diff
@@ -29,203 +29,195 @@ class DataSetNotFoundError(Exception):
...
-            f"{str(data)}\n\n" for data in self._data_sets
-        ])
+        dataset: asreview.datasets.BaseDataSet
+            A asreview BaseDataSet-like object.
+        """
+        if not issubclass(dataset, BaseDataSet):
+            raise ValueError(
+                "Expected BaseDataSet or subclass of BaseDataSet."
+            )
+        self.datasets.append(dataset)
```
JTeijema, line 189: issubclass() checks whether the argument is a class, not an instance of a class. This means that appending a dataset is not possible:

image

image

asreview/datasets.py
```diff
@@ -29,203 +29,195 @@ class DataSetNotFoundError(Exception):
...
+        for d in self.datasets:
+            if dataset_id.lower() == d.dataset_id.lower() or \
+                    dataset_id.lower() in [a.lower() for a in d.aliases]:
+                results.append(d)
+
         if len(results) > 1:
             raise ValueError(
                 f"Broken dataset group '{self.group_id}' containing multiple"
-                f" datasets with the same name/alias '{dataset_name}'.")
+                f" datasets with the same name/alias '{dataset_id}'.")
```
JTeijema, line 213: Shouldn't this check be an assertion at the initialisation of the object? A private "check for duplicates" function that is called during init and when appending.

Something like 
```python
    def _assert_unique_dataset_ids(self):
        for dataset in self.datasets:
            assert dataset.dataset_id not in self.list()
```

JTeijema (changes requested):
This looks very good. Very intuitive to work with. Thanks!

asreview/datasets.py
JTeijema: When a list, append and find function are available, then shouldn't `self.datasets` be a private attribute?

asreview/datasets.py
J535D165: I removed the text in the GUI. This might be very useful for the extension data tab. It distinguishes the datasets from various plugins.

asreview/datasets.py
```diff
@@ -256,34 +248,34 @@ def find(self, dataset_name):
             If the dataset with that name is found, return it
```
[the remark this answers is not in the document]
J535D165: Thanks

asreview/datasets.py
JTeijema: Maybe I am using the function wrong! I should be giving classes. One second, let me test this.

asreview/datasets.py
```diff
@@ -29,203 +29,195 @@ class DataSetNotFoundError(Exception):
...
+            'img_url': self.img_url,
+            'aliases': self.aliases,
+            **self.kwargs
+        }
 
-    def list(self, latest_only=True):
-        """Return a list of itself."""
-        return [self]
 
+class BaseDataGroup():
```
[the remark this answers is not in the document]
J535D165: Made an ABC out of now :)

asreview/datasets.py
terrymyc: Good point. I added text for extension datasets.

terrymyc (approved):

asreview/datasets.py
```diff
@@ -29,203 +29,195 @@ class DataSetNotFoundError(Exception):
...
-        all_dataset_names = [(d, d.aliases) for d in self.datasets]
-        for dataset, aliases in all_dataset_names:
-            if dataset_name.lower() in aliases:
-                return dataset
-
-        raise DataSetNotFoundError(
-            f"Dataset {dataset_name} not found"
-        )
+    def __dict__(self):
+        return {d.dataset_id: d for d in self.datasets}
```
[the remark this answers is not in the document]
J535D165: Maybe you are right. I hope to work on this later. It's internal API I think atm.

PR #873, opened 2021-12-03
Disable RIS export option when imported dataset is not RIS

terrymyc:
This PR disabled the option to export the RIS file when the imported dataset is not a RIS file. This is controlled dynamically through the REST API.

image


JTeijema:
Looks like the perfect solution :)

asreview/webapp/src/ProjectComponents/ExportComponents/ExportPage.js
```diff
@@ -91,6 +93,10 @@ const ExportPage = (props) => {
     setExporting(true);
   };
 
+  const disableRIS = () => {
+    return props.dataReader !== "ris-reader";
+  };
+
```
J535D165, line 99: This should be moved to the class/backend.

asreview/webapp/src/ProjectComponents/ExportComponents/ExportPage.js
```diff
@@ -91,6 +93,10 @@ const ExportPage = (props) => {
     setExporting(true);
   };
 
+  const disableRIS = () => {
+    return !props.dataWriter.includes("ris-writer");
+  };
```
terrymyc, line 98: @J535D165 Available writers are returned in the API, but we still need an indicator to disable an option.

ottomattas:
The screenshot will be added to the documentation with the following PRs: https://github.com/asreview/asreview/pull/990, https://github.com/asreview/asreview/pull/966.

terrymyc:
> The screenshot will be added to the documentation with the following PRs: #990, #966.

@ottomattas Which screenshot do you use? The one in this PR does not meet our [screenshot requirement](https://github.com/asreview/asreview/blob/master/DEVELOPMENT.md#screenshots)

ottomattas:
@terrymyc I used the screenshot you posted in your first comment in the current PR. Would you be able to share a screenshot that follows the requirements?

terrymyc:
@ottomattas Sorry for the late response. The RIS option will not be visible in the frontend if the imported dataset is not a RIS file. So no more screenshots for this PR...

asreview/io/csv_writer.py
```diff
@@ -0,0 +1,46 @@
...
+    fp: str, NoneType
+        Filepath or None for buffer.
+    sep: str
+        Seperator of the file.
+    labels: list, numpy.ndarray
+        Current labels will be overwritten by these labels
+        (including unlabelled). No effect if labels is None.
+    ranking: list
+        Reorder the dataframe according to these (internal) indices.
+        Default ordering if ranking is None.
```
J535D165, line 31: Actually, this should be in the `write_data` method. The arguments here should describe the `__init__`. I think there are none.

asreview/data/base.py
```diff
@@ -419,71 +455,3 @@ def to_dataframe(self, labels=None, ranking=None):
             result_df.loc[result_df[col_label] == LABEL_NA, col_label] = np.nan
 
         return result_df
-
-    def to_csv(self, fp, sep=",", labels=None, ranking=None):
-        """Export to csv.
-
```
J535D165, line 425: Makes sense.

asreview/data/base.py
```diff
@@ -362,14 +396,16 @@ def to_file(self, fp, labels=None, ranking=None):
         ranking: list, numpy.ndarray
             Optionally, dataframe rows can be reordered.
         """
+        df = self.to_dataframe(labels=labels, ranking=ranking)
+
         if Path(fp).suffix in [".csv", ".CSV"]:
```
J535D165, line 401: This should be based on the available entry points isn't it?

J535D165: Like in from_file

J535D165 (changes requested):
Great work, some suggestions, and questions. Maybe more will follow.

asreview/data/base.py
terrymyc: Should `export_to_string` function in `project.py` be integrated into `to_file`?

J535D165 (approved):

J535D165:
Great work and what a journey :)

PR #881, opened 2021-12-08
Fix dashboard stats value not JSON serializable

terrymyc:
This PR resolved the problem of getting dashboard stats for old projects and added a test.

### DIFF
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -207,9 +207,6 @@ def api_get_projects_stats():  # noqa: F401
 
     project_stats = dict(stats_counter)
 
-    for keys in project_stats:
-        project_stats[keys] = int(project_stats[keys])
-
     response = jsonify({"result": project_stats})
     response.headers.add('Access-Control-Allow-Origin', '*')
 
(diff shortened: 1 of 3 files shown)


J535D165:
Given the reuse potential of `get_statistics`, I think it is better to change the returned types of that function. 

https://github.com/asreview/asreview/blob/137a7849fcf6dc295c965fea9833a58237e316c9/asreview/webapp/utils/project.py#L488-L494

I think the following will also solve the issue:

```python
n_included = int(sum(labels == 1))
 n_excluded = int(sum(labels == 0))
  
 if len(labels) > 0: 
     n_since_last_relevant = int(labels.tolist()[::-1].index(1))
 else: 
     n_since_last_relevant = 0 
```

terrymyc:
> Given the reuse potential of `get_statistics`, I think it is better to change the returned types of that function.

Good point. Thanks!

J535D165 (approved):

J535D165:
Thanks! Shall we add the test in another PR?

J535D165:
Nice!

PR #883, opened 2021-12-09
Displace quick tour

terrymyc:
This PR displaced the quick tour. It has been placed in the project setup dialog when the model is training.


JTeijema (approved):

SagevdBrand (changes requested):
I really really love the way you did this, gives me some Sims Vibes hahaha! Great job on the balloons at the end!

Some suggestions:
- Maybe add a button to manually go to the quick tour somewhere in the menu on the left? If people are using the software for the first time, they might not know yet what they should do! 
- When waiting for the algorithm to load, would it be possible to show a progress bar as well?

Perhaps not related to this specific PR:
1. When I created the project, I wanted to go back to the beginning, to go to exploration mode instead of oracle mode, but I could not change it anymore? Is there a way to allow for changing the screening mode?

2. When going back to the previous setup page I would also like to be able to click on the progress bar to go back. For example clicking on details will send you back to the details page.
image

MartijnUX:
Hi Sofie, good suggestion. 

1. Nevertheless it is the last step in starting your review/screening. When the researcher wants to change the model, you should start a new project.
2. For the Setup, we choose for a one page. 3 steps is not that much. With 4 a stepper (your suggestion) would be great. We could take you suggestion in the Usability test for the next big release.

My UX suggestion would be, for the end of the Setup to go away of the screen en see the projectpage (filling in). Blank to fill. It will give Terry an example.

MartijnUX:
@terrymyc Can we work on a feature onboarding for the next big release? This way we know for sure, the change to another UI will be smooth and supportive.

J535D165:
@SagevdBrand we don't have any idea how to implement a progress bar yet. I don't think this will be an option.

SagevdBrand:
> Nevertheless it is the last step in starting your review/screening. When the researcher wants to change the model, you should start a new project.

@MartijnUX, I think we might be talking about different things. Perhaps I did not explain clearly enough, so let me try again. When setting up the project I made a simple mistake, I wanted to use a benchmark dataset, but used oracle mode.  I realized my mistake in the second step already, because I could not use the benchmark dataset. That is why I wanted to change screening mode, not the model. 

To be honest, I find it a bit odd that is allowed to go back and change the title, but not the mode, which is on the same page. Anyway, I hope this is more clear now!

terrymyc:
@SagevdBrand Thanks!

> To be honest, I find it a bit odd that is allowed to go back and change the title, but not the mode, which is on the same page. Anyway, I hope this is more clear now!

This is a workaround for now. We don't allow the change of mode because it would remove the dataset and prior knowledge you probably have added. This may be improved in the future.

> When going back to the previous setup page I would also like to be able to click on the progress bar to go back. For example clicking on details will send you back to the details page.

This may be possible but involves more changes. The current stepper is linear, which means you have to complete steps in sequence. I will look into it later.

SagevdBrand:
@terrymyc Ahh that makes sense! Thanks for your explanation.
Again, I really have to say that you did such a wonderful job! 😻

Rensvandeschoot (approved):

ottomattas:
Can we finalise this PR?

J535D165 (approved):

PR #898, opened 2021-12-16
Change folder structure for legacy files and change record_ids in converter

PeterLombaers:
Change the record table and converter.


PeterLombaers:
This fixes the concerning the order in which the records are presented. Also, when converter an old format project file, the old files are moved to a 'legacy' folder, so they can stilll be recovered if necessary.

One important issue is that the project deletion test is failing. I've disabled it for now, and propose to fix it in a new pull request.

J535D165 (approved):

J535D165:
Thanks a lot. Hopefully we can close this book now :)

PR #899, opened 2021-12-20
Add Sofie to .zenodo.json

SagevdBrand:
With this PR I added my own name to the .zenodo.json file.
That's all 👯


J535D165:
We are thinking about removing all authors from the zenodo publication and use "ASReview developers" or something similar instead. This is more inclusive given the number of contributors (of which many contributed on github, while others in different ways.) In our opinion.

If we change our minds, we will reopen this PR.

PR #923, opened 2022-01-13
Fix progress charts getting incorrect data

terrymyc:
This PR fixed progress density and recall charts getting incorrect data. It also fixed minor styling issues.


asreview/webapp/api.py
```diff
@@ -1135,6 +1135,7 @@ def api_get_progress_density(project_id):
         # create a dataset with the rolling mean of every 10 papers
         df = data \
             .to_frame(name="Relevant") \
+            .reset_index(drop=True) \
```
J535D165, line 1138: Thanks!

J535D165 (approved):

PR #924, opened 2022-01-13
Update footer to current year

JTeijema:
This PR fixes the documentation footer to show the current year instead of 2020.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -6,6 +6,8 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/master/config
 
+import datetime
+
 try:
     from asreview import __version__ as asreview_verion
 except Exception:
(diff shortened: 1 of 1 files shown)


JTeijema:
image

J535D165 (approved):

VeenDuco (commented):
Should it not be a range of years, see for instance a discussion [here](https://opensource.stackexchange.com/questions/2872/why-does-the-fsf-make-this-recommendation-about-ranges-of-copyright-years) or [scikit-learn docs](https://scikit-learn.org/stable/user_guide.html).

JTeijema:
> Should it not be a range of years, see for instance a discussion [here](https://opensource.stackexchange.com/questions/2872/why-does-the-fsf-make-this-recommendation-about-ranges-of-copyright-years) or [scikit-learn docs](https://scikit-learn.org/stable/user_guide.html).

@J535D165

J535D165:
Thanks for the feedback. Sounds good to me, but let's implement this for all files at the same time then.

VeenDuco:
See here for a discussion: https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

Bottom line: It's not needed. It helps to establish a claim, but in principle copyright holds for the life of the author plus 70 years in Europe. If you do mention a date, the first date of the document is relevant to leave there because it helps establish timelines. Otherwise if someone copies your materials, you update your date to a year later, it makes it harder to see you had the materials first.

PR #925, opened 2022-01-13
Extend REST API with option to add and update notes

J535D165:
#832 and #820. 

It requires also changes to the javascript code @terrymyc. Can you add that to this PR?

@peterlombaers, Can you review the "update_decision" method? Made this one slightly more flexible because everything is now properly handled in the javascript code (the difference between PUT and POST)

### DIFF
--- a/asreview/state/sqlstate.py
+++ b/asreview/state/sqlstate.py
@@ -620,24 +620,19 @@ class SqlStateV1(BaseState):
         con.commit()
         con.close()
 
-    def change_decision(self, record_id):
+    def update_decision(self, record_id, label, note=None):
         """Change the label of a record from 0 to 1 or vice versa."""
-        current_label = self.get_data_by_record_id(record_id)['label'][0]
-
-        # New label should be of type int, not np.int, to insert into sql.
-        new_label = int(1 - current_label)
-        current_time = datetime.now()
 
         con = self._connect_to_sql()
         cur = con.cursor()
 
         # Change the label.
-        cur.execute("UPDATE results SET label = ? WHERE record_id = ?",
-                    (new_label, record_id))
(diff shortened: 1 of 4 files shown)


asreview/state/sqlstate.py
```diff
@@ -620,24 +620,19 @@ def add_labeling_data(self, record_ids, labels, notes=None, prior=False):
...
         # Change the label.
-        cur.execute("UPDATE results SET label = ? WHERE record_id = ?",
-                    (new_label, record_id))
+        cur.execute("UPDATE results SET label = ?, notes = ? WHERE record_id = ?",
+                    (label, note, record_id))
 
         # Add the change to the decision changes table.
         cur.execute("INSERT INTO decision_changes VALUES (?,?, ?)",
-                    (record_id, new_label, current_time))
+                    (record_id, label, datetime.now()))
```
PeterLombaers, line 635: Do we want to update this table if we only add a note, but keep the label the same? Or we could add the notes to the decision_changes table, that way you also save the history of the notes. Note sure if that is desirable though.

PeterLombaers (commented):
Just one comment regarding the decision_changes table, other than that it looks good. I like it better this way, it is more flexible indeed.

asreview/state/sqlstate.py
J535D165: I was thinking about privacy-related issues. A decision doesn't harm anyone. However, a note can contain info like "Discuss with John Doe". The note is removed by the reviewer after the discussion, however, it's never removed from the history.

PeterLombaers: Yes, that makes sense. I agree that you should be able to remove notes, but not the previous labels.

terrymyc (approved):
I think I need to start another PR to implement the note functionality in the front end after this is merged.

PR #938, opened 2022-01-24
Fix get statistics of number since last relevant

terrymyc:
This PR fixed the `get_statistics` function if no relevant prior knowledge is added when creating a new project. It is not possible to return `n_since_last_relevant` if no relevant record is provided. The dashboard analytics uses this function and may return the error.


J535D165 (approved):

J535D165:
Smart.

PR #942, opened 2022-01-27
Rephrase exploration mode banner and labeled record alert

terrymyc:
This PR rephrased the text of the exploration mode banner on the project review page and the text of the existing label alert on the record card.

image

image

### DIFF
--- a/asreview/webapp/src/StyledComponents/StyledAlert.js
+++ b/asreview/webapp/src/StyledComponents/StyledAlert.js
@@ -7,7 +7,7 @@ export function ExplorationModeRecordAlert(props) {
       severity="info"
       sx={{ borderBottomRightRadius: 0, borderBottomLeftRadius: 0 }}
     >
-      Labeled as relevant in the dataset
+      {`Labeled as ${props.label} in the dataset`}
     </Alert>
   );
 }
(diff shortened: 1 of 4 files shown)


J535D165:
Thanks, I like the wording. 

Why don't we see this prior knowledge info box on all the cards?

J535D165 (approved):

terrymyc:
> Why don't we see this prior knowledge info box on all the cards?

We didn't highlight the `irrelevant` labeled record in the beginning. So no infobox means irrelevant. It could be confusing what the labels of other cards are without infobox. But it could also be a burden that makes it difficult to quickly see which records have relevant labels in the dataset if all records have the infobox. You need to read carefully to distinguish two words `relevant` and `irrelevant`.

I think it can be explained in the documentation. This is a feature of the exploration mode.

asreview/webapp/src/StyledComponents/StyledAlert.js
```diff
@@ -7,7 +7,7 @@ export function ExplorationModeRecordAlert(props) {
       severity="info"
       sx={{ borderBottomRightRadius: 0, borderBottomLeftRadius: 0 }}
     >
-      Labeled as relevant in an earlier study
+      Labeled as relevant in the dataset
```
J535D165, line 10:
```suggestion
      Labeled as RELEVANT in the dataset
```

What about this? Then makes it easier to distinguish from "Labeled as IRRELEVANT in the dataset".

terrymyc: I prefer your initial idea. See screenshots.

Issue #920, opened 2022-01-13
API labeled_stats returns one more labeled count after creating a project


terrymyc:
For example, I labeled two priors when creating a new project; when the model has been trained and the project is ready, `labeled_stats` returns `"n": 3`. It was `"n": 2` before training the model.

J535D165:
@PeterLombaers Interesting :) This can be a lead on why things are failing.

PeterLombaers:
I think this is related to the way the stats are gathered, not the actual number of labeled items. Probably it treats a pending record as already labeled. I can check and fix it later.

PR #948, opened 2022-01-27
Fix labeled_stats

PeterLombaers:
Fix the api labeled_stats call by dropping the pending records. This should fix https://github.com/asreview/asreview/issues/920
The problem is that there are records pending a decision in the results table. It might be good to have a clear philosophy about which methods of the state return the pending records and which don't, but that is something for another pull request I'd think.


J535D165:
Thanks for the fix. I agree that we should rethink the API in relation to the pending records and makes things internally consistent. I think we should avoid the use of `get_dataset` outside the state module. This might help. In this specific case, we can have something like `get_labeled` with an extra argument to return other columns as well. `get_labeled` should never return pending (by default?).

PR #949, opened 2022-01-28
Fix invalid date in Safari

terrymyc:
This PR fixed the date display on the project table in Safari.

### DIFF
--- a/asreview/webapp/src/HomeComponents/DashboardComponents/ProjectTable.js
+++ b/asreview/webapp/src/HomeComponents/DashboardComponents/ProjectTable.js
@@ -191,7 +191,7 @@ const ProjectTable = (props) => {
    * Format date and mode
    */
   const formatDate = (datetime) => {
-    let date = new Date(datetime);
+    let date = new Date(datetime.split(" ")[0]);
     let dateString = date.toDateString().slice(4);
     let dateDisplay =
       dateString.replace(/\s+\S*$/, ",") + dateString.match(/\s+\S*$/);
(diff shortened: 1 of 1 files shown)


asreview/webapp/src/HomeComponents/DashboardComponents/ProjectTable.js
```diff
@@ -191,7 +191,7 @@ const ProjectTable = (props) => {
    * Format date and mode
    */
   const formatDate = (datetime) => {
-    let date = new Date(datetime);
+    let date = new Date(datetime.split(" ")[0]);
```
J535D165, line 194: Using `created_at_unix` instead of `datetimeCreated` might be a better fix. Unix time is easier and more efficient for storing datetime. 

https://github.com/asreview/asreview/blob/7b85eb843e17260bc3ba866be5734f170326e1e4/asreview/webapp/src/HomeComponents/DashboardComponents/ProjectTable.js#L345

asreview/webapp/src/HomeComponents/DashboardComponents/ProjectTable.js
```diff
@@ -342,8 +342,8 @@ const ProjectTable = (props) => {
                           variant="subtitle1"
                           noWrap
                         >
-                          {row["datetimeCreated"]
-                            ? formatDate(row["datetimeCreated"])
+                          {row["created_at_unix"]
+                            ? formatDate(row["created_at_unix"])
```
J535D165, line 346: Nice. 

One day we should look at property names I think. Google's style guide is nice. https://google.github.io/styleguide/jsoncstyleguide.xml#Property_Name_Guidelines

J535D165:
Thanks!

PR #951, opened 2022-01-29
Add warning about duplicates when uploading dataset

DominiqueMaciejewski:
Added that no duplicates should be in the dataset either

### DIFF
--- a/asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
+++ b/asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
@@ -75,7 +75,8 @@ const AddDataset = (props) => {
               <code>.txt</code>) and tabular datasets (<code>.csv</code>,{" "}
               <code>.tab</code>, <code>.tsv</code>, <code>.xlsx</code>). The
               selected dataset should contain the title and abstract of each
-              record. It is highly recommended to add a dataset without duplicate records.{" "}
+              record. It is highly recommended to add a dataset without{" "}
+              duplicate records.{" "}
               <Link
                 underline="none"
                 href="https://asreview.readthedocs.io/en/latest/intro/datasets.html"
--- a/asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
+++ b/asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
@@ -75,7 +75,8 @@ const AddDataset = (props) => {
               <code>.txt</code>) and tabular datasets (<code>.csv</code>,{" "}
               <code>.tab</code>, <code>.tsv</code>, <code>.xlsx</code>). The
               selected dataset should contain the title and abstract of each
-              record. It is highly recommended to add a dataset without{" "}
+              record. To optimally benefit from the performance of the active
+              learning model, it is highly recommended to add a dataset without
               duplicate records.{" "}
               <Link
                 underline="none"
--- a/asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
+++ b/asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
@@ -75,7 +75,7 @@ const AddDataset = (props) => {
               <code>.txt</code>) and tabular datasets (<code>.csv</code>,{" "}
               <code>.tab</code>, <code>.tsv</code>, <code>.xlsx</code>). The
               selected dataset should contain the title and abstract of each
-              record.{" "}
+              record and be free of any duplicates.{" "}
               <Link
                 underline="none"
                 href="https://asreview.readthedocs.io/en/latest/intro/datasets.html"
--- a/asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
+++ b/asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
@@ -75,7 +75,7 @@ const AddDataset = (props) => {
               <code>.txt</code>) and tabular datasets (<code>.csv</code>,{" "}
               <code>.tab</code>, <code>.tsv</code>, <code>.xlsx</code>). The
               selected dataset should contain the title and abstract of each
-              record and be free of any duplicates.{" "}
+              record. It is highly recommended to add a dataset without duplicate records.{" "}
               <Link
                 underline="none"
                 href="https://asreview.readthedocs.io/en/latest/intro/datasets.html"


J535D165 (approved):

J535D165:
Thanks!

General comment (also for the other reviewers): I think it should be a recommendation. However, I'm fine to merge.

DominiqueMaciejewski:
Shit, sorry! Still figuring out GitHub!

terrymyc (changes requested):
How about "It is recommended that duplicate records are removed."

DominiqueMaciejewski:
I think that is better Terry! I just wanted t keep it as short as possible, but I like your wording better (although "it is recommended" might sound like it is not that big of a deal when they don't do delete them)

asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
```diff
@@ -75,7 +75,7 @@ const AddDataset = (props) => {
               <code>.txt</code>) and tabular datasets (<code>.csv</code>,{" "}
               <code>.tab</code>, <code>.tsv</code>, <code>.xlsx</code>). The
               selected dataset should contain the title and abstract of each
-              record.{" "}
+              record and be free of any duplicates.{" "}
```
J535D165, line 78: What about this?

```suggestion
              record. It is highly recommended to upload a dataset without duplicate records.{" "}
```

terrymyc: `...to add a dataset`
Be consistent with the wording on the page.

J535D165: Thanks

DominiqueMaciejewski: Good! The addition of highly definitely makes it more urgent!

asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
```diff
@@ -75,7 +75,7 @@ const AddDataset = (props) => {
               <code>.txt</code>) and tabular datasets (<code>.csv</code>,{" "}
               <code>.tab</code>, <code>.tsv</code>, <code>.xlsx</code>). The
               selected dataset should contain the title and abstract of each
-              record.{" "}
+              record. It is highly recommended to add a dataset without duplicate records.{" "}
```
J535D165, line 78:
```suggestion
              record. It is highly recommended to add a dataset without{" "}
              duplicate records.{" "}
```

Rensvandeschoot:
I propose to explain to the reader why this is essential. For example "to optimally benefit from the performance of the active learning models it is essential to upload high-quality data: provide only unique records (after de-duplication) and text in every record (after missing information imputation)"

terrymyc:
> I propose to explain to the reader why this is essential. For example "to optimally benefit from the performance of the active learning models it is essential to upload high-quality data: provide only unique records (after de-duplication) and text in every record (after missing information imputation)"

I would say this belongs to the documentation. When the user wants to learn more, they can be directed to the documentation.

DominiqueMaciejewski:
> > I propose to explain to the reader why this is essential. For example "to optimally benefit from the performance of the active learning models it is essential to upload high-quality data: provide only unique records (after de-duplication) and text in every record (after missing information imputation)"
> 
> I would say this belongs to the documentation. When the user wants to learn more, they can be directed to the documentation.

I agree with Terry. Otherwise the box might get too big. Here it is just about a simple reminder!

Rensvandeschoot:
Sorry, I disagree. Let me try to explain: 

Anyone working on systematic reviews using reference manager software can easily deal with duplicates by just ignoring the duplicate. A human does not face any consequence in levels of performance, and in reference manager software you can easily delete a dup[licate record. The same goes for missing information, a human can easily search for the missing information outside the reference manager software without consequences. It takes quite some effort to remove duplicates and add missing info, and we humans are typically lazy, so why improve the quality of the data if it does not matter that much?

When using active learning-based software, the performance is sensitive for duplicates and missing information. A typical user does not know this. As a consequence, we have received several issues and even more emails from users complaining about lower-than-expected performance. After close inspection of their data, it appeared the low performance was mostly due to duplicates and/or missing information. Remember the famous acronym GIGO often used in statistics? The same holds for Active Learning-based software: Garbage in, Garbage out (I actually like QIQO better: Quality In, Quality Out). 

Therefore, I believe we should explain better to our users **why** they need to spend more time improving the quality of the data, not only what they should do (because most humans, including myself, are lazy and won't even bother reading the warning if they don't directly benefit from it).

asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/AddDataset.js
```diff
@@ -75,7 +75,8 @@ const AddDataset = (props) => {
               <code>.txt</code>) and tabular datasets (<code>.csv</code>,{" "}
               <code>.tab</code>, <code>.tsv</code>, <code>.xlsx</code>). The
               selected dataset should contain the title and abstract of each
-              record.{" "}
+              record. It is highly recommended to add a dataset without{" "}
+              duplicate records.{" "}
```
terrymyc, line 79:
```suggestion
              record. To optimally benefit from the performance of the active
              learning model, it is highly recommended to add a dataset 
              without duplicate records.{" "}
```

Rensvandeschoot: I propose to replace:

"The selected dataset should contain the title and abstract of each record. It is highly recommended to add a dataset without duplicate records."

with:

"To optimally benefit from the performance of the active learning model it is essential to upload high-quality data: provide unique records (after de-duplication) and avoid missing information."

only 4 more words

terrymyc: Before we can think about the text, we need to come to a consensus on what is necessary. 

> @J535D165 General comment (also for the other reviewers): I think it should be a recommendation.

It is essential to add a dataset that contains:
- [x] title
- [ ] abstract
- [ ] no duplicate records

Rensvandeschoot: Duplicate records can be intentional (ping @bmkramer) so we should not use the phrase "no duplicate records".

terrymyc: So adding a dataset without duplicate records is highly recommended but not essential.

terrymyc: > upload high-quality data: provide unique records (after de-duplication) and avoid missing information

This is not concise according to the writing principle. https://www.material.io/design/communication/writing.html#principles

"High-quality data" is a lengthy description. "Missing information" is an implication.
“Unique records" is also an implication so it has "after de-duplication" attached to it. Then it becomes lengthy.

I still suggest that half a sentence explaining why it is highly recommended to add a dataset without duplicate records is sufficient.

terrymyc (approved):

Rensvandeschoot (approved):

DominiqueMaciejewski:
Yeah that sounds good. Clear and concise and i get Rens comment. I think if i had seen that, i would've actually gone back to my refman

J535D165:
Thanks @DominiqueMaciejewski for your first contribution!

PR #956, opened 2022-02-02
Adjust project review page when no unlabeled record found

terrymyc:
This PR implemented the project review page when all the records have been reviewed.

image


J535D165 (approved):

J535D165:
Superb @terrymyc. 

Are you about to use the same screen for the "Mark project as finished" function? So, you mark a project as finished; you can't review anymore obviously. So, what do you see when you navigate to the review page? I would suggest having something similar. However, the label "You have reviewed all the records" wouldn't fit there. Maybe we need an extra label in the project file like "allRecordsLabeled"?

asreview/webapp/src/ProjectComponents/ReviewComponents/ReviewPageFinished.js
```diff
@@ -41,26 +34,69 @@ const StyledBox = styled(Box)(({ theme }) => ({
...
+              <Typography className={classes.textTitle} variant="h5">
+                Congratulations! You have finished this project.
+              </Typography>
+              <Typography className={classes.text}>
+                You have stopped reviewing and marked this project as finished.
+                If you want to resume the review, please{" "}
+                <Link>update project status</Link>.
+              </Typography>
+            </Stack>
+          )}
```
terrymyc, line 87: @J535D165 When the project is marked as finished but still has unlabeled records, you will see something different (the link to update project status not working yet).

terrymyc: Since marking a project as finished and finishing reviewing all records share the same property `reviewFinished`, when this property is true, the project review finished page needs to call `get_document` to decide which text should be shown.

J535D165: Ah sorry, I overlooked that. Nice solution.

terrymyc: Suppose we add an additional property `allRecordsLabeled`, this would bring two things:
- Marking a project as finished only changes the value of `reviewFinished`.
- After reviewing all the records, the values of `allRecordsLabeled` and `reviewFinished` change.

After reviewing all the records, the project should not be marked as in review. This may call `get_document` again somewhere. Perhaps adding a property `allRecordsLabeled` to the project info looks to be a better solution?

PR #957, opened 2022-02-03
Implement basic function to count duplicates

J535D165:


J535D165:
This proposal was introduced to facilitate the discussion in #951. This function can be used to return statistics after the dataset was uploaded. This can help to develop additional checks in the front end.

PR #965, opened 2022-02-04
add option to re-import RIS files into reference software to documentation

Rensvandeschoot:
This PR adds the option to re-import RIS files into reference software to the documentation.

PR #991, opened 2022-02-23
Add information on labeling in RIS format; upload screenshot of Endnote

ottomattas:
For https://github.com/asreview/asreview/pull/965


docs/source/intro/datasets.rst
```diff
@@ -36,6 +36,46 @@ formats:
    IEEE Xplore, Scopus and ScienceDirect. Citation managers Mendeley, RefWorks,
    Zotero, and EndNote support the RIS file format as well.
 
+For parsing RIS file format, the software uses a Python RIS files parser and reader (`rispy
+    <https://pypi.org/project/rispy/>`_). Successful import/export depends on a proper
+    data set structure. To validate your data set, the complete default mapping can be found
+    on the developer's `GitHub page <https://github.com/MrTango/rispy#complete-default-mapping>`_.
+
+After a labeling decision is made and the record is exported from ASReview Lab, the decision
```
J535D165, line 44:
```suggestion
After a labeling decision is made and the record is exported from ASReview LAB, the decision
```

docs/source/intro/datasets.rst
```diff
@@ -36,6 +36,46 @@ formats:
...
+    | AU          | Author                                                                                                  | no\*\*    |
+    +-------------+---------------------------------------------------------------------------------------------------------+-----------+
+    | DO          | DOI                                                                                                     | no        |
+    +-------------+---------------------------------------------------------------------------------------------------------+-----------+
+    | N1          | Notes                                                                                                   | no\*\*    |
+    +-------------+---------------------------------------------------------------------------------------------------------+-----------+
+    | ER          | End of Reference (must be empty and the last tag)                                                       | yes       |
+    +-------------+---------------------------------------------------------------------------------------------------------+-----------+
```
J535D165, line 68: Very technical, I'm not sure what the benefits are for an end-user. I would prefer to move this table to a technical section and explain to the user what to expect when importing/exporting. The user will use a ref man in 99% of the cases.

docs/source/intro/datasets.rst
```diff
@@ -36,6 +36,46 @@ formats:
    IEEE Xplore, Scopus and ScienceDirect. Citation managers Mendeley, RefWorks,
    Zotero, and EndNote support the RIS file format as well.
 
+For parsing RIS file format, the software uses a Python RIS files parser and reader (`rispy
+    <https://pypi.org/project/rispy/>`_). Successful import/export depends on a proper
+    data set structure. To validate your data set, the complete default mapping can be found
+    on the developer's `GitHub page <https://github.com/MrTango/rispy#complete-default-mapping>`_.
```
J535D165, line 42: Please resolve warnings

J535D165 (changes requested):

J535D165:
Thanks, would be nice to have some discussion on the table. Too technical for this section I think.

docs/source/intro/datasets.rst
ottomattas: I just followed the documentation page and what was on it previously to match the style.
Maybe we can move this table to a technical section when we overhaul the documentation for v1?

ottomattas: (There is also a similar table for tabular data sets on the same page.)

J535D165: > (There is also a similar table for tabular data sets on the same page.)

Yes, but that's something that's not standardized. I don't expect any of our users to open de RIS file outside the ref man. So they will never see something like `TY` and `ER`.

docs/source/intro/datasets.rst
ottomattas: The documentation looks to be riddled with these warnings. This might be due to having an outdated version.

docs/source/intro/datasets.rst
ottomattas: I have removed the table. It can be easily recreated if necessary at a later stage.

docs/source/intro/datasets.rst
ottomattas: This section was removed together with the technical table.

ottomattas:
The requested changes was committed.

PR #1002, opened 2022-03-04
Remove hover info for modes on project init

J535D165:
I'm struggling a lot with this hover info box. The box overlaps with the project name and doesn't hide itself. I think we don't need it and therefor I propose to remove it.


terrymyc (commented):
Thanks. I know this issue and haven't worked on it yet. Currently you need to hover over the popover again after selecting the mode. I think the popover should be at least kept on the project details page (when `isProjectSetup()` is false). Either merge this first or leave it for now. I will anyway come back to improve this later.

J535D165:
I leave this up to you!

PR #1004, opened 2022-03-10
Adjust interface for the simulation project

terrymyc:
This PR adjusted the user interface for creating a simulation project and the corresponding project page.

localhost_3000_projects

image

image


asreview/entry_points/simulate.py
```diff
@@ -89,6 +89,7 @@ def _mark_review_finished(project_path, review_id=None):
         review_index = [x['id']
                         for x in project_config['reviews']].index(review_id)
 
+    project_config['reviewFinished'] = True
```
J535D165, line 92: Oke. Let's replace in future PR.

asreview/webapp/src/ProjectComponents/ProjectPage.js
```diff
@@ -85,13 +104,62 @@ const ProjectPage = (props) => {
     }
   );
 
+  const isSimulating = () => {
+    return (
+      data?.mode === projectModes.SIMULATION &&
+      data?.projectInitReady &&
+      !data?.reviewFinished
```
J535D165, line 111: Oke, this is the upcoming status. 

I think we might need a discussion about whether mode and status should be in the Redux store. This makes this a lot easier.

asreview/webapp/src/ProjectComponents/ProjectPage.js
```diff
@@ -85,13 +104,62 @@ const ProjectPage = (props) => {
     }
   );
 
+  const isSimulating = () => {
```
J535D165, line 107: Where is `data` defined? I don't fully understand how this works.

terrymyc: It is the data fetched from the `info` API above. Basically, it uses `mode`, `projectInitReady`, and `reviewFinished` in the returned project info to judge if the simulation is finished.

asreview/webapp/src/ProjectComponents/ProjectPage.js
terrymyc: Perhaps when we have the `status` (`setup`, `running`, and `stopped`), this would be already resolved. Namely, the returned status `running` can replace `isSimulating` defined here.

asreview/webapp/src/ProjectComponents/ProjectPage.js
```diff
@@ -70,6 +91,13 @@ const ProjectPage = (props) => {
         } else if (!data["projectNeedsUpgrade"]) {
           // open project page
           console.log("Opening project " + project_id);
+          // if simulation is running
+          if (
+            data["mode"] === projectModes.SIMULATION &&
+            !data["reviewFinished"]
+          ) {
+            setIsSimulating(true);
+          }
```
terrymyc, line 100: @J535D165 Changed. I add a state `isSimulating` and set it here.

J535D165:
Very nice work @terrymyc I really appreciate your sense for detail.

PR #1012, opened 2022-03-24
Clean state code

PeterLombaers:
This PR clean up the state code and uniformizes the return of the state get functions. Get functions now allow for dropping the priors and pending records from the return. By default, priors are included and pending records are excluded. Also, the docstrings of the sqlstate and basestate have been updated.


asreview/state/sqlstate.py
```diff
@@ -953,99 +912,262 @@ def get_priors(self):
...
+
+        query = """SELECT record_table.record_id, results.label,
+                results.rowid AS label_order, results.query_strategy,
+                last_ranking.ranking
+                FROM record_table
+                LEFT JOIN results
+                ON results.record_id=record_table.record_id
+                LEFT JOIN last_ranking
+                ON record_table.record_id=last_ranking.record_id
+                """
```
J535D165, line 1156: `ORDER BY label_order, ranking` would solve the ordering below, isn't it?

That simplifies 
```
        labeled = df.loc[~df['label'].isna()] \
            .sort_values('label_order') \
            .loc[:, ['record_id', 'label']] \
            .astype(int)
```

```
        labeled = df.loc[~df['label'].isna(), ['record_id', 'label']].astype(int)
```

tests/test_state.py
```diff
@@ -175,6 +175,37 @@ def test_get_dataset():
...
 
+def test_get_dataset_drop_prior():
+    with open_state(TEST_STATE_FP) as state:
+        assert len(state.get_dataset(priors=False)) == \
+               len(TEST_RECORD_IDS) - TEST_N_PRIORS
+        assert (state.get_dataset(priors=False)['query_strategy'] !=
+                'prior').all()
+        assert 'query_strategy' in state.get_dataset(priors=False).columns
+        assert 'query_strategy' not in state.get_dataset('label',
+                                                         priors=False)
```
J535D165, line 186: Thanks!

asreview/state/sqlstate.py
```diff
@@ -927,21 +851,56 @@ def get_dataset(self, columns=None):
         """
         if type(columns) == str:
             columns = [columns]
+
+        # Make sure the right columns are available to drop priors/pending.
```
J535D165, line 855: Why would those columns not be available?

J535D165: Ah, I start to understand (see also comment below). I would prefer to do the slicing with SQL. That resolves most of the challenges here.

J535D165 (commented):
Thanks a lot. Thanks for the tests as well. 

Two general remarks:
- please see PEP 257 for docstring. 
- I think we can improve performance and maintainability in `get_document` by using SQL only.

asreview/state/sqlstate.py
```diff
@@ -927,21 +851,56 @@ def get_dataset(self, columns=None):
...
+            contains_query_strategy = True
+            contains_labels = True
+        else:
+            contains_query_strategy = 'query_strategy' in columns
+            contains_labels = 'label' in columns
+
+            if not priors and not contains_query_strategy:
+                columns.append('query_strategy')
+            if not pending and not contains_labels:
+                columns.append('label')
```
J535D165, line 866:
```suggestion
        sql_where = []
        if priors or pending:
            if priors:
                 sql_where.append("query_strategy is not 'prior'")
            if pending:
                 sql_where.append("label is not NULL")
        
        print("WHERE " + " AND ".join(sql_where)
```

J535D165: Something like this will do (not tested)

PeterLombaers:
I agree with your comments and incorporated them. As far as I'm concerned, this is ready to merge.

J535D165:
Great work @PeterLombaers. Super happy you found the time to work on this.

PR #953, opened 2022-01-31
Validate project info before updating info

J535D165:
This PR proposes to validate the schema of the project file before updating the file. This should avoid potential issues for users and confusion for developers about the (undocumented) schema. @terrymyc what do you think?

I got inspired by the discussions in #928 and #952.

PR #1011, opened 2022-03-24
Refactor status handling in ASReview backend

J535D165:

PR #1014, opened 2022-03-26
Add jsonschema validator for project file

J535D165:
Replaces #953


J535D165:
This PR waits for #1011. The front-end test will probably work then.

PR #1023, opened 2022-03-30
Fix converter for projects with tmp folder (exported projects)

J535D165:
Also remove the tmp content from the new project file.

### DIFF
--- a/asreview/state/sql_converter.py
+++ b/asreview/state/sql_converter.py
@@ -92,7 +92,7 @@ def upgrade_asreview_project_file(fp, from_version=0, to_version=1):
 
     if from_version != 0 and to_version != 1:
         raise ValueError(
-            f"Not possible to upgrade from {from_version} to {to_version}.")
+            f"Not possible to upgrade from v{from_version} to v{to_version}.")
 
     # Check if it is indeed an old format project.
     if not is_old_project(fp):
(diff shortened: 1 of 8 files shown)


asreview/state/sql_converter.py
```diff
@@ -178,14 +178,23 @@ def move_old_files_to_legacy_folder(fp):
...
-        if file_path.name not in files_to_keep:
-            file_path.unlink()
+
+    project_content = list(fp.iterdir())
+
+    # copy to legacy folder
+    shutil.copytree(fp, Path(fp, 'legacy'))
+
+    # remove files and folders
+    files_to_keep = ['project.json', 'data', 'lock.sqlite']
```
J535D165, line 188: @PeterLombaers check `__MACOSX` bug ;)

asreview/webapp/api.py
```diff
@@ -212,16 +212,8 @@ def api_init_project():  # noqa: F401
 def api_upgrade_project_if_old(project):
     """Get upgrade project if it is v0.x"""
 
-    try:
-        upgrade_asreview_project_file(project.project_path)
-
-    except ValueError:
```
J535D165, line 218: ValueErrors can occur outside of the range of intended failures.

PR #971, opened 2022-02-08
Documentation for ASReview LAB version v1.0

J535D165:
Main branch for the documentation of ASReview LAB version 1. 

Part of #688.

Issue #738, opened 2021-09-13
Installing ASReview backend fails on Apple M1 chip


ottomattas:
**Describe the bug**
Trying to install ASReview in [DEVELOPMENT mode](https://github.com/asreview/asreview/blob/master/DEVELOPMENT.md) fails on an Apple M1 chip. Process takes about 1 hour to complete.

**To Reproduce**
Steps to reproduce the problem:
1. Pull the latest code from the repository
2. Navigate to the repository locally
3. `pip install -e .`

**Version information**
 - OS: macOS Big Sur Version 11.5.2 (20G95)
 - MacBook Pro (13-inch, M1, 2020)

**Additional context**
Full [CLI output](https://github.com/asreview/asreview/files/7153812/cli_output.txt.zip) attached as a zip container.

ottomattas:
From the looks of it, there are three packages that need a workaround - `h5py`, `scikit-learn`, `scipy`.

Relevant section from the end of CLI output:
```
Successfully built gevent greenlet pandas zope.interface
Failed to build h5py scikit-learn scipy
Installing collected packages: scipy, numpy, Jinja2, itsdangerous, click, zope.interface, zope.event, scikit-learn, pytz, python-dateutil, greenlet, flask, et-xmlfile, xlrd, sklearn, rispy, pandas, openpyxl, h5py, gevent, flask-cors, dill, asreview
```

ottomattas:
There is a quick workaround, using [miniforge](https://formulae.brew.sh/cask/miniforge).

```
# Install miniforge
% brew install miniforge

# Create environment
% conda create --name asreview_m1

# Initialise (zsh) shell
% conda init zsh

# Source (zsh) shell directly
% . ~/.zshrc

# Activate environment
% conda activate asreview_m1

# Install packages which fail to install via pip
% conda install h5py scipy scikit-learn

# Install asreview
% pip install -e .

# Delete the environment, in case you need to clean up afterwards
% conda deactivate
% conda env remove -n asreview_m1
```

This workaround was inspired by [Hendrik Erz](https://github.com/nathanlesage), a researcher and an Open Source contributor. If you want to read up on the method proposed here, you can see his [blog post](https://hendrik-erz.de/post/update-setting-up-python-numpy-and-pytorch-natively-on-apple-m1) on the issue.

MarielleZZ:
Starting with a git clone + cd on asreview and proceeding with the workaround above was successful for me:

`git clone https://github.com/asreview/asreview.git
[...]

PR #1035, opened 2022-04-12
Update install section of documentation

terrymyc:
This PR updates the install section of the documentation. The [installation guide on the website](https://asreview.nl/download) should link to the documentation showing Docker and server installation options.

Questions:
- Is the issue with M1-chipped Apple computers still there?
- Linux is supported, but no installation guide for Linux users on the website.

Todo:
- [x] server installation docs
- [x] style troubleshooting section

image

image

image

image

image


J535D165:
> * Is the issue with M1-chipped Apple computers still there?

Yes. Not 100% sure, but I think there are still some issues. 

> * Linux is supported, but no installation guide for Linux users on the website.

https://asreview.nl/download/ there is!

docs/source/install/docker.rst
```diff
@@ -0,0 +1,40 @@
...
+<https://hub.docker.com/r/asreview/asreview>`__ Docker Hub repository.
+
+Open `Terminal
+<https://support.apple.com/guide/terminal/open-or-quit-terminal-apd5265185d-f365-44cb-8b09-71a064a42125/mac>`__
+(macOS) or `Command Prompt
+<https://www.lifewire.com/how-to-open-command-prompt-2618089>`__ (Windows) and enter:
+
+.. code-block:: bash
+
+    docker pull asreview/asreview  # latest stable release
```
J535D165, line 25: No need to pull, isn't it? You can skip this. See https://github.com/asreview/asreview/tree/master/docker#quickstart-using-docker

docs/source/install/install.rst
```diff
@@ -0,0 +1,29 @@
+Install ASReview LAB
+====================
+
+ASReview LAB is supported on Windows and macOS.
```
J535D165, line 4: Linus is also supported. Download stats show quite some users!

`macOS` -> `MacOSX`

terrymyc: Isn't `Mac OS X` an outdated name? https://en.wikipedia.org/wiki/MacOS

docs/source/install/docker.rst
terrymyc: Indeed, if the image is not installed, `docker run` will pull anyway. My question is: maybe users should be aware of this download step? The downloaded image is quite a large file.

terrymyc:
@J535D165 The installation guide for Linux is the same as that for macOS on the website, but Miniconda has Linux installers. Should we separate them?

J535D165:
Most Linux distros have Python installed. And the users are probably familiar with installing Python packages. I dont think we need any action for the website atm.

docs/source/install/docker.rst
J535D165: Docker users will probably know. They can read about Docker when installing Docker or from the various resources online.

terrymyc:
@J535D165 The issue #738 and the workaround are based on installing ASReview LAB in DEVELOPMENT mode. Is it an issue with doing `pip install asreview`?

J535D165:
Oef, yes this is super tricky now development is merged. This requires a hotfix to the docs and website indeed.

terrymyc:
> Oef, yes this is super tricky now development is merged. This requires a hotfix to the docs and website indeed.

I mean the issue states that the installation from the local git repo by `pip install -e .` will encounter the issue. A [recent comment](https://github.com/asreview/asreview/issues/738#issuecomment-1072085716) mentioned `pip install asreview` worked.

J535D165:
> I mean the issue states that the installation from the local git repo by pip install -e . will encounter the issue.

Yes I know. That's not what we want. 

I just tried to reproduce it on my M1, and still having the issue with sklearn (the main problem).

terrymyc:
@J535D165 Another question: we instruct users to check the Python version by `python --version`, but Linux and macOS ship with Python 2 (if I am still right). In this case, `python3 --version` should be used to check the version. Also, `pip3` is used. For example, see [TensorFlow](https://www.tensorflow.org/install/pip#1.-install-the-python-development-environment-on-your-system).

terrymyc:
@J535D165 Is [server installation](https://asreview.readthedocs.io/en/latest/intro/installation.html#server-installation) a part of `Install with pip`?

J535D165:
I dont think so. True, it is usually installed with pip, but you dont want to bother 99% of the users with ports and hosts.

terrymyc:
> I dont think so. True, it is usually installed with pip, but you dont want to bother 99% of the users with ports and hosts.

Makes sense. It can also be installed with Docker on a server?

Rensvandeschoot:
> The [installation guide on the website](https://asreview.nl/download) should link to the documentation showing Docker and server installation options.

Thank you for the reminder! After the release, I will update the website so that the hyperlinks do not get messed up.

J535D165:
Most of your changes are incorporated in #971 now. Thanks for all the valuable suggestions. If there is anything missing, please let me know.

PR #1043, opened 2022-04-21
Display wordmark on boot up

terrymyc:
This PR makes it possible to display the ASReview LAB wordmark when starting the software.

image

image

When there is an error when starting:
image


J535D165 (approved):
Nice. Merge.

PR #1046, opened 2022-04-25
Add authentication in the front end

terrymyc:
This PR added authentication in the front end. The protected routes are accessible after sign-in. This function relies on back-end support.

image

image

image


J535D165:
Very promising

PR #1052, opened 2022-05-02
Display project check box when clicking Export of project created in <1.0 version

terrymyc:
This PR displayed the project check dialog when the Export button of the project created in the <1.0 version is clicked, rather than directly exporting the `.asreview` file, to be consistent with clicking other quick buttons and to avoid unexpected exporting behavior. However, it is still not possible to export the dataset before upgrading the project.


J535D165:
Nice

PR #1057, opened 2022-05-02
Add button to delete prior knowledge when creating new project

terrymyc:
This PR added a button to delete the added prior knowledge in the setup dialog.

image

### DIFF
--- a/asreview/webapp/src/ProjectComponents/HistoryComponents/LabeledRecordCard.js
+++ b/asreview/webapp/src/ProjectComponents/HistoryComponents/LabeledRecordCard.js
@@ -263,7 +263,11 @@ const LabeledRecordCard = (props) => {
                   </span>
                 </Tooltip>
                 {props.is_prior && (
-                  <Tooltip title="Remove">
+                  <Tooltip
+                    title={`Remove ${
+                      value.included !== 1 ? "irrelevant" : "relevant"
+                    } label`}
+                  >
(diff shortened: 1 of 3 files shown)


asreview/webapp/src/ProjectComponents/HistoryComponents/LabeledRecordCard.js
```diff
@@ -183,121 +195,139 @@ const LabeledRecordCard = (props) => {
...
-              mobileScreen={props.mobileScreen}
-              mutate={mutate}
-              note={note}
-              setNote={setNote}
-              is_prior={props.is_prior}
-            />
-          </Card>
-        ))}
+                {props.is_prior && (
+                  <Tooltip title="Remove">
```
J535D165, line 266:
```suggestion
                  <Tooltip title="Remove from prior knowledge">
```

What about this. Might be confusing to users (delete completely).

J535D165 (commented):
Nice!

asreview/webapp/src/ProjectComponents/HistoryComponents/LabeledRecordCard.js
Rensvandeschoot: yes! we don't want to give the impression the record is deleted. Maybe add **label**?
 ```
Remove label from prior knowledge
```

PR #1059, opened 2022-05-02
Resolve failing github action on excel export

J535D165:
Probably due to race confict. This is the first attempt to reveal the actual bug.


J535D165:
It turns hard to reproduce the error after these minor changes. I don't think this solves the actual problem, but it should raise more informative errors when failing. I vote for merging this and if the error returns, we can study the problem in more detail given the improved errors.

PR #1065, opened 2022-05-05
Add informative error when feature extractor is poorly designed

J535D165:
Check number of records after feature extraction. This should be equal to the number of records in the dataset. This improved error might help developers to debug errors with the feature extraction.

### DIFF
--- a/asreview/review/base.py
+++ b/asreview/review/base.py
@@ -110,6 +110,15 @@ class BaseReview(ABC):
                     as_data.bodies,
                     as_data.keywords
                 )
+
+                # check if the number of records after the transform equals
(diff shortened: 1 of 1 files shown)


asreview/review/base.py
```diff
@@ -110,6 +110,15 @@ def __init__(
                     as_data.bodies,
                     as_data.keywords
                 )
+
+                # check if the number of records after the transform equals
+                # the number of in the dataset
```
J535D165, line 115:
```suggestion
                # the number of records in the dataset
```

PeterLombaers (approved):
Looks good!

Issue #1064, opened 2022-05-05
OverflowError: Python int too large to convert to C long


Rensvandeschoot:
### Discussed in https://github.com/asreview/asreview/discussions/1063

<div type='discussions-op-text'>

<sup>Originally posted by **juliandwain** May  5, 2022</sup>
Hi,
thanks for this amazing library!

**Describe the bug**
I created a dataset by myself from a Scopus search and transformed it so that it can be interpreted by ASReview.
The dataset can be uploaded without any issues.
I know this is not an issue of ASReview itself, I assume that the issue is within the dataset, but I am not able to figure out why the overflow error appears.

**To Reproduce**
Upload the dataset and follow the instructions. The dataset is added as attachment.

**Screenshots**
![image](https://user-images.githubusercontent.com/38472982/166891141-286780b4-4cbe-421d-9d63-820b99485c75.png)
![image](https://user-images.githubusercontent.com/38472982/166890566-66385a8e-cb41-406d-a6cb-4ea3fd8c632f.png)

**Version information**
 - OS: Windows
 - ASReview version 0.19.1

**Additional context**
[220505_Scopus_data.csv](https://github.com/asreview/asreview/files/8629829/220505_Scopus_data.csv)

</div>

J535D165:
@juliandwain I can't reproduce this issue in version 0.19 on macOS. 

The issue might be an integer overflow on Windows systems caused by the `record_id` column. You can rename or delete this column such that ASReview ignores it (record_ids are used by ASReview internally, but are generated on the fly if not present). 

This dataset is also "fully labeled". This means that there are labels identified by ASReview. This is the result of the `label` column. If this is intentional, there is nothing to screen. If this wasn't intentional, then renaming this column is also advised.

PeterLombaers:
Apparently this does not happen on mac because the python basic int type is 64 bit long: https://stackoverflow.com/questions/38314118/overflowerror-python-int-too-large-to-convert-to-c-long-on-windows-but-not-ma
Renaming the 'record_id' column should fix it though, as should (#1066).

juliandwain:
I am able to resolve the issue when changing the 'record_id' column from the EID of the articles to a simple index numbering (i.e., 0,1,2,....).
Nevertheless, I do not understand why there is an overflow since the EID number is not an integer that cannot be handled by the OS.

Anyway, thank you all for your help! I really appreciate it!

J535D165:
Closed via #1066. 

[...]

PR #1066, opened 2022-05-05
Fix int overflow bug

PeterLombaers:
This fixes the int overflow bug (#1064) that occurs on Windows systems if the dataset has a column record_id and the entries are integers that are too large for the Python/Windows 32 bit int. Apparently this is a thing that does not occur on mac: https://stackoverflow.com/questions/38314118/overflowerror-python-int-too-large-to-convert-to-c-long-on-windows-but-not-ma

### DIFF
--- a/asreview/webapp/run_model.py
+++ b/asreview/webapp/run_model.py
@@ -126,8 +126,9 @@ def train_model(project_id, label_method=None):
                 "Project {project_id} - No new labels since last run.")
             return
 
-        query_record_ids = np.array([x[0] for x in diff_history], dtype=int)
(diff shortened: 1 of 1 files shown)


J535D165:
Very nice to have the reference as well. Thanks for doing the PR against the 0.19.x branch. Do you think we have to port this to master as well? Record_ids by column are removed there.

PeterLombaers:
It seemed to work fine when I tried that dataset in version 1 and the references to record_ids are all removed there.

J535D165:
Check, I see. This part of the code is removed.

PR #1074, opened 2022-05-10
Fix infinite query of simulation status

terrymyc:
This PR fixed the infinite query of the simulation status after finishing setting up a simulation project.


J535D165:
Great!

PR #1077, opened 2022-05-10
Fix Flask route

terrymyc:
This PR fixed the back-end route to redirect all URL to the index template.


JTeijema:
Thanks! Great job 👍🏻

PR #1078, opened 2022-05-10
Rename first step of setup into project info

J535D165:


terrymyc (changes requested):
Thanks! One little thing: only the first letter is capitalized. `Project information`.

J535D165:
Thanks for the feedback.

PR #1082, opened 2022-05-10
Remove function to convert idx to record_id

J535D165:


J535D165:
Although not used, it might be useful in the future. We can revert commit [755f421](https://github.com/asreview/asreview/commit/755f421d0d175566f7f0a6a3251aed54093ec9a9) if needed.

PR #1083, opened 2022-05-10
Remove references to abstract included

J535D165:


J535D165:
The end of an era @Rensvandeschoot

PR #1084, opened 2022-05-10
Display relevant/irrelevant random prior candidates

terrymyc:
This PR aims to display both relevant and irrelevant random records when adding prior knowledge in Exploration and Simulation modes. Improvements are still needed on the back end to improve efficiency.

image

### DIFF
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -730,8 +730,10 @@ def api_random_prior_papers(project):  # noqa: F401
 
     if subset:
 
-        rel_indices = as_data.df[as_data.df["debug_label"] == 1].index.values
-        irrel_indices = as_data.df[as_data.df["debug_label"] == 0].index.values
+        rel_indices = as_data.df[
+            as_data.df["debug_label"] == 1].index.values
+        irrel_indices = as_data.df[
+            as_data.df["debug_label"] == 0].index.values
 
         rel_indices_pool = np.intersect1d(pool, rel_indices)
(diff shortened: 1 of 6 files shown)


asreview/webapp/api.py
```diff
@@ -713,35 +713,103 @@ def api_random_prior_papers(project):  # noqa: F401
     the already labeled items.
     """
 
+    subset = request.args.get("subset", default=False, type=bool)
```
J535D165, line 716: Not clear why this is of type `bool`. I would expect the subset to be similar to history subsets for example.

asreview/webapp/api.py
```diff
@@ -713,35 +713,103 @@ def api_random_prior_papers(project):  # noqa: F401
...
 
-    try:
-        pool_random = np.random.choice(pool, 1, replace=False)[0]
-    except Exception:
-        raise ValueError("Not enough random indices to sample from.")
+    payload = {"result": []}
 
-    try:
-        record = read_data(project.project_path).record(pool_random)
+    if not subset:
```
J535D165, line 723: Always better to have `if subset` over if not.

asreview/webapp/api.py
terrymyc: This is an indicator of project mode. In the simulation and exploration modes, the returned random records need to be extracted from relevant and irrelevant subsets.

J535D165 (approved):

J535D165:
Nice step forward!

PR #1085, opened 2022-05-11
Update API example

JTeijema:
Hi, this is a new API example using the v1.0 code. It's a work in progress, so please adjust where needed.

The following things were not clear to me making this code:
- Whats the difference between update_review and mark_review_finished? Why not have a function that does both?
- Why can project_path not be assigned as a str, but only as a path
- Why is ASReviewProject not in the __init__ file? I think you always need it when running an API simulation
- Shouldn't update_review assign an error status automatically on a failed review?

I am dumping these here, as these might have to be adjusted in the API example (or in the code itself)

### DIFF
--- a/docs/source/guides/api.rst
+++ b/docs/source/guides/api.rst
@@ -13,31 +13,48 @@ This example shows how to use the API in simulation mode:
 
 .. code-block:: python
 
-	import asreview
-	from asreview.models import LSTMBaseModel
-	from asreview.query_strategies import MaxQueryModel
-	from asreview.balance_strategies import SimpleBalanceModel
-	from asreview.feature_extraction import EmbeddingLSTM
+	from pathlib import Path
 
-	# Load data
-	as_data = asreview.ASReviewData.from_file(DATA_FILE)
+	from asreview import ASReviewData
+	from asreview.review import ReviewSimulate
+	from asreview.project import ASReviewProject
 
-	train_model = LSTMBaseModel()
-	query_model = MaxQueryModel()
-	balance_model = SimpleBalanceModel()
-	feature_model = EmbeddingLSTM()
+	from asreview.models.classifiers import NaiveBayesClassifier
+	from asreview.models.query import MaxQuery
+	from asreview.models.balance import SimpleBalance
+	from asreview.models.feature_extraction import Tfidf
 
-	# Load the embedding matrix, only necessary for LSTM models
-	train_model.embedding_matrix = feature_model.get_embedding_matrix(
-		as_data.texts, EMBEDDING_FILE)
+	# Create a project object and folder
+	project = ASReviewProject.create(
+		project_path=Path(PROJECT_PATH),
+		project_id=None,
+		project_mode="simulate",
+		project_name=None,
+		project_description=None,
+		project_authors=None
+	)
+
+	# Initialize the simulation reviewer
+	reviewer = ReviewSimulate(
+		as_data=            ASReviewData.from_file(project.config['dataset_path']),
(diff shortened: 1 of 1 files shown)


JTeijema:
image

J535D165:
> Whats the difference between update_review and mark_review_finished? Why not have a function that does both?

The latter is a shortcut to the first one. So the first one does both. For simulate, you probably need only `mark_review_finished` and therefore the shortcut is available. 

> Why can project_path not be assigned as a str, but only as a path

Resolved

> Why is ASReviewProject not in the init file? I think you always need it when running an API simulation

Yes can be nice. Added it. 

> Shouldn't update_review assign an error status automatically on a failed review?

Current API design doesn't support this. When the simulation fails, there should not be a project file (not sure if this is correct for the CLI atm). To me, this feature is only useful in the frontend. But let me know if you think otherwise.

docs/source/guides/api.rst
```diff
@@ -9,35 +9,64 @@ strategy and use it with ASReview.
...
-	from asreview.query_strategies import MaxQueryModel
-	from asreview.balance_strategies import SimpleBalanceModel
-	from asreview.feature_extraction import EmbeddingLSTM
+	from pathlib import Path
 
-	# Load data
-	as_data = asreview.ASReviewData.from_file(DATA_FILE)
+	from asreview import ASReviewData
+	from asreview.review import ReviewSimulate
+	from asreview.project import ASReviewProject
```
J535D165, line 23:
```suggestion
	from asreview import ASReviewData, ASReviewProject
	from asreview.review import ReviewSimulate
```

docs/source/guides/api.rst
```diff
@@ -9,35 +9,64 @@ strategy and use it with ASReview.
...
+	project.update_review(status="review")
+	try:
+		reviewer.review()
+	except Exception as err:
+		project.update_review(status="error")
+		raise err
+
+	# Finish and export the project
+	project.export(EXPORT_PATH)
+	project.mark_review_finished()
```
J535D165, line 72:
```suggestion
	project.update_review(status="review")
	try:
		reviewer.review()
		project.mark_review_finished()
	except Exception as err:
		project.update_review(status="error")
		raise err

	# Finish and export the project
	project.export(EXPORT_PATH)
```

docs/source/guides/api.rst
```diff
@@ -9,35 +9,64 @@ strategy and use it with ASReview.
...
 
-	train_model = LSTMBaseModel()
+	from asreview.models.classifiers import NaiveBayesClassifier
+	from asreview.models.query import MaxQuery
+	from asreview.models.balance import SimpleBalance
+	from asreview.models.feature_extraction import Tfidf
+
+	# Create a project object and folder
+	project = ASReviewProject.create(
+		project_path=Path(PROJECT_PATH),
```
J535D165, line 32:
```suggestion
		project_path=PROJECT_PATH,
```

docs/source/guides/api.rst
```diff
@@ -9,35 +9,64 @@ strategy and use it with ASReview.
...
+the results will be exported as a `.asreview` file.	
 
 .. code-block:: python
 
-	import asreview
-	from asreview.models import LSTMBaseModel
-	from asreview.query_strategies import MaxQueryModel
-	from asreview.balance_strategies import SimpleBalanceModel
-	from asreview.feature_extraction import EmbeddingLSTM
+	from pathlib import Path
```
J535D165, line 19:
```suggestion
```

J535D165 (changes requested):

PR #1091, opened 2022-05-18
Fix min simulate option and make it default

J535D165:
This PR re-implements the `--n_query min` option in the simulation interface. It fixes various bugs and sets the default to `min`. 

The options are now:

```
min => stop after last relevant record is found
1,2,3,... => any int to indicate to stop after that number of iterations
-1 => simulate till the very very end
```

- Also added tests to cover future problems. 
- Changed the insights package as well to work well with this feature.


PeterLombaers:
Seems good to me! The _stop_review looks cleaner and I agree that min is a better default.

PeterLombaers (approved):

J535D165:
Thanks for the review @PeterLombaers!

PR #1093, opened 2022-05-19
Submit prior knowledge search query on pressing Enter

terrymyc:
This PR makes it possible to submit a prior knowledge search query when pressing Enter on the keyboard.


J535D165:
Thanks. users will like.

PR #1095, opened 2022-05-19
Fix page not found due to trailing slash in URL

terrymyc:
This PR fixed page not found due to trailing slash in URL.


J535D165:
But doesn't this break the opposite now?

terrymyc:
Tested on the port 5000. A trailing slash is automatically added and the page is directed to `index.html`.

PR #1096, opened 2022-05-19
Change radius of note field

J535D165:
A minor change to make the design a bit more clean. What do you think?


terrymyc (commented):
I suggest keeping the radius. The latest material design principle even chooses a higher radius for this kind of card. The squared corner is not advocated.

J535D165:
But also for overlapping cards?

terrymyc:
I think the design can be further improved according to the design principle but more work is needed. Let's use the default radius (8px) for now. The squared corner is not advocated in any case.

PR #1100, opened 2022-05-19
Fix issue with RIS exporter for A2-A3-A4 fields

### LINKED ISSUE #1099: Error while downloading RIS results from ASReview Lab
### Discussed in https://github.com/asreview/asreview/discussions/1094

<div type='discussions-op-text'>

<sup>Originally posted by **numland** May 19, 2022</sup>
When I try to download the RIS results from my current project I get an error message: "The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()"
Other download options like Excel work without issues. 

Any suggestions on how to export to RIS?'
</div>

J535D165:
Resolves #1099.

### DIFF
--- a/asreview/io/ris_writer.py
+++ b/asreview/io/ris_writer.py
@@ -56,8 +56,13 @@ class RISWriter():
         # Iterate over all available records
         for rec in records:
 
+            def _notnull(v):
+                if isinstance(v, list):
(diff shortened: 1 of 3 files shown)


J535D165:
This issue should also be merged into the v0.19.x branch.

PR #1105, opened 2022-05-21
Deprecate n_papers option in command line and API

J535D165:


asreview/review/base.py
```diff
@@ -1,4 +1,5 @@
 from abc import ABC
+import logging
```
J535D165, line 2:
```suggestion
import logging
from abc import ABC
```

PR #1116, opened 2022-05-24
Fix embedding not found for lstm models

J535D165:
This is the first part of the fix


asreview/webapp/run_model.py
```diff
@@ -56,9 +56,9 @@ def get_lab_reviewer(as_data,
 
     # LSTM models need embedding matrices.
     if classifier_model.name.startswith("lstm-"):
-        texts = as_data.texts
+        print("download features")
```
J535D165, line 59:
```suggestion
```

PR #1119, opened 2022-05-24
Use icons for DOI and URL variables

J535D165:
Screenshot 2022-05-25 at 00 58 11


terrymyc (approved):
Thanks! Made a few adjustments.

J535D165:
Thanks, very interesting to see your changes.

PR #1120, opened 2022-05-25
Remove aggregated record label statistics

terrymyc:
This PR removed the aggregated statistics of labeled and relevant records for all projects.

image


J535D165:
Thanks, a good starting point to think about new items.

PR #1123, opened 2022-05-25
Fix actionable options for project with error status

terrymyc:
This PR fixed the direct export of the project in setup status. This issue was accidentally introduced in #1052


J535D165:
> @J535D165 Do we allow downloading a project with the `error` status?

Yes, this helps when a user runs into a problem and we want to debug.

PR #1124, opened 2022-05-25
External links use urls file to control url changes

J535D165:
@terrymyc can you help me to exclude `/l/*` from the `Router`? I don't have any idea yet. These URLs should navigate away from the app.


terrymyc (commented):
If you open `/l/*` on `localhost:5000`, it will direct you away from the app. Since navigating to `/l/*` doesn't return `render_template("index.html")`, the react app doesn't take over, and I surmise react-router doesn't work in this case.

J535D165:
Thanks, will test this!

PR #1125, opened 2022-05-27
Improve note field actionable items

terrymyc:
This PR replaced the icon buttons with text buttons in the note field and removed icon buttons with ambiguity.


J535D165:
Merci!

PR #1127, opened 2022-05-27
adjust intro readme

Rensvandeschoot:
This PR proposes to adjust the introduction of ASReview in the readme to make it more generic. Also, the exploration mode is added and hyperlinks are corrected.


J535D165:
Changed the base because changes will have an effect immediately.

PR #1133, opened 2022-05-31
Restore complete progress and recall charts in simulation mode

terrymyc:
This PR restored the complete progress and recall charts of the finished project in Simulation mode.

### DIFF
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -1232,6 +1232,22 @@ def _get_stats(project):
     }
 
 
+def _get_labels(state_obj, priors=False):
+
+    # get the number of records
+    n_records = state_obj.n_records
+
+    # get the labels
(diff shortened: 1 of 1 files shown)


asreview/webapp/api.py
```diff
@@ -1252,9 +1268,16 @@ def api_get_progress_density(project):
     include_priors = request.args.get('priors', False, type=bool)
 
     try:
+        project_config = project.config
```
J535D165, line 1271: `project.config` is optimized in many ways. There is no need to make a copy :) 

The next thing is perfectly fine:

```
            if project.config['reviews'][0]['status'] == 'finished' \
                    and project.config['mode'] == PROJECT_MODE_SIMULATE:
```

J535D165:
Thanks!

PR #1137, opened 2022-06-02
Fix simulation training set bug

PeterLombaers:
In simulation mode, the training set was not properly stored. This PR fixes that.


J535D165:
pfieewww, just in time ;)

PR #1139, opened 2022-06-02
Add project API example notebook

PeterLombaers:
This PR adds a jupyter notebook containing examples of how to use the project and state API to get data from an ASReview project file.


J535D165:
Thanks, very useful.

PR #1142, opened 2022-06-03
Confirm label is saved explicitly during the review

terrymyc:
This PR changed the text of the decision undo bar to explicitly confirm the label is saved.

image


Rensvandeschoot:
Thanks

PR #1164, opened 2022-06-14
Responsive project setup screen

terrymyc:
Changes proposed in this pull request:

- The creating a new project screen is responsive on screens of different ratios.

image
---
image
---
image
---
image
---
image
---
image
---
image
---
image


diederickstoel:
Look great. In the review mode it would be nice if I can swipe articles. Like on Tinder.  “Swipe right” means to accept the article, while “swipe left” means to reject it.

PR #1168, opened 2022-06-22
Add authentication and multi-user functionality for ASReview LAB

cskaandorp:
Make sure you have 2 new Python packages installed: flask-login and SQLAlchemy


mathijsvanderkroft:
Hi, I'm wondering, if a front end user is signed in via the implemented authentication process, is their data on the server side stored as an encrypted file (so mainly talking about the state files here)? And how will this work with multi user labeling tasks?

asreview/webapp/auth.py
```diff
@@ -0,0 +1,144 @@
...
+    username = request.form.get('username').strip().lower()
+    password = request.form.get('password')
+    first_name = request.form.get('first_name', '').strip()
+    last_name = request.form.get('last_name', '').strip()
+    affiliation = request.form.get('affiliation', '').strip()
+    email = request.form.get('email', '').strip()
+    public = bool(int(request.form.get('public', '0')))
+
+    # check if username already exists
+    user = User.query.filter(User.username == username).one_or_none()
```
J535D165, line 88: This will also expose the public flag of the user. Maybe I misunderstood the idea behind the public flag. It's more something like: Open for collaboration.

asreview/webapp/auth.py
```diff
@@ -0,0 +1,144 @@
...
+            'logged_in': current_user.is_authenticated,
+            'username': current_user.username,
+            'id': current_user.id
+        })
+    else:
+        result = (404, 'No user found')
+
+    status, message = result
+    response = jsonify(message)
+    return response, status
```
J535D165, line 129: Interesting API entry. Would be nice to study this in more detail.

asreview/webapp/authentication/login_required.py
```diff
@@ -0,0 +1,80 @@
...
+            # get current user
+            current_user = _get_user()
+
+            if not current_user.is_authenticated:
+                return jsonify({'message': 'login required'}), 401
+
+        # flask 1.x compatibility
+        # current_app.ensure_sync is only available in Flask >= 2.0
+        if callable(getattr(current_app, "ensure_sync", None)):
+            return current_app.ensure_sync(func)(*args, **kwargs)
```
J535D165, line 76: We are flask 2 or later. 
```suggestion

```

asreview/webapp/auth.py
```diff
@@ -0,0 +1,144 @@
...
+@asreview_login_required
+def refresh():
+    if current_user:
+        result = (200, {
+            'logged_in': current_user.is_authenticated,
+            'username': current_user.username,
+            'id': current_user.id
+        })
+    else:
+        result = (404, 'No user found')
```
J535D165, line 125: If `asreview_login_required` is used, can we skip this? Also, `asreview_login_required` returns 401 is there is no user, this returns 404.

asreview/webapp/authentication/login_required.py
```diff
@@ -0,0 +1,80 @@
...
+def asreview_login_required(func):
+    """
+    Adjusted version of login_required of flask-login
+    =================================================
+
+    If you decorate a view with this, it will ensure that the current user is
+    logged in and authenticated before calling the actual view. (If they are
+    not, it calls the :attr:`LoginManager.unauthorized` callback.) For
+    example::
+
```
J535D165, line 33: Link to the source of the function?

asreview/webapp/__init__.py
```diff
@@ -11,3 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from flask_sqlalchemy import SQLAlchemy
+
+# this is a global variable for accessing the database
+DB = SQLAlchemy()
```
J535D165, line 18: High risk of circular imports. This should be avoided imo.

asreview/webapp/start_flask.py
```diff
@@ -102,16 +114,62 @@ def create_app(**kwargs):
...
+            DB.create_all()
+    else:
+        # Don't want to use the standard Anonymous User
+        login_manager.anonymous_user = SingleUser
+
+
+    # Register a callback function for current_user.
+    @login_manager.user_loader
+    def load_user(user_id):
+        return User.query.get(int(user_id))
```
PeterLombaers, line 153: In the documentation they say this should not raise an error, but return None:
https://flask-login.readthedocs.io/en/latest/#how-it-works
If you call this with something other than an integer it will raise a ValueError. Does this matter?

asreview/webapp/auth.py
```diff
@@ -0,0 +1,137 @@
...
+    password = request.form.get('password')
+
+    # get the user
+    user = User.query.filter(User.email == email).one_or_none()
+    # if the user exist proceed and verify
+    if not user:
+        result = (404, {'message': f'User account {email} does not exist.'})
+    else:
+        # verify password
+        if user.verify_password(password):
```
PeterLombaers, line 51: Maybe we can make this a bit more modular, so that it is easier to implement different methods of authenticating. We could make a function `authenticate_user(user, request)` that authenticates a user based on a request, which in the default case would be `def authenticate_user(user, request): return check_password_hash(user.hashed_password, request.form.get('password'))`. In this way, if someone wants to implement a different method of authenticating, they would only need to give a different implementation of `authenticate_user`. Then we could allow extensions of the ASReview authentication in the same way that we allow extension of models.

asreview/project.py
```diff
@@ -58,46 +60,112 @@ class ProjectExistsError(Exception):
...
         The id of the current project.
+    user: User
+        User account object.
     """
 
     if asreview_dir is None:
         asreview_dir = asreview_path()
 
-    return Path(asreview_dir, project_id)
+    if isinstance(user, User) and isinstance(user.id, int):
```
PeterLombaers, line 83: Twice the same if statement. You can remove the outer layer I think.

asreview/project.py
```diff
@@ -236,7 +304,7 @@ def create(cls,
                 f'Project folder {project_path} already exists.')
 
         try:
-            project_path.mkdir(exist_ok=True)
+            project_path.mkdir(parents=True, exist_ok=True)
```
PeterLombaers, line 307: Remove `parents=True`

asreview/utils.py
```diff
@@ -20,6 +20,8 @@
 import numpy as np
 import pkg_resources
 
+from asreview.webapp.authentication.models import SingleUser
```
PeterLombaers, line 23: Remove this import, it isn't used and utils should not be importing from asreview.

asreview/webapp/start_flask.py
```diff
@@ -31,15 +33,24 @@
...
+from asreview.webapp import DB
+from asreview.webapp.authentication.models import (
+    SingleUser,
+    User
+)
+from asreview.webapp.authentication.oauth_handler import OAuthHandler
 
 # set logging level
-if os.environ.get('FLASK_ENV', "") == "development":
+if os.environ.get('FLASK_DEBUG', "") == "1":
```
PeterLombaers, line 47: Maybe support both version? Then we don't have do bump flask in setup.py.

asreview/webapp/start_flask.py
```diff
@@ -102,16 +113,106 @@ def create_app(**kwargs):
 
     # Get the ASReview arguments.
     app.config['asr_kwargs'] = kwargs
+    app.config['AUTHENTICATION_ENABLED'] = \
+        kwargs['enable_auth'] if 'enable_auth' in kwargs else False
```
PeterLombaers, line 117: `kwargs.get('enable_auth', False)`

cskaandorp: Done

asreview/webapp/start_flask.py
cskaandorp: Done

asreview/utils.py
cskaandorp: Done

asreview/project.py
cskaandorp: I like having parents=True in there. Pro: it ensures missing parents are created, it enables the end-user to come up with a path that does not exist (less hassle). Con: typos by end-user will be ignored and create a 'wrong' path. Anyways, let me know if you really want to remove this.

asreview/project.py
cskaandorp: I did.

Question: in the authenticated version we may or may not have collaboration. If a collaborator needs to work on a project, the project_id must come from owner id plus project id. In the current situation I have an else statement in case all fails. That is not supposed to happen. Shall I remove the else clause? The else clause is a fail-safe.

asreview/project.py
```diff
@@ -59,60 +63,126 @@ class ProjectNotFoundError(Exception):
...
+    if isinstance(user, User) and isinstance(user.id, int):
+        # if we do authentication, then the project must be 
+        # registered in the database
+        project_from_db = Project.query.filter(
+            Project.project_id == project_id
+        ).one_or_none()
 
+        # project exists and user -is- owner OR this is a new project
+        if (project_from_db and user == project_from_db.owner) or (
+            project_from_db == None
```
PeterLombaers, line 96: `project_from_db is None` is preferred over `project_from_db == None`. In theory a class can have custom implementation of the comparison operator and here we are really want to know if the value is None.

asreview/project.py
```diff
@@ -59,60 +63,126 @@ class ProjectNotFoundError(Exception):
...
+        # project exists but user is a collaborator
+        elif project_from_db and user in project_from_db.collaborators:
+            folder_id = f"{project_from_db.owner_id}_{project_id}"
 
-        project_path = get_project_path(project_id)
-        project = ASReviewProject(project_path, project_id=project_id)
-        return f(project, *args, **kwargs)
+        # default to project_id
+        else:
+            folder_id = project_id
```
PeterLombaers, line 107: Maybe the best logical sturcture would be:
```[python]
if project_from_db is None:
    folder_id = f"{user.id}_{project_id}"
elif (user == project_from_db.owner) or (user in project_from_db.collaborators):
    folder_id = f"{project_from_db.owner_id}_{project_id}"
else:
    folder_id = project_id
```

asreview/project.py
PeterLombaers: If it's not wanted behavior that the else clause is reached maybe raising an error there is better?

asreview/project.py
```diff
@@ -59,60 +63,126 @@ class ProjectNotFoundError(Exception):
...
-    return decorated_function
+    return Path(asreview_dir, project_folder)
 
 
-def list_asreview_projects():
-    """List the projects in the asreview path"""
+def project_from_id(user):
+    """Decorator function that takes a user account as parameter,
+    the user account is used to get the correct sub folder in which
+    the projects is
```
PeterLombaers, line 120: 'the project is located.'

asreview/project.py
```diff
@@ -59,60 +63,126 @@ class ProjectNotFoundError(Exception):
...
+
+# !=============================================
+# @TODO: Another thing that needs consideration:
+# I need ownership id's to avoid collaborators inviting
+# new collaborators. That means I'd like to alter
+# the ASReviewProject class -> adding an ownership
+# attribute. I could add it on the class itself, but
+# this is only needed in the authenticated version.
+# Instead I will add it dynamically in this function...
+# Feedback is required.
```
PeterLombaers, line 143: I agree that this is a bit ugly. I think I would prefer a `project.owner_id` attribute which can be None. Then you are always sure that the attribute is there. But maybe we should first work towards merging with the asreview/auth-v1 branch and only after try to change this. @J535D165 Do you have an opinion?

asreview/project.py
```diff
@@ -59,60 +63,126 @@ class ProjectNotFoundError(Exception):
     pass
 
 
-def get_project_path(project_id, asreview_dir=None):
+# !=============================================
+# @TODO: I HATE THE FACT THAT I AM TALKING ABOUT
+# Users in this module, but I need a quick fix
+# for now
+# !=============================================
```
PeterLombaers, line 70: I don't think there is a way around talking about users here. I also don't really mind it.

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+
+# TODO: not sure if this file is the right place for this function
+def send_email(msg, config):
+    context = ssl.create_default_context()
+    with smtplib.SMTP_SSL(config["SERVER"], config["PORT"], context=context) as server:
+
+        # login smtp server
+        server.login(config["USERNAME"], config["PASSWORD"])
+        # send email
+        return server.sendmail(msg)
```
PeterLombaers, line 67: Why does this function use `smtplib` and the next function `flask_mail`?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+
+# TODO: not sure if this file is the right place for this function
+def send_forgot_password_email(user, cur_app):
+    # get necessary information out of user object
+    name = user.name or "ASReview user"
+    # email config
+    config = cur_app.config.get("EMAIL_CONFIG")
+    # TODO: this is horrible => what if there is a domain name,
+    # where is it coming from? Where can I get it?
+    root_url = "http://127.0.0.1:3000/"
```
PeterLombaers, line 78: Can this be taken from the config? Not sure what is the right way. Maybe @J535D165 knows?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+
+# TODO: not sure if this file is the right place for this function
+def send_confirm_account_email(user, cur_app):
+    # get necessary information out of user object
+    name = user.name or "ASReview user"
+    # email config
+    config = cur_app.config.get("EMAIL_CONFIG")
+    # TODO: this is horrible => what if there is a domain name,
+    # where is it coming from? Where can I get it?
+    root_url = "http://127.0.0.1:3000/"
```
PeterLombaers, line 108: Same comment as above. Should be fixed.

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+                confirmed = not email_verification
+                # create the User account
+                user = User(
+                    identifier=identifier,
+                    origin=origin,
+                    email=email,
+                    name=name,
+                    affiliation=affiliation,
+                    password=password,
+                    confirmed=confirmed,
```
PeterLombaers, line 217: It's probably more clear to use `not email_verification` remove the `confirmed` variable. I don't like the double negation.

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+        user = User.query.filter(
+            or_(User.identifier == email, User.email == email)
+        ).one_or_none()
+        # return error if user doesn't exist
+        if isinstance(user, User):
+            result = (403, f'User with email "{email}" already exists.')
+        else:
+            # password confirmation is done by front end, so the only
+            # thing that remains is to add the user (password will be
+            # hashed in the User model)
```
PeterLombaers, line 198: The confirmation email is sent below right? Or is that not what you mean here?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+        password = request.form.get("password")
+        public = bool(int(request.form.get("public", "1")))
+
+        # check if email already exists
+        user = User.query.filter(
+            or_(User.identifier == email, User.email == email)
+        ).one_or_none()
+        # return error if user doesn't exist
+        if isinstance(user, User):
+            result = (403, f'User with email "{email}" already exists.')
```
PeterLombaers, line 194: Do we want to give information on whether a user exists or not in these response messages? It will make for an easier UI experience, but maybe less secure. One [source](information: https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#account-creationl) says the message and status_cose should not reveal any information. Do you have an opinion @J535D165 ?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+    return response, status
+
+
+@bp.route("/confirm_account", methods=["POST"])
+def confirm_account():
+    """Confirms account with email verification"""
+
+    if current_app.config.get("EMAIL_VERIFICATION", False):
+        # find user by token and user id
+        user_id = request.form.get("user_id", 0)
```
PeterLombaers, line 265: Can users actually have 0 as a user_id?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+
+    status, message = result
+    response = jsonify({"message": message})
+    return response, status
+
+
+@bp.route("/forgot_password", methods=["POST"])
+def forgot_password():
+
+    if current_app.config.get("EMAIL_VERIFICATION", False):
```
PeterLombaers, line 323: This is missing an else statement.

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+            # set a token
+            user = user.set_token_data(
+                current_app.config["SECRET_KEY"],
+                current_app.config["SECURITY_PASSWORD_SALT"],
+            )
+            try:
+                # store data
+                DB.session.commit()
+                # send email
+                send_forgot_password_email(user, current_app)
```
PeterLombaers, line 347: Should we also handle any error in `send_forget_password_email` here?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+                    )
+                # store user
+                DB.session.add(user)
+                DB.session.commit()
+                # at this stage, if all went well, the User account is
+                # stored in the database, send the verification email
+                # if applicable
+                if email_verification:
+                    # send email
+                    send_confirm_account_email(user, current_app)
```
PeterLombaers, line 235: Should we handle any error from `send_confirm_account_email` here?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+        try:
+            user = user.update_profile(email, name, affiliation, password, public)
+            DB.session.commit()
+            result = (200, "User profile updated")
+        except IntegrityError as e:
+            DB.session.rollback()
+            result = (500, f"Unable to update your profile! Reason: {str(e)}")
+        except SQLAlchemyError as e:
+            DB.session.rollback()
+            result = (500, f"Unable to update your profile! Reason: {str(e)}")
```
PeterLombaers, line 381: Sometimes an SQLAlchemyError gives status_code 500, sometimes 403. We should probably make these and the corresponding messages a bit more uniform.

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+        id = None
+
+    result = {"logged_in": logged_in, "name": name, "id": id}
+
+    response = jsonify(result)
+    return response, 200
+
+
+@bp.route("/signout", methods=["DELETE"])
+@asreview_login_required
```
PeterLombaers, line 438: I probably don't fully understand how this works, but this is the first time I see the `@asreview_login_required` decorator. Does this mean other routes like `/reset_password` don't need it?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+    redirect_uri = request.form.get("redirect_uri", "").strip()
+
+    # assuming we have this provider
+    oauth_handler = current_app.config.get("OAUTH", False)
+    if (
+        isinstance(oauth_handler, OAuthHandler)
+        and provider in oauth_handler.providers()
+    ):
+        # get user credentials for this user
+        response = oauth_handler.get_user_credentials(provider, code, redirect_uri)
```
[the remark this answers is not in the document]
PeterLombaers: Should we handle errors here?

asreview/webapp/api/auth.py
```diff
@@ -0,0 +1,512 @@
...
+                DB.session.commit()
+                created_account = True
+            except SQLAlchemyError:
+                DB.session.rollback()
+                message = "OAuth: unable to create your account!"
+                # return this immediately
+                return jsonify({"data": message}), 500
+
+        # log in this user
+        if bool(user):
```
PeterLombaers, line 496: Is there any way this can evaluate to false? We already know that user is not None.

asreview/webapp/authentication/oauth_handler.py
```diff
@@ -0,0 +1,154 @@
+import requests
+
+
+class OAuthHandler:
+    def __init__(self, configs=None):
+        if not (bool(configs) and isinstance(configs, dict)):
+            raise f"OAuthHandler needs a configuration dictionary."
```
PeterLombaers, line 7: Which exception is being raised?

asreview/webapp/authentication/oauth_handler.py
```diff
@@ -0,0 +1,154 @@
...
+                all(
+                    [
+                        bool(authorization_url),
+                        bool(token_url),
+                        bool(client_id),
+                        bool(secret),
+                    ]
+                )
+            ):
+                raise f"OAuthHandler has insufficient data for f{provider}"
```
PeterLombaers, line 28: Which exception is being raised?

asreview/webapp/authentication/oauth_handler.py
```diff
@@ -0,0 +1,154 @@
...
+        """Extract User credentials with the help of a code"""
+        result = False
+        if provider == "github":
+            result = self.__handle_github(code)
+        elif provider == "orcid":
+            result = self.__handle_orcid(code)
+        elif provider == "google":
+            result = self.__handle_google(code, redirect_uri)
+        else:
+            raise f"Could not find provider {provider}"
```
PeterLombaers, line 65: Which exception is being raised?

asreview/webapp/authentication/oauth_handler.py
```diff
@@ -0,0 +1,154 @@
+import requests
+
+
+class OAuthHandler:
```
PeterLombaers, line 4: Maybe this class could use a dedicated exception like `OAuthException`. That way when this class is called upon somewhere in the code, we know which exception we get if something happens in the oauth process.

asreview/webapp/authentication/oauth_handler.py
```diff
@@ -0,0 +1,154 @@
...
+                    "token_url": token_url,
+                    "client_id": client_id,
+                    "secret": secret,
+                    "scope": scope,
+                }
+        # set the config dictionary
+        self.services = services
+
+    def front_end_params(self):
+        """prepare a service dictionary for front-end: remove secrets"""
```
PeterLombaers, line 42: There should be clear documentation how to use `OAuthHandler` without showing secrets anywhere.

asreview/webapp/authentication/oauth_handler.py
```diff
@@ -0,0 +1,154 @@
...
+        result = {}
+        # remove secret from service parameters
+        for k, v in self.services.items():
```
PeterLombaers, line 45: `for provider, config in self.services.items()` is probably a bit more clear.

asreview/webapp/authentication/oauth_handler.py
```diff
@@ -0,0 +1,154 @@
...
+            result[k] = c
+        return result
+
+    def providers(self):
+        """Returns a list with stored providers"""
+        return list(self.services.keys())
+
+    def get_user_credentials(self, provider, code, redirect_uri=None):
+        """Extract User credentials with the help of a code"""
+        result = False
```
PeterLombaers, line 57: It's not really necessary to declare a variable.

asreview/webapp/authentication/oauth_handler.py
```diff
@@ -0,0 +1,154 @@
...
+                "client_id": params["client_id"],
+                "client_secret": params["secret"],
+                "grant_type": "authorization_code",
+                "scope": "/authenticate",
+            },
+            headers={"Accept": "application/json"},
+        ).json()
+        id = response["orcid"]
+        name = response.get("name", "")
+        # TODO@Casper: I don't understand why I can't
```
PeterLombaers, line 85: Is this still an issue?

asreview/webapp/api/projects.py
```diff
@@ -171,102 +186,150 @@ def api_get_projects_stats():  # noqa: F401
...
         project_config = upgrade_project_config(project_config)
         project_config["projectNeedsUpgrade"] = True
 
+    # add user_id of owner to response if authenticated
+    if current_app.config["AUTHENTICATION_ENABLED"]:
+        db_project = Project.query.filter(
+            Project.project_id == project.config.get("id", 0)
+        ).one_or_none()
+        if db_project:
+            project_config["ownerId"] = db_project.owner_id
```
PeterLombaers, line 267: This should be "owner_id" I think.

asreview/webapp/api/projects.py
```diff
@@ -171,102 +186,150 @@ def api_get_projects_stats():  # noqa: F401
...
+@bp.route("/projects/<project_id>/info", methods=["PUT"])
+@asreview_login_required
+@project_from_id(current_user)
 def api_update_project_info(project):  # noqa: F401
     """Get info on the article"""
 
     # rename the project if project name is changed
-    if request.form.get('name', None) is not None:
-        project.rename(request.form['name'])
+    if request.form.get("name", None) is not None:
```
PeterLombaers, line 279: This could be more clear:
```[python]
new_project_name = request.form.get("name", None)
if (new_project_name is not None) and (project.config["name] != new_project_name:
   ....
```

asreview/webapp/api/projects.py
```diff
@@ -1529,13 +1612,31 @@ def api_delete_project(project):  # noqa: F401
...
+            # remove from database if applicable
+            if current_app.config["AUTHENTICATION_ENABLED"]:
+                project = Project.query.filter(
+                    and_(
+                        Project.project_id == project.project_id,
+                        Project.owner_id == current_user.id,
+                    )
+                ).one_or_none()
+
+                if project != None:
```
PeterLombaers, line 1625: `project is not None`

asreview/webapp/api/projects.py
```diff
@@ -1529,13 +1612,31 @@ def api_delete_project(project):  # noqa: F401
...
+                    DB.session.delete(project)
+                    DB.session.commit()
+                else:
+                    return jsonify(message="Failed to delete project in DB."), 500
+
+            # and remove the folder
             shutil.rmtree(project.project_path)
+
         except Exception as err:
```
PeterLombaers, line 1634: If this exception is triggered, there was an error during `shutil.rmtree`. In this case the project is deleted from the database, but the project folder was not yet deleted. Maybe move the database commit to after this exception handler and add a rollback to the exception handler.

asreview/webapp/authentication/login_required.py
```diff
@@ -0,0 +1,84 @@
...
+            if not (bool(current_user) and current_user.is_authenticated):
+                return jsonify({"message": "login required"}), 401
+
+        # # TODO@Jonathan/Casper
+        # # flask 1.x compatibility <= WE REQUIRE FLASK >= 2.0 but
+        # # this function was used by Flask. Don't understand.
+        # # current_app.ensure_sync is only available in Flask >= 2.0
+        # if callable(getattr(current_app, "ensure_sync", None)):
+        #     return current_app.ensure_sync(func)(*args, **kwargs)
```
PeterLombaers, line 80: Is this solved?

asreview/webapp/authentication/models.py
```diff
@@ -0,0 +1,272 @@
...
+class User(UserMixin, DB.Model):
+    """The User model for user accounts."""
+
+    __tablename__ = "users"
+    id = Column(Integer, primary_key=True)
+    identifier = Column(String(100), nullable=False)
+    origin = Column(String(100), nullable=False)
+    email = Column(String(100), unique=True)
+    name = Column(String(100))
+    affiliation = Column(String(100))
```
PeterLombaers, line 46: Do we really need an affiliation for the user? In my opinion we should leave this one out,  but please convince me otherwise :) @J535D165 What do you think?

asreview/webapp/authentication/models.py
```diff
@@ -0,0 +1,272 @@
...
+from asreview.webapp import DB
+
+
+class User(UserMixin, DB.Model):
+    """The User model for user accounts."""
+
+    __tablename__ = "users"
+    id = Column(Integer, primary_key=True)
+    identifier = Column(String(100), nullable=False)
+    origin = Column(String(100), nullable=False)
```
PeterLombaers, line 43: This one is related to the source of the password, right? A docstring explaining the use of each user attribute would be nice. Also something in the documentation.

asreview/webapp/authentication/models.py
```diff
@@ -0,0 +1,272 @@
...
+    @validates("name")
+    def validate_origin(self, _key, name):
+        if not bool(name):
+            raise ValueError("Name is required")
+        return name
+
+    # this is pretty weird, it 'loops', starting at
+    # origin, stores origin because of the return, moves
+    # on to email and now self contains origin so
+    # I can compare it... and so on.
```
PeterLombaers, line 86: Could you explain what is happening here?

PeterLombaers: Is this also where password requirements are checked?

asreview/webapp/authentication/models.py
```diff
@@ -0,0 +1,272 @@
...
+        if bool(self.token) and bool(self.token_created_at):
+            # what is now
+            now = dt.datetime.utcnow()
+            # get time-difference in hours
+            diff = now - self.token_created_at
+            # give me hours and remaining seconds
+            [hours, r_secs] = divmod(diff.total_seconds(), 3600)
+            # return if token is correct and we are still before deadline
+            return self.token == provided_token and (
+                hours <= max_hours or (hours == max_hours and r_secs < 60)
```
PeterLombaers, line 184: So the token is valid for max_hours + 60 seconds if I understand it correctly? Any reason why it's like this? It seems a bit complicated.

asreview/webapp/authentication/models.py
```diff
@@ -0,0 +1,272 @@
...
+from itsdangerous import URLSafeTimedSerializer
+from sqlalchemy import Boolean
+from sqlalchemy import Column
+from sqlalchemy import DateTime
+from sqlalchemy import ForeignKey
+from sqlalchemy import Integer
+from sqlalchemy import String
+from sqlalchemy.orm import relationship
+from sqlalchemy.orm import validates
+from sqlalchemy.orm.session import Session
```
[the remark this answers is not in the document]
cskaandorp: Removed

asreview/webapp/authentication/models.py
cskaandorp: Giving it a shot here: if you have different people from different institutes it might be helpful to have this information if you want to search people to collaborate with. How about reversing this: convince me why it should be left out. It's just a small piece of information that tells me who you are and where you work in a collaborating environment :)

asreview/webapp/api/projects.py
```diff
@@ -171,102 +186,150 @@ def api_get_projects_stats():  # noqa: F401
...
     return response
 
 
-@bp.route('/projects/<project_id>/info', methods=["GET"])
-@project_from_id
+@bp.route("/projects/<project_id>/info", methods=["GET"])
+@asreview_login_required
+@project_from_id(current_user)
 def api_get_project_info(project):  # noqa: F401
     """Get info on the article"""
```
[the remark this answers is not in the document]
cskaandorp: I have removed these docstrings for now. I would like to have all the code reviewed first and then take care of the docstrings. Is that OK?

asreview/webapp/api/auth.py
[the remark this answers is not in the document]
cskaandorp: I have create the one-liner

asreview/webapp/api/auth.py
cskaandorp: It's just me being paranoid, I have removed the if statement.

asreview/webapp/authentication/models.py
cskaandorp: No, this comment should have been removed. I made a note because the thing was curious, and I forgot to remove it. Password requirements are not checked here! I am not sure password requirements are checked in the back-end! But in the table we only store the hashed password which doesn't require a validation other than being there if the origin of the account is 'asreview'.

PR #1195, opened 2022-08-11
Converter rollback

### LINKED ISSUE #1191: Rollback for project converter
It would be nice to have a rollback function for the converter between version 0 and version 1. All the data is stored in the `legacy` folder in the version 1 project, so this should not be difficult to implement. This would make it easier to solve problems in the project conversion when the original version 0 project was already corrupt, as in (#1190).

PeterLombaers:
This pull requests adds a function to roll back the conversion of a project file from v1 back to v0, as in (closes #1191 ).

### DIFF
--- a/asreview/state/sql_converter.py
+++ b/asreview/state/sql_converter.py
@@ -572,3 +572,58 @@ def create_decision_changes_table(sql_fp):
                                     time INTEGER)''')
 
         con.commit()
+
+
+def rollback_conversion(fp, from_version=1, to_version=0):
+    if from_version != 1 and to_version != 0:
+        raise ValueError(
+            f"Not possible to roll back conversion from v{from_version} "
+            f"to v{to_version}."
+        )
+
+    if not is_converted_project(fp):
+        raise ValueError(f"Project file at {fp} is not a converted "
+                         f"project file.")
+
+    fp = Path(fp)
+    legacy_fp = Path(fp, 'legacy')
+
+    # Delete everything other than the legacy folder.
+    for item in fp.iterdir():
+        if item.is_file():
+            item.unlink()
+        elif item.is_dir() and (item.name != 'legacy'):
+            shutil.rmtree(item)
+        else:
+            pass
+
+    # Copy from legacy folder and delete it after.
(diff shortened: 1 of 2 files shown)


tests/test_converter.py
```diff
@@ -1,12 +1,19 @@
...
 from asreview.project import open_state
+from asreview.state.legacy.json import JSONState
 from asreview.state.legacy.utils import open_state as open_state_legacy
-from asreview.state.sql_converter import upgrade_asreview_project_file
+from asreview.state.legacy.utils import state_from_file
+from asreview.state.sql_converter import (
+    is_converted_project,
+    rollback_conversion,
+    upgrade_asreview_project_file
+)
```
J535D165, line 16:
```suggestion
from asreview.state.sql_converter import  is_converted_project
from asreview.state.sql_converter import  rollback_conversion
from asreview.state.sql_converter import  upgrade_asreview_project_file
```

J535D165:
Thanks for this PR! I'm wondering where you want to implement the rollback. It's not added to the try-except block atm isn't it?

PeterLombaers:
> 

Yeah, I didn't think about where to put it yet, right now you can use it as a debug tool. It would make sense to put a try-except in the upgrade_project function, I guess? If anything goes wrong in the conversion it is always logical to roll back. Should there be one in the API call?

J535D165:
I think you should add it to the converter itself. And then raise the error again after the rollback. Similar to SQLite module.

PeterLombaers:
I added the try-except.

asreview/state/sql_converter.py
```diff
@@ -105,65 +109,72 @@ def upgrade_asreview_project_file(fp, from_version=0, to_version=1):
...
-        json.dump(project_config_new, f)
+        # dump the project json
+        with open(Path(fp, 'project.json'), 'w') as f:
+            json.dump(project_config_new, f)
+    except Exception as e:
+        rollback_conversion(fp, check_is_converted=False)
+        raise StateConversionException(
+            f"An error occurred during conversion of state "
+            f"from {from_version} to {to_version}."
+        ) from e
```
J535D165, line 177: Nice, learned something new today (the `from e`)

PR #1239, opened 2022-10-05
Update deduplication

gimoAI:
We left this out at first with the assumption that data never has both title and abstract missing. However, after some testing this turns out to be quite a common problem. Take the following data:

| DOI    | Title | Abstract |
|--------|-------|----------|
| 10.100 |       |          |
| 10.200 |       |          |

Now it will still remove the second entry, this PR prevents that from happening.

### DIFF
--- a/asreview/data/base.py
+++ b/asreview/data/base.py
@@ -486,7 +486,7 @@ class ASReviewData():
             .str.lower().str.strip().replace("", None)
 
         # save boolean series for duplicates based on titles/abstracts
-        s_dups_text = ((s.duplicated(keep=keep)) & (s.notnull()))
+        s_dups_text = ((s.duplicated()) & (s.notnull()))
 
         # final boolean series for all duplicates
         if s_dups_pid is not None:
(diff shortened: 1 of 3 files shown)


asreview/data/base.py
```diff
@@ -240,7 +240,7 @@ def texts(self):
             return self.title
 
         cur_texts = np.array([
-            self.title[i] + " " + self.abstract[i] for i in range(len(self))
+            self.title[i] + self.abstract[i] for i in range(len(self))
```
J535D165, line 243: Why concatenate the last word of the title and the first word of the abstract? This affects the models

J535D165 (commented):
Looks good to me!

asreview/data/base.py
gimoAI: yes you are right. I though this may cause an issue when they were both empty but it doesn't.

J535D165:
Can you fix the broken tests?

gimoAI:
> test-core gives a trailing whitespace error, but those trailing whitespaces are supposed to be there 😄

Instead of using StringIO I added a csv test file to demo_data. This saves an import and many lines of code.

J535D165 (approved):

J535D165:
Nice work!!

Issue #1232, opened 2022-09-30
Not possible to read RIS files from remote location/URL


J535D165:
```python
>>> from asreview import ASReviewData
>>> d = ASReviewData.from_file("https://raw.githubusercontent.com/asreview/systematic-review-datasets/master/datasets/Appenzeller-Herzog_2020/output/Appenzeller-Herzog_2020.csv")
>>> d.df
                                                       title  ... duplicate_record_id
record_id                                                     ...                    
0          Binding Selectivity of Methanobactin from Meth...  ...                 NaN
1          Characteristics and prevalence of Wilson's dis...  ...                 NaN
2          Modality of treatment and potential outcome of...  ...                 NaN
3          The early history of manganese and the recogni...  ...                 NaN
4          Zinc monotherapy for young children with presy...  ...                 NaN
...                                                      ...  ...                 ...
3448       Dietary copper intake in artificially fed infants  ...                 NaN
3449       WTX101 in patients newly diagnosed with Wilson...  ...              3403.0
3450       A two-center randomized controlled feasibility...  ...                 NaN
3451       Zinc-carnosine and vitamin E supplementation d...  ...                 NaN
3452       A clinical study of bone mesenchymal stem cell...  ...                69.0

[3453 rows x 10 columns]
>>> d_ris = ASReviewData.from_file("https://raw.githubusercontent.com/asreview/systematic-review-datasets/master/datasets/van_de_Schoot_2017/raw/schoot-lgmm-ptsd-included-3.ris")
WARNING:root:[Errno 2] No such file or directory: 'https://raw.githubusercontent.com/asreview/systematic-review-datasets/master/datasets/van_de_Schoot_2017/raw/schoot-lgmm-ptsd-included-3.ris'
WARNING:root:[Errno 2] No such file or directory: 'https://raw.githubusercontent.com/asreview/systematic-review-datasets/master/datasets/van_de_Schoot_2017/raw/schoot-lgmm-ptsd-included-3.ris'
WARNING:root:[Errno 2] No such file or directory: 'https://raw.githubusercontent.com/asreview/systematic-review-datasets/master/datasets/van_de_Schoot_2017/raw/schoot-lgmm-ptsd-included-3.ris'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/Bruin056/Documents/GitHub/asreview/asreview/data/base.py", line 187, in from_file
    df, column_spec = reader.read_data(fp)
[...]

PR #1246, opened 2022-10-18
Read RIS files from URL

gimoAI:
Add handling for RIS files from remote location / URL, related to issue #1232

### DIFF
--- a/asreview/io/ris_reader.py
+++ b/asreview/io/ris_reader.py
@@ -170,10 +170,3 @@ class RISReader():
             # Return the standardised dataframe
             return _standardize_dataframe(df)
 
-
-if __name__ == '__main__':
-    from asreview import ASReviewData
-    ris_url = "https://raw.githubusercontent.com/asreview/systematic-review-datasets/master/datasets" \
-              "/van_de_Schoot_2017/raw/schoot-lgmm-ptsd-included-3.ris"
-    d = ASReviewData.from_file(ris_url)
-    print(d.df)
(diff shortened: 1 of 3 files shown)


asreview/io/ris_reader.py
```diff
@@ -124,20 +122,35 @@ def read_data(cls, fp):
...
-            except UnicodeDecodeError:
-                pass
-            except IOError as e:
-                logging.warning(e)
         if entries is None:
-            raise ValueError("Cannot find proper encoding for data file.")
+            for encoding in encodings:
+                if is_url(fp):
+                    try:
+                        url_input = requests.get(fp)
```
J535D165, line 131: I think you can urllib to prevent a new dep.

J535D165 (commented):
Nice PR, a test would be very nice.

setup.py
```diff
@@ -123,7 +123,7 @@ def get_cmdclass():
         'openpyxl',
         'gevent',
         'jsonschema',
-        'filelock'
+        'filelock',
```
J535D165, line 126:
```suggestion
        'filelock'
```

asreview/io/ris_reader.py
```diff
@@ -124,20 +122,36 @@ def read_data(cls, fp):
...
+                        bibliography_file = io.StringIO(
+                            url_input.read().decode(encoding)
+                        )
+
+                        entries = list(
+                            rispy.load(bibliography_file, skip_unknown_tags=True)
+                        )
+                        bibliography_file.close()
+                    except UnicodeDecodeError:
+                        pass
```
J535D165, line 141: It is more efficient to iter the encodings after you downloaded the file. However, encoding detection might not be efficient at all and can be improved. Maybe even upstream. Just my brainwave.

Issue #1233, opened 2022-09-30
ASReviewData can't handle URL without file extension in URL


J535D165:
We need content disposition checks. 

See https://stackoverflow.com/questions/31804799/how-to-get-pdf-filename-with-python-requests

```python
from asreview import ASReviewData
ASReviewData.from_file("https://osf.io/t397h/download")
```

J535D165:
https://github.com/asreview/asreview/blob/45b8600bfd3f11e7e92fe4fed387317ae546ceb4/asreview/data/base.py#L179-L181

PR #1249, opened 2022-10-19
URL extension checks

gimoAI:
Enable ASReviewData to handle URLs without file extensions as mentioned in #1233 

Also updates the test for URL without extension containing RIS dataset.


J535D165 (approved):
Nice!!

PR #1256, opened 2022-11-02
Add , "--model_seed" as argument for the seed selection

JTeijema:
This PR adds `--model_seed` as an alternative to `--seed` as there also exists an `--init_seed` argument, making the original name of the seed ambiguous. `--seed` remains for backward compatibility.

### DIFF
--- a/asreview/entry_points/base.py
+++ b/asreview/entry_points/base.py
@@ -93,7 +93,7 @@ def _base_parser(prog=None, description=None):
              "and parameter values."
     )
     parser.add_argument(
-        "--seed",
+        "--seed", "--model_seed",
         default=None,
         type=int,
         help="Seed for the model (classifiers, balance "


lastoel (commented):
If I understand correctly, both arguments "--seed" and "--model.seed" can now be used to call the same function within ASReview. Using "--model.seed" as end-user is clearer and matches with the Makita syntax. 

Would you like me include changing the argument --seed into --model.seed in the ASReview RTD in my PR to update the RTD?

Rensvandeschoot:
I agree with using model/seed to refer to specific functionality. @J535D165 ?

JTeijema:
@J535D165 I wasn't planning on this. I just want to add the additional option, we can still deprecate it later.

J535D165:
We will focus on improving the workflow in asreview-makita and leave this for now.

PR #1267, opened 2022-11-10
Prepare SqlState for adding extra columns

PeterLombaers:
This pull requests makes all 'INSERT' statements used in the SqlState explicit about which columns they insert into. 

Before, when data was inserted into all columns of a table, the column names were not explicitly set. This means that if a future version of the state class adds a column to the table, then the insert statement will have to change as well. Then it will no longer be backward compatible.

Now, all insert statements mention explicitly into which column they are inserting. If a column is added to a table, the insert statement will keep on working. A test has been added to verify that adding a column to the results table does not break anything.

### DIFF
--- a/asreview/state/sqlstate.py
+++ b/asreview/state/sqlstate.py
@@ -423,7 +423,9 @@ class SQLiteState(BaseState):
         cur = con.cursor()
         cur.execute("DELETE FROM record_table")
         cur.executemany(
-            """INSERT INTO record_table VALUES (?)""", record_sql_input)
+            "INSERT INTO record_table (record_id) VALUES (?)",
+            record_sql_input
+        )
         con.commit()
 
     def add_last_probabilities(self, probabilities):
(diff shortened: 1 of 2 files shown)


J535D165:
Nice contribution :) 

I wonder why you are changing the xfail in [e54752e](https://github.com/asreview/asreview/pull/1267/commits/e54752e7764c8a4e2ae7de4f1dfa3e3797befed6). We are losing test functionality in this way: https://docs.pytest.org/en/7.1.x/how-to/skipping.html.

PeterLombaers:
In this case, we are not expecting a test to fail because code is under development or some other reason. We are testing for an error that we want to occur in a specific situation. For example, we want a `ProjectExistsError` to occur if you try to initialize a new project in the folder of an existing one. The test should pass if the error occurs, and the test should fail if the error doesn't occur.
https://docs.pytest.org/en/latest/how-to/assert.html#assertions-about-expected-exceptions

J535D165:
Clear!

J535D165 (approved):

PR #1275, opened 2022-11-24
Fix RIS export userwarning

gimoAI:
In some cases, when exporting in RIS format, the following userwarning is shown:
```
C:\Python310\lib\site-packages\rispy\writer.py:114: UserWarning: label `included` not exported
  warnings.warn(UserWarning(f"label `{label}` not exported"))
```
The 'included' label is properly processed to the output file despite the warning.

To solve this, I propose popping the 'included' key from the dict. Performance remains equal (tested a couple of runs on the Brouwers_2019 dataset).

### DIFF
--- a/asreview/io/ris_writer.py
+++ b/asreview/io/ris_writer.py
@@ -71,17 +71,12 @@ class RISWriter():
                 except Exception:
                     rec_copy[m] = []
 
-            # Relevant records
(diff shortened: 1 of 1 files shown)


gimoAI:
If this PR is approved, I will also remove the code in datatools that is used to catch the warnings there.

J535D165 (approved):
Outstanding contribution

PR #1276, opened 2022-11-25
Update gevent dep min version

JTeijema:
As discovered in #1273, the minimal version of the `gevent` package dependency is 20.4.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ setup(
         'flask>=2.0',
         'flask_cors',
         'openpyxl',
-        'gevent>=20.4.0',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -121,7 +121,7 @@ def get_cmdclass():
         'flask>=2.0',
         'flask_cors',
         'openpyxl',
-        'gevent',
+        'gevent>=20.4.0',
```
J535D165, line 124:
```suggestion
        'gevent>=20',
```

J535D165:
Thanks!

PR #1311, opened 2023-02-08
Add url to dataformat description

LukasWallrich:


J535D165:
Thanks again for contributing to ASReview!

PR #1314, opened 2023-02-14
Fix bug in type_from_column_spec

rohitgarud:
Fixes bug in type_from_column_spec, which is used in _standardize_dataframe. The bug was in the variable names while unpacking the col_specs. Here column_spec_name should be the key and the data type should be the value.

### DIFF
--- a/asreview/io/utils.py
+++ b/asreview/io/utils.py
@@ -27,7 +27,22 @@ from asreview.utils import list_writer_names
 
 
 def type_from_column(col_name, col_definitions):
-    """Transform a column name to its standardized form."""
+    """Transform a column name to its standardized form.
+
+    Arguments
+    ---------
+    col_name: str
+        Name of the column in the dataframe.
+    col_definitions: dict
+        Dictionary of {standardized_name: [list of possible names]}.
+        Ex. {"title": ["title", "primary_title"],
+            "authors": ["authors", "author names", "first_authors"]}
+
+    Returns
+    -------
+    str:
+        The standardized name. If it wasn't found, return None.
+    """
     for name, definition in col_definitions.items():
         if col_name.lower() in definition:
             return name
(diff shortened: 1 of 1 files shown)


J535D165:
Interesting! Thanks for the contribution. 

I agree that the documentation is not in line with behavior and therefore this is a bug. 

What confuses me is the difference between the default implementation and the `column_spec` implementation. 
First:
https://github.com/asreview/asreview/blob/942f755368e71cd7abe77232a3cc292e98a66465/asreview/io/utils.py#L117-L120
Second:
https://github.com/asreview/asreview/blob/942f755368e71cd7abe77232a3cc292e98a66465/asreview/io/utils.py#L121-L124

The second accepts column specs in this format:

https://github.com/asreview/asreview/blob/942f755368e71cd7abe77232a3cc292e98a66465/asreview/config.py#L48-L65

A small refactor can align the two (where we replace the `column_spec` implementation with the default format). What do you think?

rohitgarud:
Hi @J535D165, thank you for your quick response. I found this bug because I was trying to use the `type_from_column_spec` function. 
From what I understand, FIRST and SECOND can be achieved with a single function `type_from_column`, and `type_from_column_spec` is not at all necessary. If the users want to define custom column_specs (or definitions), they can use the same format as `COLUMN_DEFINITIONS` and use `type_from_column`

J535D165:
Indeed! Sorry for the confusion. Interested in deprecating that function in our code base?

rohitgarud:
Should I refactor the code and create a pull request? How w can go about it?

J535D165:
Yes, you can refactor the code in this PR (no need to start a new one). I don't think there is a need to add a deprecation warning as this is not a public part of the API and you are probably the first one to use it ;) Internally, it is never used haha

rohitgarud:
Yeah, I was thinking the same. I will refactor the code. Do you think any additional tests are needed or the current tests will suffice?

J535D165:
Let me rerun. Looks like a timeout error. Nice refactor, everything looks good to me!

rohitgarud:
Thank you..Passed all tests :) .. Will this change be available right away if I upgrade ASReview using pip?

J535D165:
It was a pleasure to merge your first contribution to asreview core. Looking forward to colab in the future!

J535D165:
The new version will not be available immediately, but I will create a new release soon.

rohitgarud:
Thank you @J535D165 . I was my first ever contribution to a big open-source project. I am working on few things and plan to contribute to ASReview frequently in future as I see it as a great opportunity for me to learn. Thank you for your guidance.

PR #1316, opened 2023-02-16
Add Endnote XML file reader

rohitgarud:
Adds Endnote XML file reader which can be used to import XML files generated using Endnote reference manager. There are many open source systematic review datasets available in Endnote xml format, which can be added to the dataset repository

### DIFF
--- a/asreview/io/xml_reader.py
+++ b/asreview/io/xml_reader.py
@@ -1,0 +1,118 @@
+import pandas as pd
+import xml.etree.ElementTree as ET
+
+from asreview.io.utils import _standardize_dataframe
+
+
+class EndnoteXMLReader:
+    """Endnote XML file reader."""
+
+    read_format = [".xml"]
+    write_format = [".csv", ".tsv", ".xlsx"]
+
+    @classmethod
+    def read_data(cls, fp):
+        """Import dataset from Endnote XML file.
+
+        Arguments
+        ---------
+        fp: str, pathlib.Path
+            File path to the XML file.
+
+        Returns
+        -------
+        list:
+            List with entries.
+        """
+        tree = ET.parse(fp)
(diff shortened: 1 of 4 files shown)


rohitgarud:
Is there a particular reason for not implementing this? I think it is a reasonable thing to do to track the records.
https://github.com/asreview/asreview/blob/9064aa61bf89721a9e56e1c49291c3782b4e7089/asreview/io/utils.py#L157-L169

asreview/io/xml_reader.py
```diff
@@ -0,0 +1,119 @@
...
+import pandas as pd
+
+from asreview.io.utils import _standardize_dataframe
+
+
+class EndnoteXMLReader:
+    """Endnote XML file reader."""
+
+    read_format = [".xml"]
+    write_format = [".csv", ".tsv", ".xlsx"]
```
J535D165, line 12:
```suggestion
    write_format = [".csv", ".tsv", ".xlsx", ".ris"]
```

@terrymyc what does it imply to add RIS here?

PeterLombaers:
@rohitgarud We've changed in #1591 how the frontend finds the available data reader classes. This means that a custom data reader like the one of this pull request should be easy to implement as an ASReview extension, just like you can make you own model extension. I'll close this pull request because we think that the reader is a bit too specific to include in the base code of ASReview, but I hope you can make it work as an extension, because it does look useful!

rohitgarud:
Sure.. makes sense.. will try to develop an extension for this reader

J535D165:
Would be great @rohitgarud. And sorry it took so long. We are always very grateful for your contributions.

rohitgarud:
No worries.. I was just exploring things at that time.. I want to start contributing again but not getting time due to other commitments..

PR #1317, opened 2023-02-16
Refactor get project path

PeterLombaers:
This pull request moves the use of the database models User and Project to webapp. I also think that it is no longer necessary to dynamically add the owner_id attribute to the ASReviewProject instances, since these are only used in webapp/api/projects.py. I think that tests not related to javascript now pass.

### DIFF
--- a/asreview/project.py
+++ b/asreview/project.py
@@ -42,8 +42,6 @@ from asreview.config import SCHEMA
 from asreview.state.errors import StateNotFoundError
 from asreview.state.sqlstate import SQLiteState
 from asreview.utils import asreview_path
-from asreview.webapp.authentication.models import Project
-from asreview.webapp.authentication.models import User
 from asreview.webapp.io import read_data
 
 PATH_PROJECT_CONFIG = "project.json"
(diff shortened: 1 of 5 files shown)


asreview/project.py
```diff
@@ -132,46 +94,24 @@ def decorated_function(project_id, *args, **kwargs):
...
-# @TODO: Another thing that needs consideration:
-# I need ownership id's to avoid collaborators inviting
-# new collaborators. That means I'd like to alter
-# the ASReviewProject class -> adding an ownership
-# attribute. I could add it on the class itself, but
-# this is only needed in the authenticated version.
-# Instead I will add it dynamically in this function...
-# Feedback is required.
-# !=============================================
-def list_asreview_projects(user=None):
```
J535D165, line 145: I think we should deprecate this one. This is public API and there can be users. I like the rename btw.

J535D165:
Thanks a lot!!

asreview/project.py
PeterLombaers: You mean make it a private function, or remove it altogether? Right now `get_projects` is used only in `webapp/api/projects.py` (2 times) and `webapp/start_flask.py`. It would be easy enough to completely remove it. Though I guess it could be useful so making it private would make some sense.

PR #1326, opened 2023-02-21
Add return key option to URL data upload 

J535D165:


asreview/webapp/src/ProjectComponents/SetupComponents/DataComponents/DatasetFromURL.js
```diff
@@ -30,9 +30,16 @@ const DatasetFromURL = (props) => {
     if (props.isAddDatasetError) {
       props.reset();
     }
+
```
J535D165, line 33:
```suggestion
```

PR #1331, opened 2023-02-28
Add option to add files from DOI and improve data upload from URL UX

J535D165:
This PR introduces importing datasets from URL and DOI via [Datahugger](https://github.com/J535D165/datahugger). Datahugger is a new technology that makes it possible to import files from 200+ data repositories via the DOI. 

## **FROM URL**

From URL is still more or less identical in UI. The main difference is the 2 step approach. In this PR, you first have to look up the URL, and the filename is returned. Thereafter the dataset is uploaded by clicking ADD. This has additional benefits not directly clear to the user, but there is support for more URLs due to the initial validation and lookup. 

localhost_3000_projects(iPad Mini)

localhost_3000_projects(iPad Mini) (1)

 ## **FROM DOI**

This is the cool part. The DOI can be used to build a highly reproducible workflow. First, you publish your search result and thereafter, you start labeling. Like preregistration of data. This makes it also very easy to simulate the performance of a dataset from an existing systematic review. See the example here with `Yu, Zhe, Kraft, Nicholas, & Menzies, Tim. (2017). Data sets for FASTREAD (v1.3) [Data set]. Zenodo.` https://doi.org/10.5281/zenodo.1162952. 

localhost_3000_projects(iPad Mini) (2)
localhost_3000_projects(iPad Mini) (3)
localhost_3000_projects(iPad Mini) (4)
localhost_3000_projects(iPad Mini) (5)

## Remarks

@terrymyc, I really need a refactor of the AddDataset.js module to improve the workflows and build the OpenAlex pipeline. I think the react query calls should not be handled in AddDataset.js anymore, but only in the subcomponents. In this PR, I can hack myself around it, but it is needed for further improvements. We need to check your work before we can work on this.

### DIFF
--- a/asreview/data/base.py
+++ b/asreview/data/base.py
@@ -68,7 +68,7 @@ def load_data(name, *args, **kwargs):
 def _get_filename_from_url(url):
 
     if not is_url(url):
-        raise ValueError(f"{url} is not a valid URL.")
+        raise ValueError(f"'{url}' is not a valid URL.")
 
     if Path(urlparse(url).path).suffix:
         return Path(urlparse(url).path).name, url
--- a/asreview/webapp/api.py
+++ b/asreview/webapp/api.py
@@ -331,9 +331,21 @@ def api_upload_data_to_project(project):  # noqa: F401
         url = request.form.get('url')
         filename, url = _get_filename_from_url(url)
 
-        if not filename or not Path(filename).suffix:
-            # TODO use datahugger in the future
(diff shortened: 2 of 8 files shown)


Rensvandeschoot:
This is truly a great addition to the way datasets can be imported! It is also a great showcase why open science is so important!!!

terrymyc (approved):
Thanks, @J535D165! This feature extends the use case. I made some UI improvements. See the new screenshot below.

image

Issue #1250, opened 2022-10-21
Error message when trying to export a dataset from running ASReview on a server


markvaneck:
**Describe the bug**
When running ASReview on a server, I receive an error when I am trying to do an export. This error only shows when I want to export as a dataset, but not as a project. The following error message is shown:  
Failed to export the csv dataset. [Errno 2] No such file or directory: '/usr/local/lib/python3.8/site-packages/asreview/webapp/projects/demo/tmpm9zb3r19/export_dataset.csv' Please try again.

**To Reproduce**
Select dataset when exporting a file when ASReview is running on a server, and then export. 

**Screenshots**
![image](https://user-images.githubusercontent.com/47869770/197163923-c5c8be3e-6974-4758-a82d-cc3d34a2677e.png)

**Version information**
 - OS: Linux (not sure which distro)
 - ASReview version: 1.0

Rensvandeschoot:
Could you try to upgrade to v1.0.3?

markvaneck:
Thank you for the quick reply! I will relay the message to the party that hosts our server and I'll get back to you.

Rensvandeschoot:
You can always extract the data from the `.asreview` file via the [API](https://asreview.readthedocs.io/en/latest/example_api_asreview_file.html)

J535D165:
Is the path to the project folder adjusted? It looks like a very odd path to me. There shouldn't be data in that place. 

> Could you try to upgrade to v1.0.3?

I don't think this will solve the issue.

markvaneck:
Firstly, apologies for the late follow-up. 

Secondly, @J535D165 was correct, Our hosting party had changed the location where projects were saved, because that was not a practical place for their server. He said he changed that via an `ASREVIEW_PATH` environment variable as per the documentation. After removing this, exporting projects is working as normal once again.

J535D165:
I think the server admin set the environment variable to a relative path instead of an absolute path. Absolute paths are a better solution in my opinion. Therefore the projects were stored in the application folder (not a good idea because of upgrading!!). 

Please ensure that the admin sets up a backup properly. Feel free to ask questions here or on the discussion platform. Feel free to reopen this issue if there is something left to discuss.

PR #1341, opened 2023-03-06
Create workflow for building and pushing docker images to GitHub registry

### LINKED ISSUE #1332: Docker use fails due to platform mismatch
**Describe the bug**
When running the docker image `asreview/asreview:latest`, it fails with

```
WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64/v3) and no specific platform was requested
exec /usr/local/bin/asreview: exec format error
```

Using `--platform linux/amd64/v3` also fails, with the error

```
Unable to find image 'asreview/asreview:latest' locally
latest: Pulling from asreview/asreview
Digest: sha256:85b2fe3ab3f121670d61bf8b6f977c912d4f8c45c975788ef021a5799426d806
Status: Image is up to date for asreview/asreview:latest
WARNING: image with reference asreview/asreview was found but does not match the specified platform: wanted linux/amd64/v3, actual: linux/arm64/v8
docker: Error response from daemon: image with reference asreview/asreview:latest was found but does not match the specified platform: wanted linux/amd64/v3, actual: linux/arm64/v8.
See 'docker run --help'.
```

Maybe it was built on Mac? Building the image locally and running it works.

**To Reproduce**
Run `docker run --platform linux/amd64/v3 asreview/asreview:latest`.

**Version information**
 - OS: Linux
 - ASReview version: NA

abelsiqueira:
Ref.: https://github.com/asreview/asreview/discussions/1333

Also closes #1332 

This PR introduces two things, separated into 2 commits, but let me know if you prefer two PRs.

The first change is that instead of using `pip install asreview` in the Dockerfile, this builds the project from the project folder. This is not strictly necessary, but it allows building from master and removes the dependency of needing to wait for deployment to PyPI to create the deployment to Docker.
The Dockerfiles have a two-stage build to try to keep them light. So the code to `asreview` is not available in the images.
I also changed from `python:3.8` to `python:3.8-slim`.

The second change is introducing a GitHub workflow to build and deploy the docker images to Docker Hub. The action uses a few `docker` actions to login, build a test version, query tags and labels metadata, build and push the final version.
The workflow runs for pull requests as well, but only for testing, not for pushing.
For the `master` branch, it pushes to `asreview/asreview:master`. For `v*` tags, it pushes to the expected tag and for `latest`.

There are two issues that I noticed but still don't have an answer:

- I can't export the results from the Docker image. I think this issue was already present, so I think that creating a new issue to investigate is the best approach. (It looks like it is related to permissions)
- The `asreview-cli` image uses `asreview-datatools` and `asreview-insights`. So if those are updated, we would need to build a new image.


abelsiqueira:
The error is pip timing out when downloading `scipy`. Could be just a lack of luck, can you re-run the workflow to see if that's the case?

J535D165:
Thanks @abelsiqueira! Great work. I'm running the failed test again [Update: it works]. 

> The first change is that instead of using pip install asreview in the Dockerfile, this builds the project from the project folder. This is not strictly necessary, but it allows building from master and removes the dependency of needing to wait for deployment to PyPI to create the deployment to Docker.
> The Dockerfiles have a two-stage build to try to keep them light. So the code to asreview is not available in the images.
> I also changed from python:3.8 to python:3.8-slim.

How do you deal with the react app? Nodejs is not installed, isn't it?

> The second change is introducing a GitHub workflow to build and deploy the docker images to Docker Hub. The action uses a few docker actions to login, build a test version, query tags and labels metadata, build and push the final version.
> The workflow runs for pull requests as well, but only for testing, not for pushing.
> For the master branch, it pushes to asreview/asreview:master. For v* tags, it pushes to the expected tag and for latest.

I like this. I will get the keys and add them to the repo envs. 

> I can't export the results from the Docker image. I think this issue was already present, so I think that creating a new issue to investigate is the best approach. (It looks like it is related to permissions)

I'm not aware of this, but I can imagine what is going wrong there. An issue would be nice. 

A final remark, I never completely understood whether it is actually needed to have two Docker images [external contribution]. I have seen tools where you can launch the command line with an extra flag. That would make more sense to me and make things easier to maintain. 

Again, very nice work!

abelsiqueira:
> Thanks @abelsiqueira! Great work. I'm running the failed test again [Update: it works].

Thanks, good to know.

> How do you deal with the react app? Nodejs is not installed, isn't it?

Only the `asreview-lab` builds it. I use `apt-get install -y npm`.

> I like this. I will get the keys and add them to the repo envs.

:+1: 

> > I can't export the results from the Docker image. I think this issue was already present, so I think that creating a new issue to investigate is the best approach. (It looks like it is related to permissions)
> 
> I'm not aware of this, but I can imagine what is going wrong there. An issue would be nice.
> 

I've created #1342 

> A final remark, I never completely understood whether it is actually needed to have two Docker images [external contribution]. I have seen tools where you can launch the command line with an extra flag. That would make more sense to me and make things easier to maintain.

I don't know much about the best practices, but I was thinking about that as well. If I understood it correctly, the CLI and the LAB have some orthogonality on the requirements (one needs the web components, and the other needs other pip packages). So far I would say it makes sense to keep them separated, but I still only have a basic understanding of what `asreview-cli` does.

jspaaks:
> I can't export the results from the Docker image.

I could be misunderstanding this, but normally I'd expect the result files inside the container to be written with the same permissions as whoever runs the docker (usually root). But that's OK, you can map the workdir from inside the container to outside the container using something like:
```
docker run --rm -v $PWD:/app imagename imagecommand
```

notify @abelsiqueira

J535D165:
> > I can't export the results from the Docker image.
> 
> I could be misunderstanding this, but normally I'd expect the result files inside the container to be written with the same permissions as whoever runs the docker (usually root). But that's OK, you can map the workdir from inside the container to outside the container using something like:
> 
> ```
> docker run --rm -v $PWD:/app imagename imagecommand
> ```
> 
> notify @abelsiqueira

This issue can be prevented if we find a solution for exporting the dataset without making a local copy. We faced an issue with the Excel export back in the days. Therefore we used the local copy and sent the file with flask. 

I also wonder why the upload works correctly and the export does not.

abelsiqueira:
I have changed the commit and rebased with the discussed ideas. Again, let me know if you prefer two PRs.

The first commit keeps a single Docker image which calls `asreview`. I had to create a `ASREVIEW_HOST` env variable to deal with the `lab --ip 0.0.0.0` situation. Also, since there is a single Dockerfile, I moved it to the root. I changed some documentation since the `docker/README.md` was removed, but since there was some repetition and some very specific things there, the removal is much larger than the inclusions. In particular, the "Add your extension to your image" was removed completely. Let me know if you want to keep more things (maybe these could be new issues).

The second commit creates the Docker workflow. I used it to push the `master`, `latest`, and `v1.2.0-beta` tags on my fork: https://hub.docker.com/r/abelsiqueira/asreview/tags. 

Regarding the exporting issue #error ex, it still fails even when mounting a volume. It looks related to #1250, but I haven't looked into details yet.
Perhaps it is possible to have end-to-end testing (e.g. [Playwright](https://playwright.dev) or [Cypress](https://www.cypress.io)) for the Docker image?

abelsiqueira:
PS. I've rebased again to fix linter issues.

abelsiqueira:
I was thinking how comprehensive the Docker image should be. It includes datatools and insights. Should it include makita as well? wordcloud?

abelsiqueira:
I will look into what is needed to finish this PR with the GitHub registry.

abelsiqueira:
It should be done now. Apparently, by using `GITHUB_TOKEN`, we don't need to define anything else and it justs works.
I tested it on my fork: https://github.com/abelsiqueira/asreview/pkgs/container/asreview

abelsiqueira:
@J535D165 is ready for review and merge

J535D165 (approved):

PR #1343, opened 2023-03-06
Start auth documentation, hidden collabo item, conversion script

cskaandorp:
I have written some initial documentation for authentication (README), the collabo menu-item in the profile-popper is finally gone (well, it's hidden), fixed a weird issue with test_asreview_database.py (destroyed the development database for testing purposes) and I have written a small script that may help converting an unauthenticated app into an authenticated one. The simple conversion workflow is described in the README.

I don't know if the README is the proper place for the documentation. Probably not. Feel free to do whatever is necessary to turn the text into proper documentation and let me know if I need to add stuff. 

The conversion script is added in the root folder of the application and is called auth_conversion.py. 

Again, let me know if you need more documentation.

God knows I have pulled flake8 several times over the repo, and isort, and black. And God also knows all my tests are still green.

### DIFF
--- a/auth_conversion.py
+++ b/auth_conversion.py
@@ -1,0 +1,110 @@
+import json
+import sqlite3
+from pathlib import Path
+from uuid import NAMESPACE_URL
+from uuid import uuid5
+
+ASREVIEW_PATH = Path("/Users/casperkaandorp/.asreview")
+DATABASE_NAME = "asreview.development.sqlite"
+
+
+def get_users(cursor):
+    """Select ids, names and emails from all users"""
+    query = "SELECT id, name, email FROM users"
+    cursor.execute(query)
+    return cursor.fetchall()
+
+
+def print_user_records(selection):
+    """Print user records from database."""
+    if not bool(selection):
+        return "No records."
+    print("ID\tname (email)")
+    print("==============================")
+    for id, name, email in selection:
+        print(f"{id}.\t{name} ({email})")
+
+
+def user_project_link_exists(cursor, folder_id):
+    """Check if a project record already exists."""
+    query = "SELECT COUNT(id) FROM projects " + \
+        f"WHERE project_id='{folder_id}'"
+    cursor.execute(query)
+    return cursor.fetchone()[0] == 1
+
+
+def link_user_to_project(conn, project_id, folder_name, user_id):
+    """Inserts project record, links user id to project"""
+    query = "INSERT INTO projects(project_id, folder, owner_id)" + \
+        "VALUES(?,?,?)"
+    cursor = conn.cursor()
+    cursor.execute(query, (project_id, folder_name, user_id))
+    conn.commit()
+    return cursor.lastrowid
+
+
+if __name__ == "__main__":
+    # establish connect with database
+    con = sqlite3.connect(ASREVIEW_PATH / DATABASE_NAME)
+    # get cursor
+    cursor = con.cursor()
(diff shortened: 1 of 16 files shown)


auth_conversion.py
```diff
@@ -0,0 +1,110 @@
+import json
+import sqlite3
+from pathlib import Path
+from uuid import NAMESPACE_URL
+from uuid import uuid5
+
+ASREVIEW_PATH = Path("/Users/casperkaandorp/.asreview")
```
PeterLombaers, line 7: Get the asreview path from asreview:

`[python]
from asreview import asreview_path

ASREVIEW_PATH = asreview_path()
`

auth_conversion.py
```diff
@@ -0,0 +1,110 @@
...
+DATABASE_NAME = "asreview.development.sqlite"
```
PeterLombaers, line 8: Probably when people are actually using this, it should be the production version. Then also the debug parameter in the config should be false.

auth_conversion.py
PeterLombaers: Or better yet, replace the occurences of `ASREVIEW_PATH` by `asreview_path()`

PeterLombaers:
The conversion script works like a charm!

Btw, here is what I used to create the users:

```[python]
import requests

url = 'http://localhost:5000/auth/signup'
data = {
    'password': 'wachtwoord1!',
    'email': 'peter@mail.com',
    'name': 'peter'
}
res = requests.post(url, data=data)
print(res.status_code)
print(res.text)
```
It's probably nice to have a little command line tool to add users. Though that would probably not use the rest api, but just directly access the user database.

PeterLombaers:
If I start the application with `AUTHENTICATION_ENABLE: false` in my config and an empty asreview folder, it works just fine. I can start a project and label some records. However, if I then return to the project overview I get the following:

```
ERROR:asreview.webapp.start_flask:Exception on /api/projects [GET]
Traceback (most recent call last):
  File "C:\Users\Admin\python\asreview\asreview\auth_venv\lib\site-packages\flask\app.py", line 2528, in wsgi_app
    response = self.full_dispatch_request()
  File "C:\Users\Admin\python\asreview\asreview\auth_venv\lib\site-packages\flask\app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "C:\Users\Admin\python\asreview\asreview\auth_venv\lib\site-packages\flask_cors\extension.py", line 165, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "C:\Users\Admin\python\asreview\asreview\auth_venv\lib\site-packages\fl    current_app.login_manager._load_user()
  File "C:\Users\Admin\python\asreview\asreview\auth_venv\lib\site-packages\flask_login\login_manager.py", line 347, in _load_user    raise Exception(
Exception: Missing user_loader or request_loader. Refer to http://flask-login.readthedocs.io/#how-it-works for more info.ERROR:root:Missing user_loader or request_loader. Refer to http://flask-login.readthedocs.io/#how-it-works for more info.
INFO:werkzeug:127.0.0.1 - - [16/Mar/2023 14:45:27] "GET /api/projects HTTP/1.1" 500 -
```

auth_conversion.py
cskaandorp: I have replaced all occurences of ASREVIEW_PATH with asreview_path

J535D165:
Thanks, merged

Issue #1348, opened 2023-03-20
Unexpected imports


jspaaks:
In [`/asreview/__init__.py`](https://github.com/asreview/asreview/blob/9b776ba928e1a2279867a0266d5c244afbe937da/asreview/__init__.py), 
1. it seems like the `__all__` import is unintentionally smaller than the list of regular imports. https://docs.python.org/3/tutorial/modules.html#importing-from-a-package
1. when I do an `import asreview` then `for item in dir(asreview): print(item)`, I get various things I don't want

Strictly speaking this affects the semantic versioning 'major' label, because resolving this issue probably involves removing (as in, not exporting anymore) functionality that used to be unintentionally accessible.

PR #1372, opened 2023-03-27
Add tests to safeguard against accidental changes to the library api

jspaaks:
**Changes proposed in this pull request**

This PR adds 2 tests in `tests/test_api_lib.py` to help us avoid accidentally growing or shrinking the API for ASReview-the-library. The first test takes a list of what we expect to see in the API and checks whether all of the expected items are present. The second test builds a list of what we actually have by recursing into the modules, and checks whether we have not exposed anything by accident.

Currently, the list of what we have is rather long (625 items). I hope that with these tests in hand, we can start work on cleaning up the API, getting rid of re-exporting system libraries and dependency libraries, getting rid of exporting the same object multiple times in different parts of the API, etc.

This PR also updates the `__init__.py` file for each module with the required imports to meet the current API. It further adds an `__all__` variable to help developer tooling such as code completion. The `__all__` is provided as a static list because that's what the tooling can deal with; however this makes `__all__`'s contents a bit brittle. Hence I've also added 2 asserts at the bottom of each `__init__.py` to make sure the contents of `__all__` remains consistent with what `dir()` returns.

Refs:

- #1348

~~*Add screenshots for proposed visual changes*~~

**Checklist**

- [x] Unit tests are added for new features and bug fixes
- [x] Documentation is added for new features
- [x] Title of the pull request is self-explaining

**How to check**

```shell
git checkout test-library-api
python3 -m venv venv
source venv/bin/activate
pip install flake8
pip install flake8-import-order
flake8 ./asreview ./tests  # should be quiet
pip install .[test]
python setup.py compile_assets
pytest
```


jspaaks:
The changes in this PR are too big to be manageable. After an in person discussion with @J535D165 we decided to go with just step by step adding `__all__` to the modules that need it, so tooling like code completion and sphinx can start using that. That means that this PR can be closed now.

PR #1379, opened 2023-03-28
Structure the package into cli, lib, and webapp parts

jspaaks:
**Describe the changes**

The `asreview` package consists of 
1. a library
2. a command line program,
3. a web application.

Both the command line program and the webapp use the library as its dependency; however, the current directory structure does not reflect this. As a matter of fact the structure suggests that the webapp and cli are dependencies/lower order modules of the `asreview` package. This makes the development experience more difficult than it needs to be. 

In this PR, I moved all the files and directories that together make up the `asreview` library to a new directory `lib`. Of course I then also had to change the imports and the entrypoint targets.

So you end up with 

```text
asreview
├── entry_points/
├── lib/
└── webapp/
```

which is a bit neater than what we had before:

```text
asreview/
├── compat.py
├── config.py
├── data/
├── datasets.py
├── entry_points/
├── exceptions.py
├── io/
├── models/
├── project.py
├── review/
├── search.py
├── settings.py
├── state/
├── types.py
├── utils.py
├── _version.py
└── webapp/
```

**How to check the library part**

```shell
git checkout cli-lib-webapp
python3 -m venv venv
source venv/bin/activate
pip install flake8
pip install flake8-import-order
flake8 ./asreview ./tests  # should be quiet
pip install .[test]
python setup.py compile_assets
pytest

python3
# try using the library, e.g.
import asreview
from asreview import lib as asreviewlib
dir(asreviewlib)
```

**How to check the cli and webapp parts**

```
asreview --help     # should show help
asreview algorithms # should show algorithms overview
asreview lab        # should start web app

# (fiddle with the web app, at least make a
# project <test-project> and add some data)

asreview state-inspect <test-project> results  # should show some data
```

**How to check building the docs**
```shell
pip install -r docs/requirements.txt
python3 -m sphinx -W --keep-going --color docs/source/ build/html/

# use a browser to navigate to `build/html/index.html`, check the API reference
```


jspaaks:
Awaiting merge of PR
- #1054

jspaaks:
After an in person discussion, it seems that instead of this PR, we'll go with 1 of the following options:
1. keep monolithic repo as it is (seems most likely)
2. split into separate repositories (one for the webapp, one for the CLI, and one for the library), to be published and versioned separately.

That means this PR can be closed.

PR #1386, opened 2023-03-30
Add regex line that removes start of url from doi pid

laurens88:


asreview/data/base.py
```diff
@@ -498,6 +498,9 @@ def duplicated(self, pid='doi'):
             s_dups_pid = ((s_pid.duplicated()) & (s_pid.notnull()))
         else:
             s_dups_pid = None
+           
+        if pid == 'doi':
+            s_pid = s_pid.str.replace("^https?://(www\.)?doi\.org/", "", regex=True)
```
J535D165, line 503: Very nice addition. For extra protection, I'd rather move this to
```python
        if pid in self.df.columns:
            # in case of strings, strip whitespaces and replace empty strings with None
            if is_string_dtype(self.df[pid]):
                s_pid = self.df[pid].str.strip().replace("", None)
                if pid == 'doi':
                    s_pid = s_pid.str.replace("^https?://(www\.)?doi\.org/", "", regex=True)
            else:
                s_pid = self.df[pid]
```

This prevents an error if the column "doi" contains something else than then you would expect.

J535D165 (changes requested):

J535D165:
Adding a unit test would also be nice, let me know if you need some help with that.

asreview/data/base.py
J535D165: Replacing `"^https?://(www\.)?doi\.org/"` by `r"^https?://(www\.)?doi\.org/"` resolves on of the lint errors.

asreview/data/base.py
```diff
@@ -63,6 +63,24 @@ def load_data(name, *args, **kwargs):
...
+            return urlopen(url).headers.get_filename(), url
+        except HTTPError as err:
+            # 308 (Permanent Redirect) not supported
+            # See https://bugs.python.org/issue40321
+            if err.code == 308:
+                return _get_filename_from_url(err.headers.get("Location"))
+            else:
+                raise err
+
+
```
J535D165, line 83:
```suggestion
```

J535D165 (changes requested):

J535D165 (approved):

J535D165:
Thanks! :)

PR #1399, opened 2023-04-12
tested auth conversion

cskaandorp:
Fixed decorator issues, formal test to convert unauthorised version to an authorised version and back to unauthorised (by using the conversion script). The auth_conversion script can only be executed in 'interactive' mode and takes the path of the database as a parameter. A small addition (json mapping as parameter) is forthcoming. Moved the documentation on authentication to DEVELOPMENT.md. Also: a number of tests now also test "no-permission so see/alter projects of another person" in the authenticated app. The old tests were too implicit and more focused on the forthcoming collaboration feature.


asreview/webapp/api/projects.py
```diff
@@ -114,17 +121,38 @@ def _get_authenticated_folder_id(project_id, user):
     if (project_from_db and user == project_from_db.owner) or (
         project_from_db is None
     ):
-        user_uuid = uuid5(NAMESPACE_URL, str(user.id)).hex
+        uuid = _get_project_uuid(project_id, user.id)
```
PeterLombaers, line 124: Could you change `uuid` to something like `project_uuid`? That way it doesn't conflict with the builtin `uuid` module.

asreview/webapp/api/projects.py
```diff
@@ -114,17 +121,38 @@ def _get_authenticated_folder_id(project_id, user):
...
+            # find the project
+            project = Project.query.filter(
+                Project.project_id == project_id
+            ).one_or_none()
+            if project is None:
+                return jsonify({"message": "project not found"}), 404
+            # if there is a project, check if
+            all_users = set([project.owner] + project.collaborators)
+            if current_user not in all_users:
+                return jsonify({"message": "no permission"}), 403
```
PeterLombaers, line 152: Nice, I like this. Much more explicit.

asreview/webapp/api/projects.py
cskaandorp: Done

asreview/webapp/api/projects.py
cskaandorp: Totally agree

PeterLombaers:
Thanks! This is another good step towards a beta release.

PR #1400, opened 2023-04-13
Add .vscode to .gitignore

PeterLombaers:
Adds the `.vscode` directory to `.gitignore`. It contains settings for the VSCode code editor so it should be ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,3 +160,6 @@ bower_components
 psd
 thumb
 sketch
+
+# VSCode
+.vscode


J535D165:
This is something you typically add to your global gitignore file. See https://www.mashupgarage.com/playbook/git/gitignore.html. This avoids a wide collection of editors in the gitignore :)

PeterLombaers:
I saw `.spyder` and `.sublime` so I thought this one would make sense, but your solution also works for me!

Issue #1411, opened 2023-04-20
Update dev dependencies with recent style additions?


jspaaks:
DEVELOPMENT.md was recently updated with dev dependencies for styling. It says to:

```shell
pip install black[jupyter] flake8 flake8-isort isort
```

Should we add this to the `dev` dependencies in setup.py?

https://github.com/asreview/asreview/blob/9b35dfbf42cc34e37dd4256f07d388daa2c80d3d/setup.py#L44-L52

(probably install `black` without the jupyter stuff )

J535D165:
Sounds good!

jspaaks:
PR was merged, closing

PR #1413, opened 2023-04-24
Updated development dependencies

jspaaks:
This PR adds the development dependencies black, flake8, flake8-isort, and isort to `extras_require` in `setup.py` and updates the developer documentation so they get installed.

Refs:

- #1411

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ DEPS = {
     "sbert": ["sentence_transformers"],
     "doc2vec": ["gensim"],
     "tensorflow": ["tensorflow~=2.0"],
-    "dev": ["check-manifest"],
+    "dev": ["black", "check-manifest", "flake8", "flake8-isort", "isort"],
     "test": ["coverage", "pytest"],
 }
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -45,7 +45,7 @@ def get_long_description():
     "sbert": ["sentence_transformers"],
     "doc2vec": ["gensim"],
     "tensorflow": ["tensorflow~=2.0"],
-    "dev": ["check-manifest"],
+    "dev": ["black", "check-manifest", "flake8", "flake8-isort", "isort"],
     "test": ["coverage", "pytest"],
 }
 DEPS["all"] = DEPS["sbert"] + DEPS["doc2vec"] + DEPS["dev"]
```
J535D165, line 51: I think it is better to remove `DEPS["dev"]` from `all` now.

J535D165 (changes requested):
Thanks!

setup.py
jspaaks: updated as requested

J535D165 (approved):

PR #1419, opened 2023-04-28
Add command line tool for user and project management in authenticated server

cskaandorp:
Extensive CLI tool to convert an unauthenticated app into an authenticated one. The tool can be executed as a subcommand of the asreview command. It can be used to add user accounts to the database, list existing projects and inserts converted projects (including ownership information) into the database as well. Both in bulk or interactively. Documentation is provided.


DEVELOPMENT.md
```diff
@@ -187,18 +185,66 @@ pare summarised below:
...
+
+```
+$ asreview auth-tool --help
+```
+
+#### Creating user accounts
+
+The first step is to create user accounts. This can be done interactively or by using a JSON string to bulk insert the accounts. To add user accounts interactively run the following command:
+```
+$ asreview auth-tool add-users --interactive --db-path ~/.asreview/asreview.development.sqlite
```
J535D165, line 207:
```suggestion
$ asreview auth-tool add-users --interactive --db-path ~/.asreview/asreview.production.sqlite
```

Who is the intended audience here? Developers or admins?

asreview/entry_points/auth_tool.py
```diff
@@ -0,0 +1,347 @@
...
+        except IntegrityError:
+            print(f"User with identifier {user.email} already exists")
+
+
+def rename_project_folder(project_id, new_project_id):
+    """Rename folder with an authenticated project id"""
+    folder = asreview_path() / project_id
+    folder.rename(asreview_path() / new_project_id)
+    # take care of the id inside the project.json file
+    with open(asreview_path() / new_project_id / "project.json", mode="r") as f:
```
PeterLombaers, line 137: Maybe we should put this in a try except block? I don't see why anything would go wrong during reading or writing of `project.json`, but if it does, you already renamed the folder. So something like
```
folder.rename(asreview_path() / new_project_id)
try:
    read_and_write
except Exception:
    folder.rename(asreview_path() / project_id)
```

asreview/entry_points/auth_tool.py
```diff
@@ -0,0 +1,347 @@
...
+        session.commit()
+
+
+def get_users(session):
+    return session.query(User).all()
+
+
+class AuthTool(BaseEntryPoint):
+    def execute(self, argv):
+        # from asreview.webapp.start_flask import main
```
PeterLombaers, line 179: Can be removed.

PeterLombaers:
I was trying to add three users in interactive mode. In the third one I made a typo. I did CTRL+C, but then all data was lost. I think it would make sense to add a user to the database every time you finish entering the data for one user, instead of after entering all the data. That way, if you make a typo, only the data of the last one is lost.

PeterLombaers:
Maybe interactive mode should be on by default? Then you don't have to do `-i` everytime, but only have to do `-j` if you want to use json (which will probably happen only once anyway).

DEVELOPMENT.md
cskaandorp: fixed

asreview/entry_points/auth_tool.py
cskaandorp: Done

asreview/entry_points/auth_tool.py
cskaandorp: Good call, fixed

J535D165:
Thanks, nice improvement

PR #1425, opened 2023-05-05
Introduce permanent project id and deprecate title-based project id

terrymyc:
This PR introduces a permanent `project_id` for each new project, replacing the current practice of changing the `project_id` when the project title is modified. This change will enhance the consistency and stability of project identification within the system.

Specifically, this PR includes the following updates:

- [x] Relocate decorator and helper functions from `api/projects` to their appropriate locations
- [x] Generate a unique, unchanging `project_id` upon project creation
- [x] Ensure exported dataset/project filenames contain a meaningful title
- [x] Deprecate the use of `name` and `project_name` in favor of `title` and `project_title` except in `project.json` as these are utilized in the frontend

Some related issues to discuss:

- Assign a new random project id to an imported project or not?


asreview/project.py
```diff
@@ -107,25 +111,47 @@ def get_projects(project_paths=None):
...
+    # Get the user uuid.
+    if app_is_authenticated(current_app):
+        user_uuid = _get_user_uuid(user.id)
+    else:
+        user_uuid = _get_user_uuid()
+
+    # Get the project uuid.
+    project_uuid = uuid5(NAMESPACE_URL, plaintext_project_id + user_uuid).hex
+
+    return project_uuid
```
J535D165, line 154: Can't we replace this by `uuid4().hex`? The title, namespace and username can be left out isn't it? That solves your issue with temporary projects.

J535D165 (commented):
Thanks a lot. 

I think we should keep the info in `project.json` identical, but we can rename the API from `name` into `project_title` or `title`.

J535D165:
> Assign a new random project id to an imported project or not?

I think yes. There is no need to store the `project_id` in the project itself anymore (we can do it, but not needed). Therefore, we can create a new project_id when importing.

J535D165:
> Prevent the title from being less than 3 characters

We can simplify this now to anything that's not an empty string.

asreview/project.py
```diff
@@ -107,25 +107,12 @@ def get_projects(project_paths=None):
     return [ASReviewProject(project_path) for project_path in project_paths]
 
 
-def _create_project_id(name):
-    """Create project id from input name."""
+def _get_project_uuid():
```
J535D165, line 110: This is an empty function, we can remove it and use `uuid4().hex` directly.

asreview/entry_points/simulate.py
```diff
@@ -148,7 +148,7 @@ def execute(self, argv):  # noqa
                 fp_tmp_simulation,
                 project_id=Path(args.state_file).stem,
                 project_mode="simulate",
-                project_name=Path(args.state_file).stem,
+                project_title=Path(args.state_file).stem,
```
J535D165, line 151: This is a breaking change in the documented API. We should make this backward compatible.

PeterLombaers:
I took the liberty to resolve the comments and merge master into this branch, so that we can speed up merging this pull request. @terrymyc Is there anything that should be added to this pull request, or do you consider it ready to merge?

PeterLombaers:
I was wondering one thing @terrymyc @J535D165 : In the simulate entrypoint there is a [distinction](https://github.com/terrymyc/asreview/blob/7e565d8c36e6c496a076802e8b84813abd0a2343/asreview/entry_points/simulate.py#L91) between what happens in the case of webapp and the case of cli. The part for webapp does not seem to work properly (the `as_data` variable is not always defined). So is this ever actually used from the webapp?

PeterLombaers:
Some of the test related to the auth converter are failing, but these tests are getting a rework as we speak so I'll merge this anyway. Thanks for the work and the great idea @terrymyc, this will make dealing with project identifiers a lot easier.

PR #1435, opened 2023-05-11
Store relevance scores

PeterLombaers:
Relevance scores of the last iteration of the model are now stored in state file.

The state already contained a table 'last_probabilities' which could store the relevance scores, but this was not actually used to store anything. The reason was that `QueryStrategy.query` only returned the ranking, and not the relevance scores underlying the ranking (assuming relevance scores are even used). 

I added a `return_classifier_scores` argument to `QueryStrategy.query`. By default this is set to `False` so that the default behaviour of the query strategy class is still the same as before. During review, the relevance scores are requested and stored in the state file. I also updated the documentation with an example of how to get the last ranking and relevance scores.

Some remarks:
- It might be a good idea to rename the table 'last_probabilities' to 'last_relevance_scores', but I don't think it is worth making any breaking changes over this. It should be possible to do this without breaking changes, since currently the table is unused, but this requires a bit more thought. @J535D165 any ideas?
- The reason that saving the relevance scores was somewhat complicated is that we have separate `QueryStrategy` and `Classifier` classes. This separation is a little bit arbitrary, they are both part of a model that uses a feature matrix to produce a ranking. When you are writing code in the `Review` classes, it would make more sense to have one `Model` class that gives you the ranking. Under the hood this model class could then have subcomponents `Model.classifier`, `Model.balance_strategy`, etc. This would make the review classes and entrypoints a lot simpler I think.


asreview/state/sqlstate.py
```diff
@@ -462,8 +462,16 @@ def add_last_probabilities(self, probabilities):
         Arguments
         ---------
         probabilities: list, np.array
-            List containing the probabilities for every record.
+            List containing the probabilities for every record. If this is None, the
```
Rensvandeschoot, line 465:
```suggestion
            List containing the relevance scores for every record. If this is None, the
```

J535D165 (approved):

J535D165:
Merci!!!

PR #1438, opened 2023-05-15
Star importing with __all__, nonrecursive variant

jspaaks:
Basically the approach is to list all modules that need to be part of the public interface in `docs/source/reference.rst`, and let `autosummary` generate the documentation for any objects it finds in the module's `__all__` attribute. This approach does not require renaming any files.

During generation of the documentation, Sphinx complains about various things such as
- WARNING: more than one target found
- WARNING: undefined label
- errors that seem to result from trying to document code that isn't ours

I expect this will get better once we add `__all__` to each module listed in `reference.rst`.

As a test, I also added an `__all__` in `datasets.py`, that seems to work because it no longer lists its own imports like `abstractmethod` and `ABC`.

To test it out, do the following

```shell
deactivate
git checkout star-importing-with-all-nonrecursive
python3 -m venv venv
source venv/bin/activate
pip install -r docs/requirements.txt
cd docs
make html
# point browser to build/html/index.html:
firefox "$(pwd)/build/html/index.html"
# (inspect "API reference" from the bottom of the left column)
```

The result should look somewhat like this:

image


jspaaks:
Next steps if @J535D165 agrees with this approach are:

1. add classes, functions, etc to `__all__` thus avoiding re-exporting of Python's own objects, or of imported objects such as those from pandas or Numpy
2. decide on how to deal with re-exported objects, e.g. do we use only `asreview.data.ASReviewData`, only `asreview.ASReviewData`, or both.

J535D165:
> Next steps if @J535D165 agrees with this approach are:
> 
> 1. add classes, functions, etc to `__all__` thus avoiding re-exporting of Python's own objects, or of imported objects such as those from pandas or Numpy
> 2. decide on how to deal with re-exported objects, e.g. do we use only `asreview.data.ASReviewData`, only `asreview.ASReviewData`, or both.

Lets go.

jspaaks:
Explored https://github.com/asreview/asreview/pull/1438#issuecomment-1549564953 some more, so we can see what it looks like.

Here is a screenshot of the api reference documentation at the top level:

image


and here, e.g. for `asreview.models.classifiers`:

image

and an individual class (`asreview.models.classifiers.NaiveBayesClassifier`):

image

asreview/entry_points/__init__.py
```diff
@@ -16,6 +16,7 @@
 from asreview.entry_points.auth_tool import AuthTool
 from asreview.entry_points.base import BaseEntryPoint
 from asreview.entry_points.lab import LABEntryPoint
+from asreview.entry_points.lab import WebRunModelEntryPoint
```
J535D165: I think we consider this a private entry point. Maybe we can make this explicit.

PR #1439, opened 2023-05-16
Add progress bar for sbert feature extraction

Robdboer:
Add progress bar


PeterLombaers (approved):

PeterLombaers:
Looks good to me!

PeterLombaers:
Thanks @Robdboer !

jteijema:
For future reference, we should maybe incorporate the used transformer in the waiting message, instead of using "sbert". It would be nice to show users which transformer model theyre using. @PeterLombaers

PR #1440, opened 2023-05-16
Allow SBERT to load other transformer models

rohitgarud:
Models which are not specifically trained sentence transformer models such as SciBERT, BioBERT or longformer can be loaded from Huggingface

### DIFF
--- a/asreview/models/feature_extraction/sbert.py
+++ b/asreview/models/feature_extraction/sbert.py
@@ -66,7 +66,8 @@ class SBERT(BaseFeatureExtraction):
     pooling_mode: str, optional
         Polling mode to get sentece embeddings from word embeddings
         Default: 'mean'
-        Other options available are mean, max and cls. Only used if is_pretrained_SBERT=False
+        Other options available are mean, max and cls.
+        Only used if is_pretrained_SBERT=False
         mean: Uses mean pooling of word embeddings
         max: Uses max pooling of word embeddings
         cls: Uses embeddings of [CLS] token as sentence embeddings
--- a/asreview/models/feature_extraction/sbert.py
+++ b/asreview/models/feature_extraction/sbert.py
@@ -106,5 +106,5 @@ class SBERT(BaseFeatureExtraction):
             model = SentenceTransformer(
                 modules=[word_embedding_model, pooling_layer]
             )
-        X = np.array(model.encode(texts))
+        X = model.encode(texts)
         return X
--- a/asreview/models/feature_extraction/sbert.py
+++ b/asreview/models/feature_extraction/sbert.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import numpy as np
-
 try:
     from sentence_transformers import models
     from sentence_transformers.SentenceTransformer import SentenceTransformer
--- a/asreview/models/feature_extraction/sbert.py
+++ b/asreview/models/feature_extraction/sbert.py
@@ -16,6 +16,7 @@ import numpy as np
 
 try:
     from sentence_transformers.SentenceTransformer import SentenceTransformer
+    from sentence_transformers import models
 except ImportError:
     ST_AVAILABLE = False
 else:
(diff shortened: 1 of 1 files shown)


asreview/models/feature_extraction/sbert.py
```diff
@@ -60,19 +61,50 @@ class SBERT(BaseFeatureExtraction):
     transformer_model : str, optional
         The transformer model to use.
         Default: 'all-mpnet-base-v2'
-
+    is_pretrained_SBERT: boolian, optional
```
PeterLombaers, line 64:
```
    is_pretrained_SBERT: boolean, optional
        Default: True
    pooling_mode: str, optional
        Pooling mode to get sentence embeddings from word embeddings
        Default: 'mean'
        Available options are mean, max and cls.
        Only used if is_pretrained_SBERT=False
        mean: Uses mean pooling of word embeddings
        max: Uses max pooling of word embeddings
        cls: Uses embeddings of [CLS] token as sentence embeddings
```

PeterLombaers (approved):
Nice! This looks like a useful addition.

asreview/models/feature_extraction/sbert.py
```diff
@@ -60,19 +61,50 @@ class SBERT(BaseFeatureExtraction):
...
+            # word embeddings
+            word_embedding_model = models.Transformer(self.transformer_model)
+            pooling_layer = models.Pooling(
+                word_embedding_model.get_word_embedding_dimension(),
+                pooling_mode=self.pooling_mode,
+            )
+            model = SentenceTransformer(
+                modules=[word_embedding_model, pooling_layer]
+            )
         X = np.array(model.encode(texts))
```
jteijema, line 109: Is this casting still needed?

jteijema (approved):

asreview/models/feature_extraction/sbert.py
rohitgarud: I didn't get what you mean by casting here.  Do you mean the keyword argument modules?

jteijema: Sorry, selected wrong line. I am referencing line 109. The `model.encode(texts)` is cast to an array, but I think this is not necessary.

rohitgarud: You are right. `model.encode(texts)` returns a numpy array and casting is unnecessary.

rohitgarud: Adding a progress bar will be definitely helpful, especially for extractors taking a long time. But I think rather than using SBERT's progress bar, a reusable "progress bar creator" should be developed which can be used for any other feature extractors such as Doc2vec and fasttextsum

asreview/models/feature_extraction/sbert.py
```diff
@@ -60,19 +61,50 @@ class SBERT(BaseFeatureExtraction):
...
 
     name = "sbert"
     label = "Sentence BERT"
 
-    def __init__(self, *args, transformer_model="all-mpnet-base-v2", **kwargs):
+    def __init__(
+        self,
+        *args,
+        transformer_model="all-mpnet-base-v2",
+        is_pretrained_SBERT=True,
```
J535D165, line 83:
```suggestion
        is_pretrained_sbert=True,
```

PeterLombaers:
I merged #1439. Could you update your branch to include those changes? And could you remove the type casting from line 109, so it just becomes `X = np.array(model.encode(texts))`? Then I think this is ready to merge.

rohitgarud:
Thank you @PeterLombaers, was about to push after removing the import.. will check the linter before committing from next time :)

asreview/models/feature_extraction/sbert.py
J535D165: Great suggestions. Would be nice to work on this (in another PR)!

rohitgarud: Maybe something which can be accessed by the front-end so that we can display the progress bar on Webapp as well

PeterLombaers:
Thanks you @rohitgarud !

PR #1462, opened 2023-06-01
Remove dill as dependency

J535D165:


J535D165:
I'm surprised this dependency is still in there. I think it is not used anywhere.

PR #1463, opened 2023-06-02
Replace custom training lock by filelock

J535D165:
This PR proposes to remove the custom training lock by a similar lock from the filelock package. There are no changes in functionality, but it is expected that the new lock is more reliable and future proof.


PeterLombaers:
Looks good to me. I can only ever get one model to train, no matter how fast I label!

PeterLombaers (approved):

PR #1465, opened 2023-06-07
Replace deprecated pkg_resources by importlib.metadata

### LINKED ISSUE #1394: pkg_resources is deprecated as an API
It's time to update pkg_resources as it is deprecated. 

```
=================================================== warnings summary ===================================================
../../../.pyenv/versions/3.9.7/envs/asreview-dev/lib/python3.9/site-packages/pkg_resources/__init__.py:121
  /Users/Bruin056/.pyenv/versions/3.9.7/envs/asreview-dev/lib/python3.9/site-packages/pkg_resources/__init__.py:121: DeprecationWarning: pkg_resources is deprecated as an API
    warnings.warn("pkg_resources is deprecated as an API", DeprecationWarning)
```

J535D165:
Closes #1394 

This PR deprecates two functions:

- `asreview.utils.get_entry_points` This function will be available in the entire v2 series, but will be deprecated because of limited added value compared to `importlib.metadata.entry_points`. Basically, this function from the `importlib` package is almost identical to ours. 
- `asreview.utils.get_model_names` This function never worked without changing the default argument. This indicates limited use and is most likely a bug. This function can be replaced with `importlib.metadata.entry_points` and other list functions in the submodules of `asreview.models.{COMPONENT}`. 

Breaking API changes: 

- All `list_{}` and `get_{}_class` are slightly effected by the changes introduced by the migration. All these functions behave similar except from exceptions they raise. It turned out that `KeyErrors` and `ImportErrors` were incorrectly propagated as `ValueErrors`. We consider this a bug as they are not ValueErrors. Therefore, we propose to use the default behavior of the `importlib.metadata` module instead of the ValueErrors. This simplifies our internals as well as better expectations for the user.


PeterLombaers:
Seems good to me to make the change to `importlib`. I agree that the breaking changes are basically bugfixes. 

Regarding deprecating `asreview.utils.get_entry_points`: it indeed doesn't do much different from `_entry_points(group='asreview.entry_points')`. The only advantage I see to having a dedicated function for this within `asreview` is that you could add a docstring explaining the use of the function, and maybe the entrypoint system in general. The way this part of the code works was always quite magical until I started reading about it an hour ago :)

Other than that, all seems fine to me!

J535D165:
> Seems good to me to make the change to `importlib`. I agree that the breaking changes are basically bugfixes.
> 
> Regarding deprecating `asreview.utils.get_entry_points`: it indeed doesn't do much different from `_entry_points(group='asreview.entry_points')`. The only advantage I see to having a dedicated function for this within `asreview` is that you could add a docstring explaining the use of the function, and maybe the entrypoint system in general. The way this part of the code works was always quite magical until I started reading about it an hour ago :)
> 
> Other than that, all seems fine to me!

Good comment. Maybe some documentation can help here. Currently, none of these functions are documented in our API reference. An extension of https://asreview.readthedocs.io/en/latest/extensions_dev.html would be nice. I will work on this.

PR #1472, opened 2023-06-19
Make tests and setup.py Windows OS compatible

cskaandorp:
Some tests are failing on Windows OS. Here's a possible solution for that problem.

### DIFF
--- a/.github/workflows/ci-frontend.yml
+++ b/.github/workflows/ci-frontend.yml
@@ -2,7 +2,10 @@ name: test-frontend
 on: [push, pull_request]
 jobs:
   compile-and-test:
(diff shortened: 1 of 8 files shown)


J535D165:
Maybe good to setup a matrix for the different platforms https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

cskaandorp:
If all tests are green on Windows please do not merge yet. I'd like to try the OS matrix.

PeterLombaers:
@J535D165 This also fixes a bug in `setup.py` where Windows users could not use `setup.py compile_assets`.

J535D165:
Superb

PR #1474, opened 2023-06-22
Allow relative path to flask configfile

PeterLombaers:
Currently, when starting the app through the command line with a flask configfile you need to give the absolute path to the file. This pull request changes that to a relative path, which should be more intiutive to use.

### DIFF
--- a/asreview/webapp/start_flask.py
+++ b/asreview/webapp/start_flask.py
@@ -18,6 +18,7 @@ import logging
 import os
 import socket
 import webbrowser
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


J535D165:
Shall we update the examples in the docs?

J535D165:
I like this btw :)

PeterLombaers:
I was looking at the docs. It seems the only example in the docs with a flask configfile left it open whether the path should be absolute or relative, so it seems we don't need to change anything there.

PR #1475, opened 2023-06-22
CORS issue, OAuth

cskaandorp:
The goal of this pull request is to fix the OAuth issues in production. While tracing an OAuth roundtrip I stumbled upon a CORS issue. Auth wasn't simply working. Perhaps because of a trailing slash in an origin url. After correcting this all was fine again. The first commit represents this correction, this makes all Cors inits uniform. Manual OAuth testing is forthcoming.

### DIFF
--- a/asreview/webapp/api/auth.py
+++ b/asreview/webapp/api/auth.py
@@ -44,12 +44,19 @@ bp = Blueprint("auth", __name__, url_prefix="/auth")
 
 # NOTE: not too sure about this, what if we are dealing with a
 # domain name
-ROOT_URL = "http://127.0.0.1:3000/"
+ROOT_URL = "http://127.0.0.1:3000"
 
 CORS(
     bp,
-    resources={r"*": {"origins": ["http://localhost:3000", ROOT_URL]}},
-    supports_credentials=True,
+    resources={
+        r"*": {
+            "origins": [
+                ROOT_URL,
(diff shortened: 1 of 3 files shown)


cskaandorp:
Still slightly confused about whether this pull request really solved the OAuth issue.
I have build the app and ran it on the default port 5001.
I will issue a new production flask_config file that will work with Github and Orcid. One thing that bothers me is that the app runs on http://localhost:5001 and not on 127.0.0.1 (I know it's the same thing). The problem is that OAuth providers want a precise URL. When I enter 127.0.0.1 as a callback url after authorization has been done, it won't work. Anyhow: 

Merge this PR and I think the app is ready for an alpha release
I will provide a flask_config file, this file matches OAuth services that expect a requesting URL http://localhost:5001

If the test app runs on a different URL we need to reconfig services based on the URL that is used for the test app.

J535D165:
> I will issue a new production flask_config file that will work with Github and Orcid. One thing that bothers me is that the app runs on http://localhost:5001/ and not on 127.0.0.1 (I know it's the same thing). The problem is that OAuth providers want a precise URL. When I enter 127.0.0.1 as a callback url after authorization has been done, it won't work. Anyhow:

I think this is a common issue with developing OAuth solutions local hosts. I'm not to much worried about this atm. 

Thanks for the fix.

PR #1476, opened 2023-06-22
Update about.rst - minimum python version should be 3.8

michelleliu:
according to https://github.com/asreview/asreview/pull/1457


J535D165:
Great!

PR #1478, opened 2023-06-26
Potential fixes for cloud related URL and port issues

cskaandorp:
Added .env in webapp: this enhances the way global.js refers to the back-end. Instead of hard-coding the back-end's url it is now configured. Change this by either  editing the .env file or override by creating a .env.production or .env.local file.

IP address and Port can now be configured in the  flask_config.json file as well and are stored in the config of the Flask app object. Besides that you can also put the CORS origins in the config file. These URLs are also stored in app.config. I have removed all other origins from the Blueprints.

What remained were the forgot-password and confirmation emails. In those emails I need a link towards the front-end. But what is the URL of the front-end? For now I use the Origin that can be found in the request header. That is fine for the forgot-password email (not tested). If the account is created from the front-end we will be ok (not tested yet) as well. This also works with multiple front-ends (not sure if we want that). 
But! If these accounts are created by an admin without using the front-end, how is the back-end suppose to know what the URL of the front-end is...

I have also added an after-request function which injects the 'Access-Control-Allow-Credentials' key in the header and sets it to True since I keep on having issues with it (this was taken care of in the Blueprint CORS (projects.py, team.py, etc)). Now it works always. But if feels like taking a stap back. Again: now it is very stable. I'd like to discuss this!

Last but not least: I have increased the sleeping time of test_status_project in test_projects to see if this test is more "deterministic" under Windows.


asreview/webapp/start_flask.py
```diff
@@ -253,6 +259,10 @@ def create_app(**kwargs):
     if config_file_path != "":
         app.config.from_file(Path(config_file_path).absolute(), load=json.load)
 
+    # Make sure we have front end origins
+    if not app.config.get("FRONT_END_ORIGINS", False):
+        app.config["FRONT_END_ORIGINS"] = FRONT_END_ORIGINS
```
J535D165, line 264: I'm still a bit confused about this. In production mode, the front end is served on the same port as the backend. How does this work then?

asreview/webapp/.env
```diff
@@ -0,0 +1 @@
+REACT_APP_API_URL=http://localhost:5000/
```
J535D165, line 1: Do you want to commit this (as you seem to ignore it above)?

J535D165: I think I understand after carefully reading your post.

cskaandorp: Yeah, this is the default. But somehow my port 5000 has been taken, so the app takes 5001. To overcome this I create a .env.local file which overrides .env with the correct port.

cskaandorp:
I think I have reached the final stage of this pull request.

* I have deprecated the --ip argument. It is ignored. I could use existing code that mentions that the deprecated arguments are ignored. Just riding that wave..
* I have added comments that notify developers why and when you have to add the ALLOWED_ORIGINS in the config.
* I have removed the default for ALLOWED_ORIGINS, you need them in a very specific situation (dev on front end). By removing them we won't be bothered with the allowed_origins in the response header from the back end when they are not necessary. 
* We prefer localhost above 127.0.0.1
* In the front end a redirect is placed that redirects to localhost -if- the end-user requests 127.0.0.1
* Small refactor of React signin and signup forms to response to the Enter key once the (last) password field has been filled out.

Please run this branch for a quick spin before merging (if you are OK with the changes)

Issue #1481, opened 2023-06-29
ImportError: cannot import name 'metadata' from 'importlib'


acmdorio:
**Describe the bug**
ImportError: cannot import name 'metadata' from 'importlib'

**To Reproduce**
Steps to reproduce the problem:
- just install ASReview and try to run it (w/ Python<3.8).

**Version information**
 - OS: MacOS 13
 - ASReview version 1.2

**Additional context**

J535D165:
Thanks for reporting. You are right. We will bump the minimum version to 3.7 in the next release (also because the version is EOL https://devguide.python.org/versions/). However, the current version should be fixed, so I thank you for your PR!

Duplicate #1418

acmdorio:
Nice! My pleasure to collaborate.
BR,
*
--
ANDRÉ

J535D165:
Fixed in https://github.com/asreview/asreview/releases/tag/v1.2.1 by bumping the minimum version of Python to 3.8.

PR #1482, opened 2023-06-29
Issue #1481 - Update __main__.py

acmdorio:
Please, consider this suggestion to avoid "ImportError: cannot import name 'metadata' from 'importlib'" for Python<3.8. It works for me. Congrats on this awesome work!!!

### DIFF
--- a/asreview/__main__.py
+++ b/asreview/__main__.py
@@ -15,7 +15,12 @@
 """Command Line Interface (CLI) for ASReview project."""
 import argparse
 import sys
-from importlib.metadata import metadata
+try:
+    from importlib import metadata
+except ImportError: # for Python<3.8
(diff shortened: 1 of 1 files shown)


J535D165:
Dear @acmdorio, thanks for the PR and the proposed implementation. As we planned to bump the minimum version of Python because of the EOL of 3.7, we decided to drop the support for 3.7 in 1.2.1. This is not our nicest solution, but we consider this a bug (that we didn't bump the minimum version in release 1.2). I think it's best to close this PR now and let users upgrade Python. I thank you in the release notes of version 1.2.1 https://github.com/asreview/asreview/releases/tag/v1.2.1. I hope you are fine with this solution. 

All the best, Jonathan

Issue #1494, opened 2023-07-17
Tensorflow 2.13.0 removes KerasClassifier wrapper


jteijema:
https://github.com/tensorflow/tensorflow/releases

TF 2.13.0 removed the Keras scikit-learn API wrappers (KerasClassifier and KerasRegressor), which had been deprecated in August 2021. We recommend using SciKeras instead.

This is a breaking change for our NN classifier.

Options right now:
- try to get it to work without a wrapper
- pin Tensorflow to 2.12.x
- use [adriangb/scikeras](https://github.com/adriangb/scikeras )

J535D165:
Thanks for the issue. Option 1 is the best and looks straightforward. Would you like to help me?

J535D165:
Users running into this problem can downgrade tensorflow. Fix in the next release.

PR #1495, opened 2023-07-17
Update nn_2_layer.py

jteijema:
Following #1494, this PR updates the neural network model. Model was tested and performs as before.


jteijema:
image

Interface works with the prediction output, and performance is good.

J535D165:
Very nice! I'm looking forward to the updated LSTM models :)

J535D165 (approved):

PR #1499, opened 2023-07-19
Change flask_configfile to TOML

cskaandorp:
This PR enables users to config the ASReview webapp with a TOML file. Please note: Flask 2.3.0 and up is needed. Also note that toml can be parsed in Python 3.11 with a std lib called tomllib. Versions below 3.11 must use its predecessor tomli, hence the try/except import in start_flask. On my end flake8 is satisfied.


setup.py
```diff
@@ -63,6 +63,7 @@ def get_long_description():
     "tqdm",
     "gevent>=20",
     "datahugger>=0.2",
+    "tomli",  # included in Python 3.11 as tomllib
```
J535D165, line 66: See the syntax below for importlib_metadata.

J535D165 (changes requested):
Nice!

setup.py
cskaandorp: Done.

J535D165 (approved):

J535D165:
Thanks!!

setup.py
J535D165: But where?

PR #1502, opened 2023-07-20
Force the random number generator to be seeded

PeterLombaers:
This pull request tries to improve how we deal with random number generation and random seeds.

At the moment, we use a random number generator, and optionally you can choose to set a seed. I propose to make the random seed obligatory. This will make simulations or systematic reviews that use ASReview more reproducible.

The way I force random seeds to be used is by introducing an new `SeededRandomState` class, and a constructor function `get_random_state`. The `SeededRandomState` class is exactly the same as `np.random.RandomState`, but has the added attribute `seed`, which is always set. The way to use this class in the code is by using the `get_random_state` function, which will make sure that the random seed is always set correctly. This follows the pattern from numpy where they have the `np.random.Generator` class and a constructor function `np.random.default_rng()` which is the preferred way to instantiate the class.

Remarks:
- I use the legacy class `np.random.RandomState`, and not the newer `np.random.Generator` class. The reason is that with the newer class results of simulations would change, even when the random seed is the same. The downside is that this legacy class is frozen and will not be improved anymore by numpy. So if in the future someone needs random number generating capabilities not in `np.random.RandomState` then we would need to switch to the newer class. I have some comments in the code about how it should look you we want to use the newer class.
- I have not yet the new random seed to the state file, so this is not yet stored. The main reason is that I think some more changes are needed to the settings object before we can do this properly. I think it's useful to make the change in this pull request as a first step.
- There are always some caveats with how reproducible random number generation is. See also [here](https://numpy.org/doc/stable/reference/random/compatibility.html). In general, I don't think we use very complicated algorithms, so I would expect that in our case this won't really matter.
- This pull request is created by extracting the useful parts from #1424

### DIFF
--- a/asreview/utils.py
+++ b/asreview/utils.py
@@ -14,7 +14,6 @@
 
 import functools
 import os
-import secrets
 import sys
 import warnings
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


asreview/utils.py
```diff
@@ -270,21 +270,64 @@ def is_url(url):
...
+    -------
+    SeededRandomState
+        A random number generator, seeded by the provided seed.
+    """
+    if isinstance(seed, SeededRandomState):
+        return seed
+    # # For the newer np.random.Generator class, the seed setting would be as follows:
+    # # https://numpy.org/doc/stable/reference/random/index.html#quick-start
+    # if seed is None:
+    #     seed = secrets.randbits(128)
```
J535D165, line 297: Let's bump numpy and use this.

J535D165 (commented):
Ready to merge in v2 branch.

PeterLombaers:
I tried to use the newer `np.random.Generator` class but stumbled onto a problem: https://github.com/scikit-learn/scikit-learn/issues/16988
Scikitlearn doesn't actually support the new RNG yet. It's probably smarter to wait until they support it before making the change. So I would propose to merge this pull requests as is.

J535D165:
Thanks for the update! Let's keep it like this then.

PR #1517, opened 2023-08-22
oauth/auth-cookie fixes

cskaandorp:
Fixing the 404 after oauth and the cookie problem if you transfer from auth to non-auth

### DIFF
--- a/asreview/webapp/start_flask.py
+++ b/asreview/webapp/start_flask.py
@@ -299,6 +299,11 @@ def create_app(**kwargs):
     login_manager.session_protection = "strong"
 
     if app.config["AUTHENTICATION_ENABLED"] is False:
+        # This ensures the app handles the anonymous user
+        # when authentication is disabled and there is no
+        # configuration file
+        app.config["SECRET_KEY"] = "no_secret"
+
         # This is necessary to pass the test_webapp.py tests
         @login_manager.user_loader
         def load_user(user_id):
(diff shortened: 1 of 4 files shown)


cskaandorp:
I think I understand the 404 after oauth in production: it is caused by the callback being picked up by the backend (since backend and frontend run on the same port) while the React Router should pick it up. I think the solution is simple: render index.html so React routing can pick up the callback. I have tested it, and it seems to be OK now. Tomorrow the cookie-bug after transitioning from auth to non-auth.

cskaandorp:
Please try this out as well. Dealing with several browsers, removing cookies, restarting the app a million times... it invites to overlook things. I am pretty sure all mentioned issues (the 404 after oauth, and the error when moving from auth to unauth without logging out) are solved now. But assumption is the mother of all... (I guess you are too young to have seen that film (https://youtu.be/wg4trPZFUwc?si=HwNRH20nXcYNp6AT)).

cskaandorp:
PLEASE DONT MERGE YET, sorry, forgot something @PeterLombaers and me discovered today, will fix that tomorrow

cskaandorp:
Ready, please merge when code is accepted

asreview/webapp/start_flask.py
```diff
@@ -299,6 +299,11 @@ def create_app(**kwargs):
     login_manager.session_protection = "strong"
 
     if app.config["AUTHENTICATION_ENABLED"] is False:
+        # This ensures the app handles the anonymous user
+        # when authentication is disabled and there is no
+        # configuration file
+        app.config["SECRET_KEY"] = "no_secret"
```
J535D165, line 305:
```suggestion
        app.config["SECRET_KEY"] = None
```

Does this also work?

J535D165 (commented):
Looks good!! Thanks

asreview/webapp/start_flask.py
cskaandorp: None doesn't work, empty string does. And I think the latter looks better than "no_secret".

PR #1518, opened 2023-08-28
Docker recipe for authenticated/verified app

cskaandorp:
Starting a new pull request for the NGINX/Gunicorn Docker recipe.


cskaandorp:
Having issues with multiple Gunicorn workers who want to create the same database (or so it seems). Added documentation.

Docker/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Building the authenticated, verified version
+
+If you would like to setup the ASReview application as a shared service, a more complicated container setup is required. A common, robust setup for a Flask/React application is to use [NGINX](https://www.nginx.com/) to serve the frontend, and [Gunicorn](https://gunicorn.org/) to serve the backend. We build separate containers for both front- and backend with [docker-compose](https://docs.docker.com/compose/).
+
+For account verification, but also for the forgot-password feature, an email server is required. But maintaining an email server can be very demanding. If you would like to avoid it, a third-party service like might be [SendGrid](https://sendgrid.com/) an good alternative. In this recipe we use the SMTP Relay Service from Sendgrid: every email send by the ASReview application will be relayed by this service. Sendgrid is for free if you don't expect the application to send more than 100 emails per day. Receiving reply emails from end-users is not possible if you use the Relay service, but that might be irrelevant.
```
PeterLombaers, line 22:
```suggestion
For account verification, but also for the forgot-password feature, an email server is required. But maintaining an email server can be very demanding. If you would like to avoid it, a third-party service like [SendGrid](https://sendgrid.com/) might be a good alternative. In this recipe we use the SMTP Relay Service from Sendgrid: every email send by the ASReview application will be relayed by this service. Sendgrid is for free if you don't expect the application to send more than 100 emails per day. Receiving reply emails from end-users is not possible if you use the Relay service, but that might be irrelevant.
```

Docker/auth_verified/Dockerfile_backend
```diff
@@ -0,0 +1,39 @@
...
+
+RUN apt-get update \
+    && apt-get install -y git build-essential\
+    && pip3 install --upgrade pip setuptools \
+    && pip3 install --user gunicorn \
+    && pip3 install --user . \
+    && pip3 install --user asreview-datatools asreview-insights asreview-makita asreview-wordcloud
+
+# Second stage
+FROM python:3.8-slim
```
PeterLombaers, line 22: Same comment as before.

PeterLombaers:
Maybe it's a good idea to put the configuration of the ports in a `.env` and load them from there in the dockerfiles? That makes configuring them a bit easier.

Docker/auth_verified/Dockerfile_backend
```diff
@@ -0,0 +1,39 @@
+# First stage
+FROM python:3.8-slim AS builder
```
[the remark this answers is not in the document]
cskaandorp: Done, now it's 3.11

Docker/auth_verified/Dockerfile_backend
cskaandorp: Done

cskaandorp:
Trying to use the .env file enhancement, and it seems I will get it to work. Many small hickups when it comes to juggling environment vars to entrypoints and whatnot. I think I am closing in.

Also: found a couple of things that did not work properly in the frontend!!!!

cskaandorp:
> Maybe it's a good idea to put the configuration of the ports in a `.env` and load them from there in the dockerfiles? That makes configuring them a bit easier.

Indeed it was, thanks @PeterLombaers. Although I had to add yet another config file, by now the only thing people have to do is editing the .env file if they want to and creating a SendGrid account etc.

The current verskion is working. But! I still get the feeling that multiple processes (sometimes) try to create database tables simultaneously. Therefor I will remove the database creation from start flask in the coming commits.

cskaandorp:
The PR is ready for inspection!

Docker/simple/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# First stage
+FROM python:3.8-slim AS builder
```
PeterLombaers, line 2: This should be 3.11 as well.

Docker/simple/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+COPY . /app
+RUN apt-get update \
+    && apt-get install -y git npm \
+    && pip3 install --upgrade pip setuptools \
+    && python3 setup.py compile_assets \
+    && pip3 install --user . \
+    && pip3 install --user asreview-datatools asreview-insights asreview-makita asreview-wordcloud
+
+# Second stage
+FROM python:3.8-slim
```
PeterLombaers, line 16: And here as well.

PeterLombaers:
In the simple version, if I want to export the project data out of the container, how do I do it? Should run the container with a certain volume mount? Or should I use another tool to get the data out of the container? It would be nice to have a short explanation in the readme.

PeterLombaers:
When I tried to run the simple version, first I got the error `pg_config executable not found`. When I added the line `&& apt-get install -y git build-essential libpq-dev\` to the docker file the build was succesful. When I try to run the container and go to localhost:8080 I get stuck on a loading animation. Not sure what the problem is, I can't seem to get the logs of the asreview process.

J535D165 (approved):
Thanks very nice!

PR #1528, opened 2023-09-18
Add integration tests with selenium

cskaandorp:
Using selenium to conduct integration test


.gitignore
```diff
@@ -167,3 +167,4 @@ sketch
 notes.md
 
 flask_config.json
+flask_config.toml
```
J535D165, line 170:
```suggestion
flask_config.*
```

.github/workflows/ci-frontend.yml
```diff
@@ -28,6 +28,6 @@ jobs:
           pip install pytest
           pip install pytest-random-order
           pip install --no-cache-dir .
-      - name: Test flask app
+      - name: Test flask web app
         run: |
-          pytest --random-order asreview/webapp/tests
+          pytest --random-order --ignore=asreview/webapp/tests/integration_tests asreview/webapp/tests
```
J535D165, line 33: What about configuring this in `pytest.ini`? It might be annoying for users to run this by default.

cskaandorp: It didn't work! But maybe I wasn't sharp.

J535D165, line 33:
```suggestion
          pytest --random-order asreview/webapp/tests
```

This should work now with the pytest.ini file.

J535D165 (commented):
It looks good to me!

J535D165:
Thanks!! Very nice!

PR #1529, opened 2023-09-21
Fix BaseReview.review bug

### LINKED ISSUE #1526: Class BaseReview has no method _write_to_state
Hello everybody,

I am currently using the API and try to run a review. When I execute the method 'review()' I get following error:
Screenshot 2023-09-12 191325

After looking into the documentation I saw that there is no method '_write_to_state()' for the class 'BaseReview'. Maybe there was one but I could not find any 🤔

PeterLombaers:
(Fixes #1526) Moves the call to `_write_to_state` from `BaseReview` to `ReviewSimulate`. It is a little ugly, but I don't think there is a pretty solution until these two classes are refactored.

### DIFF
--- a/asreview/review/simulate.py
+++ b/asreview/review/simulate.py
@@ -226,6 +226,10 @@ class ReviewSimulate(BaseReview):
             pending_labels = self.data_labels[pending]
             state.add_labeling_data(pending, pending_labels)
 
+    def review(self):
+        super().review()
+        self._write_to_state()
+
     def _stop_review(self):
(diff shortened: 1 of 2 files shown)


asreview/review/simulate.py
```diff
@@ -226,6 +226,10 @@ def _label_priors(self):
             pending_labels = self.data_labels[pending]
             state.add_labeling_data(pending, pending_labels)
 
+    def review(self):
+        super().review()
```
J535D165, line 230: Why do we need this :)

PR #1537, opened 2023-10-12
Remove fixed Werkzeug version

J535D165:
This causes issues now Flask 3.0 and Werkzeug 3.0 are released. 

@cskaandorp


J535D165:
We have to wait for a new release of https://github.com/maxcountryman/flask-login before we can remove the upper bound (`<3`).

PeterLombaers:
Seems like a good temporary solution.

PR #1558, opened 2023-11-01
Version fixes, psycopg2 and Docker test

cskaandorp:
Removed version caps on flask related packages, removed psycopg2 and added documentation about that, and tried to fix the github docker test

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Build the image for testing
         uses: docker/build-push-action@v4
         with:
-          context: .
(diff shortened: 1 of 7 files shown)


cskaandorp:
@J535D165, 2 things
* I haven't forced the Flask/Werkzeug-packages in the latest version in setup.py. This may be a bad idea because it allows people to install versions that will break with each other
* Because I had to upgrade Flask-Login I also had to change the size of a database column in the Users table. This will affect people who already have setup a database. Not sure if this needs to be made explicit in the documentation or version documentation...

J535D165 (approved):

Issue #1567, opened 2023-11-03
ASReview started from Docker Container is not able to talk to backend.


SamyAteia:
**Describe the bug**
When starting the docker container as described on the https://github.com/asreview/asreview/tree/master/Docker documentation,
The ASReview Frontend is not able to access the backend service. 

The documentation states "If you are creating a Docker container that runs the app with a config file, do not forget to override the IP-address of the Flask backend. Set the HOST variable to "0.0.0.0" since the default "localhost" can't be reached from outside the container."
I suspect that is the problem but it is not documented where the default config file that is used when following the documentation is located or how one could add such a config file to the docker build.

**To Reproduce**
1. Follow the steps in the "building the simple version" section: https://github.com/asreview/asreview/tree/master/Docker#building-the-simple-version
2. Try to access ASreview on your local machine: http://localhost:8080/
3. Look that the errors thrown on the web console.

**Version information**
 - OS: Windows
 - ASReview version 1.3

SamyAteia:
Recent hotfix https://github.com/asreview/asreview/pull/1570 seems to have fixed the issue.

PR #1569, opened 2023-11-03
Fix for simple Docker port issue

cskaandorp:
Fix simple docker

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -11,13 +11,25 @@ Creating the docker container for a simple, non-authenticated version of the app
 ```
 # create a volume
 $ docker volume create asreview_simple
-# build the container
-$ docker build  -t asreview .
-# run container
-$ docker run -d -v asreview_simple_volume:/project_folder -p 8080:5000 asreview
 ```
 
-with the external port 8080 being a suggestion. After the last command you find the app in your browser at `http://localhost:8080`.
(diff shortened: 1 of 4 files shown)


cskaandorp:
The issue is fixed, but read on! In the unauthenticated docker container there is not much room for configuration. Normally port 5000 is used, also exposed in the Dockerfile. Bridging port 8080 to 5000 was a bad idea because the origin of the request is not 5000 anymore. But sometimes, especially as a developer, you want another port. Long story short: for an easy fix both front and backend must run on the same port (no origin config). That means the backend port must be configurable without a TOML file. I have added an env variable PORT in start_flask and also the possibility to set it in the Docker setup.

I have tested it. But please do it as well.

And then I ran into another issue! Running from Docker, on port 8080, while creating a project, after adding the dataset I get a blank screen if I want to add prior knowledge: TypeError: Cannot read properties of undefined (reading 'label') (in frontend)

cskaandorp:
Fixed the frontend issue. Please have a go at it to make sure I have done it right.

J535D165:
#1567 reports that #1570 might fix the issue.

J535D165:
Thanks for the PR. #1570 fixed the issue, and the rest of it was fixed in 0f5d7a049828cdb9a4efa4388028053bedb92ea8. Credits to you @cskaandorp

PR #1591, opened 2023-11-20
Fetch dataset readers dynamically

### LINKED ISSUE #1385: Dynamically load accepted file formats for file uploads
Relates to issue addressed in #1316

PeterLombaers:
This pull request fetches the available dataset readers dynamically in the front end. This should allow contributions like #1316 to be turned into a dataset reader plugin. Closes #1385.

### DIFF
--- a/asreview/io/csv_reader.py
+++ b/asreview/io/csv_reader.py
@@ -20,6 +20,7 @@ from asreview.io.utils import _standardize_dataframe
 class CSVReader:
     """CVS file reader."""
 
+    name = "csv"
     read_format = [".csv", ".tab", ".tsv"]
     write_format = [".csv", ".tsv", ".xlsx"]
 
--- a/asreview/io/csv_reader.py
+++ b/asreview/io/csv_reader.py
@@ -21,6 +21,7 @@ class CSVReader:
     """CVS file reader."""
 
     name = "csv"
+    label = "CSV (UTF-8)"
     read_format = [".csv", ".tab", ".tsv"]
     write_format = [".csv", ".tsv", ".xlsx"]
 
--- a/asreview/io/csv_reader.py
+++ b/asreview/io/csv_reader.py
@@ -21,7 +21,7 @@ class CSVReader:
     """CVS file reader."""
 
     name = "csv"
-    label = "CSV (UTF-8)"
+    label = "CSV"
     read_format = [".csv", ".tab", ".tsv"]
     write_format = [".csv", ".tsv", ".xlsx"]
 
--- a/asreview/io/csv_reader.py
+++ b/asreview/io/csv_reader.py
@@ -20,8 +20,6 @@ from asreview.io.utils import _standardize_dataframe
 class CSVReader:
     """CVS file reader."""
 
-    name = "csv"
-    label = "CSV"
     read_format = [".csv", ".tab", ".tsv"]
     write_format = [".csv", ".tsv", ".xlsx"]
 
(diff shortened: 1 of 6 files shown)


asreview/webapp/api/projects.py
```diff
@@ -214,6 +214,29 @@ def api_init_project():  # noqa: F401
...
+@login_required
+def api_list_data_readers():
+    """Get the list of available data readers and read formats."""
+    readers = list_readers()
+    unique_readers = []
+    unique_names = set([])
+    for reader in readers:
+        if reader.name not in unique_names:
+            unique_names.add(reader.name)
+            unique_readers.append(reader)
```
J535D165, line 227: How does this work? I assume that all readers are unique in `readers`?

We know that there are custom readers out there in the wild. They don't have name and label. Can't we use the entry_point name such that we remain backward compatible? Label is optional then.

asreview/io/csv_reader.py
```diff
@@ -20,6 +20,8 @@
 class CSVReader:
     """CVS file reader."""
 
+    name = "csv"
+    label = "CSV (UTF-8)"
```
J535D165, line 24: CSV reader supports also latin-1 (see below).

asreview/webapp/api/projects.py
PeterLombaers: `list_readers` gives the readers as defined in `setup.py`. There we have defined the current readers as:
```
        "asreview.readers": [
            ".csv = asreview.io:CSVReader",
            ".tab = asreview.io:CSVReader",
            ".tsv = asreview.io:CSVReader",
            ".ris = asreview.io:RISReader",
            ".txt = asreview.io:RISReader",
            ".xlsx = asreview.io:ExcelReader",
        ],
```
So I'm basically just grouping this list by the class. Maybe it would be better to use the reader name instead of the extension in `setup.py`?

asreview/io/csv_reader.py
PeterLombaers: Ah, I copied the label from the

J535D165: Yes, let's leave out.

asreview/webapp/api/projects.py
J535D165: I'm concerned that the code breaks for users with reader extensions (and there are some out there). The label and name are not available for these extensions. What about using the following code: 

```python
@bp.route("/dataset_readers", methods=["GET"])
@login_required
def api_list_data_readers():
    """Get the list of available data readers and read formats."""

    payload = {"result": []}
    for e in _entry_points(group="asreview.readers"):

        reader = e.load()
        label = reader.label if hasattr(reader, "label") else None

        payload["result"].append(
            {
                "name": e.name[1:],
                "suffix": e.name,
                "label": label,
            }
        )
    return jsonify(payload)
```

I have tested this, and it works when no label exists. The client side is easy to adapt to this. It's slightly different from what happens with the writers, but not too much imo.

PeterLombaers: I guess it depends on if you want to get the available dataset readers, or if you want to get the accepted data formats. My code is more for the first, your code is more for the second. Right now we only use the accepted data formats, and in this case it's not really relevant what dataset reader class is used. So we could use your code. 

If someone makes a custom plugin that also reads `.txt` but parses it differently, you would need to be able to distinguish between the different reader classes.

J535D165: Fully agree. I don't have a favorite solution as long as it doesn't break custom readers.

PeterLombaers:
This should now allow pull requests like #1316 to be implemented as an extension.

PR #1606, opened 2023-12-07
Lower the DOI for records when looking for duplicates

laurens88:
Duplicate records are not recognized when one DOI is (partially) capitalized and the other is not. This PR lowers all characters at comparison.


J535D165 (approved):

J535D165:
Very welcome!

Issue #1605, opened 2023-12-07
Port specification in .env file


beffersr:
We try to install ASReview on a server using the dockerfile, which seems to work fine. Because the default ports that are used are already in use, we have to specify new ports. We do so in the .env file:
BACKEND_EXTERNAL_PORT=5015 -> 5003
BACKEND_INTERNAL_PORT=5005

EMAIL_PASSWORD="password"
WORKERS=4

FRONTEND_EXTERNAL_PORT=8080 -> 5002
FRONTEND_INTERNAL_PORT=80

POSTGRES_EXTERNAL_PORT=5433 -> 5004
POSTGRES_INTERNAL_PORT=5432

POSTGRES_PASSWORD="postgres"
POSTGRES_USER="postgres"
POSTGRES_DB="asreview_db"

The following error occurs:
![image](https://github.com/asreview/asreview/assets/2564749/47ad5cb9-dadd-41fe-852f-010f3369685e)

**Version information**
 - OS: Linux
 - ASReview version: latest

J535D165:
Thanks @beffersr for reporting. @cskaandorp can I assign you to this?

cskaandorp:
> Thanks @beffersr for reporting. @cskaandorp can I assign you to this?

Noted, I will look into this asap. If not this week, than on Monday.

cskaandorp:
I have found an issue, but it's about not using port 8080 for the frontend. I can not reproduce the error in the image. Will try to contact @beffersr

J535D165:
Most likely fixed in #1607. Thanks for reporting.

PR #1607, opened 2023-12-08
Fix advanced host and port config issues in Docker and change email configuration

cskaandorp:
Improve flexibility frontend port for Docker, fix issue #1605, streamlined app configuration.


asreview/webapp/.env.development
```diff
@@ -1 +1 @@
-REACT_APP_API_URL=http://localhost:5000/
+REACT_APP_API_URL=http://localhost:5000/
```
J535D165, line 1:
```suggestion
REACT_APP_API_URL=http://localhost:5000/

```

asreview/webapp/app.py
```diff
@@ -54,14 +55,18 @@ def create_app(
...
     app.config["ALLOWED_ORIGINS"] = ["http://localhost:3000", "http://127.0.0.1:3000"]
 
     app.config["SECRET_KEY"] = secret_key
     app.config["SALT"] = salt
 
     app.config.from_prefixed_env()
 
+    # If allowed origins comes as an environment variables, and is a list,
+    # then we assume a JSON string.
+    if type(app.config.get("ALLOWED_ORIGINS", False)) == str:
```
J535D165, line 67:
```suggestion
    if isinstance(app.config.get("ALLOWED_ORIGINS", False)), str):
```

Docker/auth_verified/flask_config.toml
```diff
@@ -1,24 +1,20 @@
...
-USE_SSL = true
-REPLY_ADDRESS = "casper@compunist.nl"
+# When using a mail server for email verification:
+# ================================================
+# MAIL_SERVER = "smtp.sendgrid.net"
+# MAIL_PORT = 465
+# MAIL_USE_TLS = false
+# MAIL_USE_SSL = true
+# MAIL_USERNAME = "apikey"
+# MAIL_DEFAULT_SENDER = "no_reply@example.org"
```
J535D165, line 20: Docs need to be updated in other places as well

asreview/webapp/app.py
```diff
@@ -95,23 +100,20 @@ def create_app(
...
+            app.config.get("MAIL_USERNAME", False),
+            app.config.get("MAIL_PASSWORD", False)
+        ])
+
         if app.config.get("EMAIL_VERIFICATION", False) and not app.config.get(
             "EMAIL_CONFIG", False
         ):
             raise ValueError(
                 "Missing email configuration to facilitate email verification"
             )
```
J535D165, line 116: Shall we remove this completely? If the email is not configured, an error can be raised when sending the email. I think we should avoid checks on the config if they arent covering all the config variables.

docs/source/server_configuration.rst
```diff
@@ -37,7 +37,7 @@ To create user accounts, one can use the ``add-users`` command of the
 
 .. code:: bash
 
-    asreview auth-tool add-users
+    asreview auth-tool add-users --db-uri=sqlite:////path/example.sqlite
```
J535D165, line 40: This is not required isn't it?

Docker/auth_verified/flask_config.toml
J535D165: All done as far as I can see!

asreview/webapp/app.py
cskaandorp: That was the initial idea, but email-verification as a parameter (being true or false) is sent with /boot to the frontend in which it is used in certain situations (like a forgot-password) link. With the [EMAIL_CONFIG] key lost in the TOML file I can't check anymore if there is a configuration available for emails. This was my poor-man's solution. Ever email config needs a mail server and probably a username and password.

docs/source/server_configuration.rst
cskaandorp: No, but I think it makes sense to explicitly add examples. I don't like setting env variable before you run a CLI script.  All sub commands, list-projects being a exception, require access to the auth database. Then provide it as a parameter. I am afraid people will miss the ENV option because people usually don't read.

asreview/webapp/app.py
cskaandorp: I misunderstood maybe. I had doubts about lines 110 and onwards. They can go. But I'd like to keep 103 - 108 since the frontend has to know whether an email service is configured (then it can show the "forgot password" link). Not the most elegant code, but it does the job.

J535D165: Ah, thanks; this was the missing link for me. In that case, I also favor removing lines 110 and onwards. I prefer to remove the config "EMAIL_CONFIG" and replace

```
            response["email_config"] = bool(app.config.get("EMAIL_CONFIG", False))

```
with
```
            response["email_config"] = all([
              app.config.get("MAIL_SERVER", False),
              app.config.get("MAIL_USERNAME", False),
              app.config.get("MAIL_PASSWORD", False)
          ])
```
or create a function `_is_email_configured()` to check for this.

By doing this, nobody can think you can set this config var via env or TOML. This is not something you can config yourself

J535D165: The other 2 references to EMAIL_CONFIG are in auth.py. Same can be applied.

asreview/webapp/app.py
```diff
@@ -54,14 +55,18 @@ def create_app(
...
     app.config["ALLOWED_ORIGINS"] = ["http://localhost:3000", "http://127.0.0.1:3000"]
 
     app.config["SECRET_KEY"] = secret_key
     app.config["SALT"] = salt
 
     app.config.from_prefixed_env()
 
+    # If allowed origins comes as an environment variables, and is a list,
+    # then we assume a JSON string.
+    if isinstance(app.config.get("ALLOWED_ORIGINS", False)), str):
```
J535D165, line 67:
```suggestion
    if isinstance(app.config.get("ALLOWED_ORIGINS", False), str):
```

docs/source/server_configuration.rst
J535D165: But the default option is independent of setting the ENV var. I do understand the idea however. Especially when taking your advanced Docker setup into account. It starts to become useful to move the advanced Docker to a standalone repo and write extensive documentation. In my recent PR on the docs, I focussed on the simple Docker setup (or not even using Docker). There, I assume most people don't change the location of the database.

asreview/webapp/app.py
cskaandorp: I have used your suggestion in app.py. In auth.py I have created a function since I had to check the same thing twice.

J535D165:
Great teamwork! Thanks!

Closes #1605

Issue #1599, opened 2023-11-29
Export records with flag "used as prior knowledge"


Rensvandeschoot:
I propose to export a flag with "used as prior knowledge". This can be a new column in CSV/Excel files with label `1`  if a records has been used to train the first iteration of the model and `0` otherwise. In RIS, a new flag can be used in the `N1` field: "ASReview_priors = true".

PR #1612, opened 2023-12-14
Add a new column 'ASReview_prior' to the dataset

Rensvandeschoot:
This pull request introduces the inclusion of a new label, asreview_prior, in exported datasets. This label explicitly marks whether a record was utilized as a 'prior' in the training of the model's first iteration

closes https://github.com/asreview/asreview/issues/1599


asreview/webapp/api/projects.py
```diff
@@ -1166,6 +1166,13 @@ def api_export_dataset(project):
         # read the dataset into a ASReview data object
         as_data = read_data(project)
 
+        # Retrieve prior information from state
+        with open_state(project.project_path) as s:
+            priors = s.get_priors()["record_id"].tolist()
+
+        # Add a new column 'is_prior' to the dataset
+        as_data.df['is_prior'] = as_data.df.index.isin(priors).astype(int)
```
J535D165, line 1174: isin can introduce performance issues. It is better to avoid it. We can make use of a `join` similar to the notes below. 

You might be able to avoid the open_state call (expensive) if you reuse state_df.

J535D165 (changes requested):

asreview/webapp/api/projects.py
PeterLombaers: Yeah, instead of opening the state again you could do (didn't test this):
```[python]
state_df["is_prior"] = state_df.query_strategy.eq("prior").astype(int)
as_data.df = as_data.df.join(state_df["is_prior"], on="record_id")
```

Rensvandeschoot: image

Rensvandeschoot: @PeterLombaers  your suggestion seems do work!

Rensvandeschoot:
Do we want to name the column `is_prior` or `ASReview_prior` to make it consistent with how we export to RIS files (`ASReview_relevant` etc)?

asreview/webapp/api/projects.py
```diff
@@ -1166,6 +1166,10 @@ def api_export_dataset(project):
         # read the dataset into a ASReview data object
         as_data = read_data(project)
 
+        # Add a new column 'is_prior' to the dataset
+        state_df["asreview_prior"] = state_df.query_strategy.eq("prior").astype(int)
+        as_data.df = as_data.df.drop("asreview_prior", axis=1).join(state_df["asreview_prior"].astype(int), on="record_id")
+
```
J535D165, line 1172: We can keep this and revert the changes to ris_reader.py and ris_writer.py. Then it should work. 

I don't know what to do with this info during the setup yet. Currently, it's ignored in favor of the labeled column.

J535D165 (approved):

J535D165 (approved):

PR #1611, opened 2023-12-14
Rename 'debug_label' to 'label_included' in export file

Rensvandeschoot:
This pull request modifies column names in exported datasets from ASReview to enhance clarity, based on the project mode:

- Exploration Mode: The column initially containing labeling decisions, internally renamed to 'debug_label', is exported as 'label_included'. A new column, 'label_validated', stores labels made during screening.
- Simulation Mode: Renames 'debug_label' to 'label_included' in the exported dataset.
- Oracle Mode: No changes, as 'debug_label' does not exist in this mode.

PR #1623, opened 2024-01-03
Refactor debug label and remove internal use

J535D165:
This PR removes the use of the internal `debug_label` while keeping the functionality. A side effect of this is a new feature: export Explore and Simulate projects to RIS format if they were provided as RIS files.


J535D165:
I tested the Explore mode:

- [x] old projects still work
- [x] new projects can be initialized and prior knowledge can be added

I haven't tested the Simulate mode yet, but I might expect some issues there.

J535D165:
Simulate also seems to work for me

J535D165:
Closes #1611 as it is implemented in [86a9038].(https://github.com/asreview/asreview/pull/1623/commits/86a90382feb2b2262f21f0ec619c5e0e3657497f)

PR #1638, opened 2024-01-05
Add code format check to GitHub Actions workflow

terrymyc:
This PR attempts to add code format checks to the GitHub Actions workflow to prevent merging unformatted code.

PR #1640, opened 2024-01-06
Introduce new linter and pre-commit

J535D165:
Introducing pre-commit with Python code formatter and JS formatter


J535D165:
Relates to #1638

PR #1661, opened 2024-01-25
Move AI models to another package

J535D165:
This PR proposes to move all advanced models to other packages/another package. This affects the models with dependency tensorflow, sentences transformers, and gensim. 

The main reasons to do this:
- make the installation process of ASReview easier. No more `asreview[all]` or `asreview[tensorflow]` but `pip install asreview asreview-ai-models`. 
- easier version management for dependencies
- faster iterations for developing the advanced models (lots of work in the pipeline)

Todo:

- [ ] A good name for the new package (using work title asreview-ai-models now)
- [x] Some frontend improvements in the model selection panel
- [ ] Create the new package
- [ ] Add unit tests for the new package
- [ ] Add option to switch model in frontend as part of migration workflow

I might not add all of these items to this PR.


asreview/webapp/src/ProjectComponents/SetupComponents/ModelComponents/ModelRequirement.js
```diff
@@ -24,7 +24,7 @@ const requirement = (requirement) => {
 
   return (
     <>
-      requires <code>{requirement}</code> to be installed.{" "}
```
PeterLombaers, line 27: Maybe put a link to install instructions, or the actuall CLI input here?

PeterLombaers:
I like the proposal. It removes all the ugly checks if tensorflow is installed! We should probably have a clear idea about what an 'AI' model is though. Basically we're saying that anything that can't be done with plain Python or `sklearn` is part of `asreview-ai-models`? Maybe the name should just be `asreview-models`.

We should also be very careful about versioning the models. What happens if I have two projects, in one I use `ModelA` and in the other I use `ModelB`. I want to update `ModelB` because we improved the performance dramatically. Do I just `pip install -U asreview-ai-models`? Does that also update `ModelA`? Do we record that the models changed?

asreview/webapp/src/ProjectComponents/SetupComponents/ModelComponents/ModelRequirement.js
J535D165: Thanks for this comment. I deprecate ModelRequirements.js and implement your suggestion in PR #1664. 

I make use of the link as I prevent CLI input suggestions because of the server version of ASReview.

J535D165:
Thanks for this feedback. Very valuable. Will drop some ideas on versioning the models later. 

Regarding the models: If the model depends on a stable and easy-to-install package, I propose keeping it in ASReview (sklearn, spacy, etc.). For models with challenging and problematic installation procedures, like most neural network-based packages, I propose to separate them from asreview/asreview. 

Hopefully, the others also have some suggestions for the names. I like the simplicity of asreview-models. Or something like asreview-emm (Even More Models).

Rensvandeschoot:
Can we store the version number of the extra package in project. config?

Rensvandeschoot:
Some suggestions for names:

- asreview-emm (Even More Models)
- asreview-plus
- aseview-deeplearn: Emphasizing the deep learning aspect of the models it provides
- asreview-vanguard, where "Vanguard" denotes leading the way, suggesting this package includes the most advanced, state-of-the-art AI models

J535D165:
Feedback is still welcome. Merged to prevent merge conflicts.

PR #1666, opened 2024-01-29
Idea for project type selection

J535D165:
I'm experimenting with new ideas for the project type selection. Ideas welcome. 

This idea puts more focus on the differences between the project types. It should make this information redundant: https://asreview.readthedocs.io/en/stable/project_create.html#project-modes

image


J535D165:
Another idea (that needs refinement) is this: 

image

It removes the 9 times "dataset" and add boxes for clarity. We can highlight selected boxes.

PR #1669, opened 2024-01-30
Change location of pickle data file

J535D165:
This PR changes the location and name of the pickle data file. This is the first PR in a sequence of PRs to make uploading multiple datasets or extending existing ones possible. 

Adds tests as well.

### DIFF
--- a/asreview/project.py
+++ b/asreview/project.py
@@ -320,7 +320,7 @@ class Project:
     def remove_dataset(self):
         """Remove dataset from project."""
         # reset dataset_path
-        self.update_config(dataset_path=None, datasets=None)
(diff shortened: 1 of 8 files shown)


asreview/project.py
```diff
@@ -320,10 +320,11 @@ def add_dataset(self, file_name):
     def remove_dataset(self):
         """Remove dataset from project."""
         # reset dataset_path
-        self.update_config(dataset_path=None)
+        self.update_config(dataset_path=None, datasets=None)
```
J535D165, line 323:
```suggestion
        self.update_config(dataset_path=None)
```

PR #1686, opened 2024-02-12
Improve project setup, add extra configuration in details page, and train in review page

J535D165:
This PR implements many changes and refactors:

- [x] Make use of Context for project_id in SetupDialog
- [ ] Training of the first iteration of the model is moved to the review page. This makes it possible to work on a quick project setup later on, where you can start screening right after the dataset is uploaded. 
- [x] Completely redesigned details page
- [x] Option to change model in the details page
- [x] Option to add tags in the details page
- [x] Refactor the setupDialog dialog
- [x] Reduce the number of unnecessary API calls
- [x] Reduce the number of state changes
- [x] Remove custom details page components
- [ ] Integrate prior knowledge selection better with SetupDialog
- [ ] Screen without prior knowledge
- [ ] Option to see what model was used for training the displayed record


J535D165:
The save and next button are integrated now when project info is stored: 

image

This is useful because updating this info should not auto-update because of the server-side checks like a project-title check. This required several changes (including a visual one), such as making the Stepper part of the DialogHeader.

J535D165:
As the entire dialog is now optional, there is no need to collapse the accordion anymore.

image

J535D165:
I implemented an indicator to show the details of the model used for ranking the displayed record. The presented record has a green indicator (ModelTraining icon from Material) with info when you hover when the record comes from a specific model. The presented record comes from the random or top-down stack when the indicator is orange. 

I'm planning a red indicator when model training has failed.

image
image

J535D165:
Model switching and projects without prior knowledge:

image

J535D165:
"Tag categories" are now "Tag groups". This makes more sensw given the implementation imo. 

image

J535D165:
The "I can't wait" button is available when the model is training. This button offers the option to screen random or top-down. 

I'm working on the review screen when a project without prior knowledge (and therefore is not training) is presented. Picking one of these indicators results in orange indicators during screening until enough info is found to train a model. 


image

PR #1696, opened 2024-03-11
Change default production WSGI server by waitress and add support for other WSGI servers

J535D165:
This PR introduces various changes to the Flask production server and configuration. Changes should resolve various reported issues and improve the extensibility of the review-server-stack. Although we aim to avoid breaking changes, we can not fully guarantee that current ASReview LAB Server users are affected by this. If you face issues, please let us know. 

We aim to create a version 1.6 release for these changes. 

## Changelog

- Replace default event server by full Python waitress server
- Resolve installation problems and segfault crashes for MacOS users. 
- Make use of `ASREVIEW_LAB` prefix for all environment variables, FLASK environment variables included. You can still use FLASK prefixed environment variables in version 1.x. 
- Remove arguments from `create_app()`, only argument is `config_path` (points to ASReview configuration file).
- Make `create_app()` authenticated by default
- Remove `env` from `create_app()`. Setting 'development' or 'testing' now via `app.debug` and `app.testing`. 
- Add option to use any production WSGI server to server the app. See Flask https://flask.palletsprojects.com/en/3.0.x/tutorial/deploy/ for info. Example: `gunicorn -w 4 -b "127.0.0.1:5006" "asreview.webapp.app:create_app()"`
- Deprecate `--certfile` and `-keyfile` options. They will remain available in the 1.x series. For TLS support, use for example gUnicorn with certfile and keyfile options.

## Breaking changes

- When you rely heavily on environment variables or the ASReview/FLASK config file to configure ASReview, please look at the new variable overview in the documentation. We tried to keep many of the variables working, but we might have missed something. Also start using the `ASREVIEW_LAB` prefix. 
- Developers will face the authenticated version of ASReview by default. You can change this with `ASREVIEW_LAB_LOGIN_DISABLED=true flask run --debug`. I will update the VSCode task according to this.

### DIFF
--- a/asreview/webapp/app.py
+++ b/asreview/webapp/app.py
@@ -41,13 +41,19 @@ from asreview.webapp.authentication.models import User
 from asreview.webapp.authentication.oauth_handler import OAuthHandler
 
 
-def create_app(
-    env="development",
-    config_file=None,
-    secret_key=None,
-    salt=None,
-    enable_authentication=False,
-):
+def create_app():
+    """Create a new ASReview webapp.
+
+    For use with WSGI servers, such as gunicorn:
+
+        > gunicorn -w 4 -b "127.0.0.1:5006" "asreview.webapp.app:create_app()"
+
+    Returns
+    -------
+    Flask:
+        The ASReview webapp.
+    """
+
     app = Flask(
         __name__,
         instance_relative_config=True,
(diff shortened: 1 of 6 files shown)


asreview/webapp/app.py
```diff
@@ -41,50 +41,38 @@
...
 
-    # If allowed origins comes as an environment variables, and is a list,
-    # then we assume a JSON string.
-    if isinstance(app.config.get("ALLOWED_ORIGINS", False), str):
-        app.config["ALLOWED_ORIGINS"] = json.loads(app.config["ALLOWED_ORIGINS"])
+    # if there are no cors and config is in debug mode, add default cors
+    if app.debug and not app.config.get("CORS_ORIGINS", None):
+        app.config["CORS_ORIGINS"] = ["http://localhost:3000", "http://127.0.0.1:3000"]
 
-    if config_file_path := config_file or app.config.get("CONFIGFILE", ""):
```
J535D165, line 70: Renamed this env var from "FLASK_CONFIGFILE" into "ASREVIEW_LAB_CONFIG_PATH". Needs backward compatility?

asreview/webapp/entry_points/lab.py
```diff
@@ -30,12 +31,14 @@
...
-    HOST_NAME = "localhost"
+try:
+    from gevent.pywsgi import WSGIServer
+except ImportError:
+    WSGIServer = None
 
-PORT_NUMBER = 5000
+# Host name
+HOST_NAME = os.getenv("ASREVIEW_HOST", "localhost")
+PORT_NUMBER = os.getenv("ASREVIEW_PORT", 5000)
```
J535D165, line 41:
```suggestion
HOST_NAME = os.getenv("ASREVIEW_LAB_HOST", "localhost")
PORT_NUMBER = os.getenv("ASREVIEW_LAB_PORT", 5000)
```

J535D165: Does this need backward compatibility?

asreview/webapp/entry_points/lab.py
```diff
@@ -205,30 +255,29 @@ def _lab_parser():
 
     parser.add_argument(
         "--enable-auth",
-        dest="enable_authentication",
-        action="store_true",
+        dest="login_disabled",
+        default=True,
+        action="store_false",
```
J535D165, line 260: Bit confusing syntax, but tested and it works

docs/source/server_configuration.rst
```diff
@@ -155,6 +155,49 @@ folder:
 
     FLASK_SQLALCHEMY_DATABASE_URI = "sqlite:///asreview.production.sqlite"
 
+
+Full configuration
+~~~~~~~~~~~~~~~~~~~
```
J535D165, line 160: Also adjust the rest of the docs. Prefix FLASK should become ASREVIEW_LAB

asreview/webapp/app.py
J535D165: Resolved

PR #1697, opened 2024-03-11
Add a login banner

cskaandorp:
Add a blue-colored text banner at the top of the signin and signup page to share a message to users (like "This is a demo version"). Configurate the message (string) via ENV variable FLASK_LOGIN_INFO or in the flask_config TOML file with LOGIN_INFO.

### DIFF
--- a/asreview/webapp/app.py
+++ b/asreview/webapp/app.py
@@ -165,19 +165,15 @@ def create_app(
         """Get the boot info."""
 
         authenticated = not app.config.get("LOGIN_DISABLED", False)
+        login_info = app.config.get("LOGIN_INFO", False)
 
         response = {
             "authentication": authenticated,
             "version": asreview_version,
+            "login_info": login_info,
         }
 
         if authenticated:
-            # if recaptcha config is provided for account creation
-            if app.config.get("RE_CAPTCHA_V3", False):
-                response["recaptchav3_key"] = app.config["RE_CAPTCHA_V3"].get(
-                    "KEY", False
-                )
-
             response["allow_account_creation"] = app.config.get(
                 "ALLOW_ACCOUNT_CREATION", False
             )
--- a/asreview/webapp/src/App.js
+++ b/asreview/webapp/src/App.js
@@ -60,6 +60,7 @@ const App = (props) => {
   );
   const emailConfig = useSelector((state) => state.email_config);
   const emailVerification = useSelector((state) => state.email_verification);
+  const loginInfo = useSelector((state) => state.login_info);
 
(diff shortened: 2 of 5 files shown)


asreview/webapp/app.py
```diff
@@ -165,19 +165,15 @@ def api_boot():
...
+            "login_info": login_info,
         }
 
         if authenticated:
-            # if recaptcha config is provided for account creation
-            if app.config.get("RE_CAPTCHA_V3", False):
-                response["recaptchav3_key"] = app.config["RE_CAPTCHA_V3"].get(
-                    "KEY", False
-                )
-
```
J535D165, line 180: Can you explain?

J535D165 (commented):
Thanks, It might also be nice to port this back to v1.6.

asreview/webapp/app.py
cskaandorp: Nothing is done with it. I have started it, but nothing concrete is done with it in the frontend

J535D165: Oke clear.

asreview/webapp/app.py
```diff
@@ -165,19 +165,15 @@ def api_boot():
         """Get the boot info."""
 
         authenticated = not app.config.get("LOGIN_DISABLED", False)
+        login_info = app.config.get("LOGIN_INFO", False)
 
         response = {
             "authentication": authenticated,
             "version": asreview_version,
+            "login_info": login_info,
```
J535D165, line 173:
```suggestion
        response = {
            "authentication": authenticated,
            "version": asreview_version,
            "login_info": app.config.get("LOGIN_INFO", None),
```

asreview/webapp/src/App.js
```diff
@@ -237,6 +238,21 @@ const App = (props) => {
           <CssBaseline />
 
           <div aria-label="nav and main content">
+            {typeof loginInfo === "string" && loginInfo.length > 0 && (
```
J535D165, line 241:
``` suggestion
            {loginInfo !== null && (
```

asreview/webapp/src/App.js
```diff
@@ -237,6 +238,21 @@ const App = (props) => {
...
+              <Alert
+                severity="info"
+                variant="standard"
```
J535D165, line 244: Needed?

cskaandorp: Severity determines the color. Despite te fact that "standard" is the default value for variant, I had to explicitly set it to force the less-obtrusive transparent blue color. The opaque blue seemed standard.

PR #1699, opened 2024-03-13
Add support for customizable API URL in production without rebuilding

J535D165:
Example:

Process 1
```
ASREVIEW_LAB_PORT=5008 asreview lab
```

Process 2

```
ASREVIEW_LAB_API_URL="http://localhost:5008/" asreview lab
```

Now, process 1 makes use of process 2 for backend calls.

### DIFF
--- a/asreview/webapp/src/globals.js
+++ b/asreview/webapp/src/globals.js
@@ -4,16 +4,10 @@ import { setProject } from "./redux/actions";
 import ASReviewLAB_black from "./images/asreview_sub_logo_lab_black_transparent.svg";
 import ASReviewLAB_white from "./images/asreview_sub_logo_lab_white_transparent.svg";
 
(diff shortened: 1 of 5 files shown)


asreview/webapp/src/globals.js
```diff
@@ -4,16 +4,10 @@ import { setProject } from "./redux/actions";
...
-// the same as the front-end URL.
-let b_url = "/";
-if (Boolean(process.env.REACT_APP_API_URL)) {
-  b_url = process.env.REACT_APP_API_URL;
-}
-export const base_url = b_url;
-export const api_url = base_url + "api/";
-export const auth_url = base_url + "auth/";
-export const collab_url = base_url + "team/";
+export const base_url = window.api_url;
```
J535D165, line 7:
```suggestion
```

PR #1708, opened 2024-03-22
Bug fix profile page and moved the info banner from App into Signin and Signup

cskaandorp:
A fix to solve the disabled state of the Save button in the Profile page when using OAuth, moved the info banner to signin and signup to prevent the banner from rendering in the review page.


J535D165:
Thanks, I will need some help to fix the oauth button problem in the main branch as well. The infobar will be return in a different shape and doesn't need to be ported.

PR #1710, opened 2024-03-27
Add boot call functionality to index.html

J535D165:


J535D165:
@cskaandorp This PR changes the behavior of the emailConfig variable. I wasn't able to test it completely yet. Hopefully, we can look into this together later. First it might be good to change some other things (in other PRs) as well.

PR #1718, opened 2024-04-16
Add the OneHot encoder

jteijema:
By setting the countvectorizer to binary, we effectively have a OneHot encoder.


J535D165 (approved):

J535D165:
Thanks

PR #1720, opened 2024-04-18
Rework dataset descriptives screen

laurens88:
Added charts to visualize dataset properties and give quick insights to users.

Color of charts reflect thresholds: 
green =>95%, 
orange => 75%, 
red lower

For now the chart for English language is hardcoded to 50% as placeholder for actual implementation.
image


J535D165:
Merci!!

PR #1742, opened 2024-05-13
logistic: Removing unused n_jobs arg from LogisticClassifier

### LINKED ISSUE #1740: n_jobs for logisticRegression is not usable
**Description of the issue :  **
We are proposing a `n_jobs` parameter for the [asreview.models.classifiers.logistic  classifier](https://github.com/asreview/asreview/blob/main/asreview/models/classifiers/logistic.py#L51) . This parameter is passed to the underlying [sklearn.linear_model LogisticRegression](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#sklearn.linear_model.LogisticRegression). However this classifier makes use of n_jobs>1 only in case of multi class classification (and with another solver than liblinear that is used as default in asreview). Because we are only facing a binary classification task in ASReview, this `n_jobs` parameter seems not needed.  

**Changes needed**:  
I would suggest to remove the `n_jobs` parameter from the asreview logistic classifier because changing it does not change anything to the learning pipeline. I will be happy to implement a PR if this is pertinent. 

NB: I am not aware that we are doing some HP selection on top of the ASreview logistic classifier. However, if this is the case, we should keep n_jobs=-1 for the asreview.logistic classfiier to pass it to HP selection but we can still remove it from the sklearn.linear_model Logistic regression.

strayMat:
Small fix to remove useless n_jobs from LR classifier.

closes #1740

### DIFF
--- a/asreview/models/classifiers/logistic.py
+++ b/asreview/models/classifiers/logistic.py
@@ -37,24 +37,20 @@ class LogisticClassifier(BaseTrainClassifier):
         Class weight of the inclusions.
     random_state: int, asreview.utils.SeededRandomState
         Random state for the model.
-    n_jobs: int
-        Number of CPU cores used.
     """
 
     name = "logistic"
     label = "Logistic regression"
 
-    def __init__(self, C=1.0, class_weight=1.0, random_state=None, n_jobs=1):
+    def __init__(self, C=1.0, class_weight=1.0, random_state=None):
         super().__init__()
         self.C = C
         self.class_weight = class_weight
-        self.n_jobs = n_jobs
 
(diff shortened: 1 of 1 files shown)


jteijema:
```
n_jobsint, default=None

 Number of CPU cores used when parallelizing over classes if multi_class=’ovr’”. 
This parameter is ignored when the solver is set to ‘liblinear’ regardless of whether
 ‘multi_class’ is specified or not.
```

Good catch by @strayMat

jteijema (approved):

J535D165:
Thanks!

Issue #1747, opened 2024-05-17
"Labeled since last irrelevant count" with only priors is not useful


J535D165:
**Describe the bug**
Opening a project with prior knowledge results in a "Labeled since last irrelevant count" value that is not useful and confusion. The order of the prior knowledge should be independent of the "Labeled since last irrelevant count" value. 

![image](https://github.com/asreview/asreview/assets/12981139/c4fb19d0-2bca-4198-97f2-3df622f9889f)

**To Reproduce**

1. create project
2. Add prior knowledge
3. Go to analytics screen. 

**Solution** 

Don't display "Labeled since last irrelevant count" when there is no prior info. See also the charts.

J535D165:
Fixed in https://github.com/asreview/asreview/pull/1766.

PR #1748, opened 2024-05-21
Fixing Issue #1747: Add conditional rendering for prior knowledge card in NumberCard.js

BerkeYazan:
- Added a new function `hasPriorKnowledge` to check for prior knowledge data.
- Updated the conditional rendering logic to use the new function.

Fixes: https://github.com/asreview/asreview/issues/1747

### DIFF
--- a/asreview/webapp/src/ProjectComponents/AnalyticsComponents/NumberCard.js
+++ b/asreview/webapp/src/ProjectComponents/AnalyticsComponents/NumberCard.js
@@ -18,7 +18,9 @@ export default function NumberCard(props) {
       props.progressQuery.isSuccess
     );
   };
-
+  const hasPriorKnowledge = () => {
+    return showNumber() && props.progressQuery.data["n_included"] > 0;
+  };
   return (
     <Root>
       <CardErrorHandler
@@ -84,6 +86,7 @@ export default function NumberCard(props) {
             </CardContent>
           </Card>
         </Grid>
+        {hasPriorKnowledge() && (
         <Grid item xs={12} sm={12}>
           <Card className="number-card" elevation={2}>
             <CardContent>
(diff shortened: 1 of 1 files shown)


asreview/webapp/src/ProjectComponents/AnalyticsComponents/NumberCard.js
```diff
@@ -18,7 +18,9 @@ export default function NumberCard(props) {
       props.progressQuery.isSuccess
     );
   };
-
+  const hasPriorKnowledge = () => {
+    return showNumber() && props.progressQuery.data["n_included"] > 0;
+  };
```
J535D165, line 23: This might resolve the problem partially. 

When uploading a dataset and including/excluding a couple of records, this number will still display. I think we should also disable this. The only solution I see at the moment is to extend https://github.com/asreview/asreview/blob/99bccf2ac8b9319597f60a3ba19189c3121991a3/asreview/webapp/api/projects.py#L1146 with a key-value for irrelevant since last relevant without priors.

asreview/webapp/src/ProjectComponents/AnalyticsComponents/NumberCard.js
```diff
@@ -84,35 +86,37 @@ export default function NumberCard(props) {
...
+                      }
+                      displayType="text"
+                      thousandSeparator
+                    />
+                  </Typography>
+                </Stack>
+              </CardContent>
+            </Card>
+          </Grid>
+        )}
```
J535D165, line 119: Although I proposed the solution in the issue like this, it might be better to display a different message or `-` instead of the number rather than a conditional display. What do you think?

J535D165 (changes requested):
Thanks for the implementation. Promising. 

I haven't tested it yet, but did a review of the code. I will DM you the dataset I used in the issue. Let's check if this PR resolves the issue for that dataset.

PR #1749, opened 2024-05-22
Fixing Issue #1747: Add a new key-value pair to handle the irrelevant count without priors

BerkeYazan:
Fixes the issue #https://github.com/asreview/asreview/issues/1747: 

**Description**

Addresses the issue of displaying a misleading "irrelevant since last relevant" count when prior knowledge is included in the dataset.

**Changes**

- Modified the backend to differentiate between counts with and without prior knowledge.
_(Added a new key-value pair to handle the irrelevant count without priors)_

- Updated the front-end to display '-' instead of '0' for the "irrelevant since last relevant" count when prior knowledge is present in the dataset.

**Testing**

The previously mentioned test dataset with prior knowledge was uploaded to the system. The "irrelevant since last relevant" count displays:

- **-** after immediately going to the Analytics dashboard
- **-** after adding more prior knowledge before going to the Analytics dashboard
- **0** after labelling the first relevant paper
- **The intended number** after labelling remaining papers as relevant-irrelevant


asreview/webapp/api/projects.py
```diff
@@ -1125,23 +1132,54 @@ def _get_stats(project, include_priors=False):
 
         n_records = len(as_data)
 
+        # Get labels without priors
+        labels_without_priors = s.get_labels(priors=False)
+
```
J535D165, line 1137: This causes the tests to fail I think. This should be inside the context manager.

Issue #1452, opened 2023-05-24
Mixed query strategies have misleading parameter name `mix_ratio`


jspaaks:
The mixed query strategies `MaxRandomQuery`, `MaxUncertaintyQuery` and `MixedQuery` use a confusingly named parameter `mix_ratio` whose name suggests that the ratio is between the two strategies, which it is not.

Relevant file:

https://github.com/asreview/asreview/blob/caf788b36aa1483556e903448b20aa53ba0c4995/asreview/models/query/mixed.py

PR #1760, opened 2024-06-13
Simplify Query Strategy

PeterLombaers:
This pull request aims to simplify the query strategy class. 

Currently there are two 'base' classes for query strategies: `BaseQueryStrategy` and `ProbaQueryStrategy`. The first expects someone to implement the method `query` which takes the feature matrix, a classifier and the number of result instances as input. The second expects someone to implement `_query` which takes as input the predicted relevance scores by the classifier, the feature matrix and the number of result instances. The methods should return the row indices of the feature matrix in ranked order, but the methods have an additional option to also return the relevance scores of the classifier.

This basically means that the classifier can't be separated from the query strategy, and that if you want the results of the classifier, you need to go through the query strategy.

This pull request:
- Removes `ProbaQueryStrategy`
- Renames the `query` input argument `X` to `feature_matrix`
- Replaces the `query` input argument `classifier` by `relevance_scores`
- Removes the `query` input argument `n_instances` since it wasn't used anywhere, the method will always return a ranking for all the records in the feature matrix
- Renames the `mix_ratio` argument for `MixQuery` to `mix_probability` and better describes the meaning in the docstring 

Note that this does limit a little bit what you can do with query strategies, since a query strategy no longer has access to everything that a classifier produces. However, you can always get around this by implementing this query logic at the end of the `predict_proba` method of the classifier.

### DIFF
--- a/asreview/models/query/base.py
+++ b/asreview/models/query/base.py
@@ -24,7 +24,8 @@ class BaseQueryStrategy(BaseModel):
 
     name = "base"
 
-    def query(self, feature_matrix, relevance_scores, n_instances=None, **kwargs):
+    @abstractmethod
+    def query(self, feature_matrix, relevance_scores, **kwargs):
         """Put records in ranked order.
 
         Arguments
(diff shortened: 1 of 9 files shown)


PeterLombaers:
Fixes #1452

J535D165 (approved):

PR #1763, opened 2024-06-15
Add reset-password feature for auth-tool 

cskaandorp:
Reset password with the auth-tool command line tool.


cskaandorp:
I've committed asreview/webapp/.env.development by accident, tried to remove it. Hope all is OK now.

J535D165 (approved):

PR #1765, opened 2024-06-15
Add reset-password feature for auth-tool

cskaandorp:
Reset password with the auth-tool command line tool.


asreview/webapp/entry_points/auth_tool.py
```diff
@@ -232,14 +246,39 @@ def add_users(self):
...
+
     def _ensure_valid_value_for(self, name, validation_function, hint=""):
         """Prompt user for validated input."""
         while True:
             value = input(f"{name}: ")
             if validation_function(value):
                 return value
             else:
-                sys.stderr.write(hint)
+                sys.stderr.write(hint + "\n")
```
J535D165, line 281: This causes a test to fail. What should it be?

PR #1769, opened 2024-06-18
Remove Actions that test heavy model (NN, sbert, doc2vec)

J535D165:
@jteijema  are models tested in the upcoming NEMO model extension?


jteijema:
Assuming its for me not for @jelletreep , no actual testing of the models is TODO

PR #1770, opened 2024-06-18
Resolves #1361: Inconsistencies in Y-Axis Intervals and Alignments for Progress and Recall Charts

### LINKED ISSUE #1361: minor plotting bugs in webapp's analytics page
1. plot "Progress" has unequal intervals on the yaxis (I have y=0, 3, 7, 10)
1. plot "Recall" has unequal intervals on the yaxis (I have y=0, 2, 5, 7, 9, 12, 14) 
1. plot "Recall" has horizontal gridlines that don't agree with the data being displayed, for instance the gridline at y=2 is higher than either plotted line where they are y=2

Progress screenshot:

image

Recall screenshot:

image

BerkeYazan:
### Description
Resolves https://github.com/asreview/asreview/issues/1361.

In the Analytics dashboard, the Y-axis gridlines of the _Progress_ and _Recall_ plots were not equally apart. Additionally, for the 'Recall' plot, there was a mismatch between the values in the plot and the values of the gridlines.

### Changes
These problems occurred because the corresponding tick amounts on the Y-axis weren't consistently divisible by the max amounts on the Y-axis. The following changes were made to address these issues:

1. Progress Chart: The `tickAmount` in the _Progress_ chart was changed to 5 from 3, which is a divisor of 10.
 
2. Recall Chart: The `tickAmount` in the _Recall_ chart was fixed to 7. Also, the maximum value on the Y-axis will now be switched by the closest number that is divisible by 7.

### Testing

- The grid amount in the _Progress_ chart is always fixed and displays the correct value.
- The maximum value of the _Recall_ chart is 5 when there are no labellings. With the first labelling, it goes to 7. It stays at 7 when 7 labellings are done. Then goes to 14 with the 8th labelling. This process repeats with every 7 labelled records.

### DIFF
--- a/asreview/webapp/src/ProjectComponents/AnalyticsComponents/ProgressDensityChart.js
+++ b/asreview/webapp/src/ProjectComponents/AnalyticsComponents/ProgressDensityChart.js
@@ -254,7 +254,7 @@ export default function ProgressDensityChart(props) {
         showAlways: false,
         max: 10,
         min: 0,
-        tickAmount: 3,
+        tickAmount: 5,
         title: {
           text: "Number of relevant records",
         },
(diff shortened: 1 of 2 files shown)


J535D165:
Great!

PR #1773, opened 2024-06-27
Update feature matrix reading and writing

PeterLombaers:
This pull requests moves the code for reading and writing feature matrices from the project class to the base feature extraction class. This way people can overwrite the method and implement their own format to store a feature matrix.

Some comments:
- I called the methods `read` and `write`. My other main option was `load` and `save` because that's what the matrix storing methods are called in `numpy`. In ASReview I found more mentions of `read` though. However even only in the project class I found `Project.load` and `Project.read_data`, so we don't have any consistent naming at the moment.
- I made `BaseFeatureExtraction.load` and `save` static methods since they don't use anything of the class itself at the moment. Maybe they should be class methods, or just methods?
- To call `Project.get_feature_matrix` you now need the actual feature extraction class instead of just it's name. For this I used `feature_model = load_extension("models.feature_extraction", settings.feature_extraction)`, but this seems a little ugly to ask users of the Python API to use. Maybe we should have a dedicated `load_feature_model` function?

### DIFF
--- a/asreview/models/feature_extraction/base.py
+++ b/asreview/models/feature_extraction/base.py
@@ -15,18 +15,31 @@
 __all__ = ["BaseFeatureExtraction"]
 
 from abc import abstractmethod
+from pathlib import Path
 
 import numpy as np
+from scipy.sparse import csr_matrix
 from scipy.sparse import hstack
 from scipy.sparse import issparse
+from scipy.sparse import load_npz
+from scipy.sparse import save_npz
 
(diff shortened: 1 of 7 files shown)


J535D165 (approved):
Clean

J535D165:
@jteijema please apply where beneficial.

jteijema:
Looks nice! Thanks for the headup @J535D165

PR #1784, opened 2024-07-22
Fix temp directory removal in Project.load()

J535D165:


jteijema (commented):
Looks good, it cleans up on my system now :)

PR #1796, opened 2024-07-29
Integrate Huey queue to schedule model training

cskaandorp:
Run huey consumer from root folder with:
```
$ huey_consumer asreview.webapp.run_huey.huey 
```


asreview/webapp/api/projects.py
```diff
@@ -1231,16 +1244,10 @@ def api_label_record(project, record_id):  # noqa: F401
...
-        subprocess.Popen(
-            [
-                sys.executable if sys.executable else "python",
-                "-m",
-                "asreview",
-                "web_run_model",
-                str(project.project_path),
-            ]
-        )
+    if _project_not_in_queue(project) and retrain_model:
```
J535D165, line 1247:
```suggestion
    if retrain_model and _project_not_in_queue(project):
```

cskaandorp:
1. Huey is started from the `asreview lab` command. That works.
2. As a developer it is possible to start the consumer with `$ huey_consumer asreview.webapp.tasks.huey` from the root folder. Is this good enough?
3. Tests are still failing, I would like to bypass Huey completely but not sure if this is possible
4. Docker: todo

J535D165:
Implementing Huey results in many challenges. Huey is hard to implement when integrated directly into the Flask app. Despite some good examples, it is more complicated than implementing the queue directly into ASReview. We were inspired by some smart ideas in Huey and continued in #1861.

PR #1810, opened 2024-08-22
Mark tests as requiring internet

PeterLombaers:
I'm in the train at the moment and the internet is quite spotty. So I thought it might be useful to be able to skip tests that require an internet connection to run. Not sure if this is a contribution we really need, but let me know what you think! 

I added a pytest mark `internet_required` that you can add to any test that requires an internet connection. Then you can run the tests with `pytest -k "not internet_required"` to only run the offline tests. 

I only marked tests outside of webapp, since I thought it would make less sense to have something like this in webapp. I'm also not quite sure if I've got all of the tests requiring internet (I'm also running into some other problems relating to cleaning up temporary directories on Windows, so the errors I'm getting don't make much sense). Tests that require the Synergy dataset first look locally if they can find it I think, so those only require a connection if you don't have Synergy locally.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -196,3 +196,6 @@ docstring-code-line-length = "dynamic"
 
 [tool.pytest.ini_options]
 addopts = "--ignore=asreview/webapp/tests/integration_tests"
+markers = [
+    "internet_required: requires internet connection to run",
+]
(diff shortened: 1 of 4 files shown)


J535D165:
Thanks!

Issue #1815, opened 2024-08-27
Add `latest` tag to Docker image


NN708:
The [Installation](https://asreview.readthedocs.io/en/latest/installation.html#install-with-docker) documentation states that the Docker image can be pulled from `ghcr.io/asreview/asreview:latest`. However, the `latest` tag does not exist in the container registry.

jteijema:
Should be an easy fix! Added the latest tag.

PR #1818, opened 2024-08-29
Update docker.yml to include latest tag

jteijema:
#1815 is correct, there's no latest package. Maybe we should also add this to the 1.6 tag branch.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -52,5 +52,5 @@ jobs:
           context: .
           platforms: linux/amd64,linux/arm64
           push: ${{ github.event_name != 'pull_request' }}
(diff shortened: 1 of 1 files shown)


J535D165:
Thanks. 

The use of the latest tag is widely discussed and considered to be a bad practice. That's why I removed it before, but I'm happy to introduce it again.

jteijema:
@J535D165 I agree its bad practice for the sake of reproducibility (can't verify a version if latest is used, pipelines will break if latest is used, etc). But many MANY users take this standard and I think facilitating this is more important in this case.

J535D165:
https://github.com/asreview/asreview/pkgs/container/asreview new version with latest tag can be found in the registry.

Issue #1807, opened 2024-08-15
ASReview randomly chooses data column when multiple are available


PeterLombaers:
**Describe the bug**
ASReview needs to read titles, abstract, etc. from the user input data. It accepts multiple column names for this data. When multiple options are available, it will pick one depending on the order of the columns. This can lead to unexpected results, for example, the user thinks that the titles from the `title` column are used, but actually the `primary_title` column is used.

**Relevant Code**
The allowed column names are defined [here](https://github.com/asreview/asreview/blob/ebdd44fbee4c1a7bb3a1c9f0d448144cc3676ccf/asreview/config.py#L49). The logic for picking the column is [here](https://github.com/asreview/asreview/blob/ebdd44fbee4c1a7bb3a1c9f0d448144cc3676ccf/asreview/data/base.py#L194). 

**Solutions**
It would make sense to issue a warning when multiple columns for the same data type are detected. It would also make sense to not let the order of the input columns influence which column will get picked. I think that `title` should always have higher priority than `primary_title` and `abstract` should always have higher priority than `notes_abstract`.

PeterLombaers:
Related to https://github.com/asreview/asreview-datatools/issues/45

jteijema:
I think your solution makes sense. A custom priority order in which column is selected, and then providing the user with this order via the readthedocs, seems good to me.

PR #1820, opened 2024-08-29
Store user input data in database

PeterLombaers:
This pull request implements a database to store the user input data.

Currently we are using the `Dataset` class to represent the user input data. It reads the actual input file and performs cleaning and validation of the data on the go. This is innefficient because you need to read the complete file every time you want to get the data of a single record. Also, the logic for cleaning and validating the data is quite messy and it is difficult to customize the data format.

This pull request adds two main classes: `Record` and `DataStore`. The `DataStore` class gives a Python API to write data to an SQLite database and read data from it in the form of records, or pandas DataFrames. The `Record` class represent a single input record. It defines the fields that are allowed, how they should be validated and how they should be converted to Pandas data. The idea is that it should be possible to customize the `Record`, for example add different fields, and `DataStore` should still work with this custom record. I tried to put a lot of comments in the source code of the two classes to explain why I made the things the way they are.

Most of the tests are passing, but not all, and this PR is not a finished product. In particular, I didn't touch the readers in any way, and it still uses the `Dataset` class to convert input to `Record` objects. Also, I didn't look at exporting data at all, so I assume some things will go wrong there. I did try to remove all use of `Dataset` other than for import and export, so if you still see that somewhere else, that's a mistake.

Let me know what you think, and feel free to merge/change/build further with this PR!

Here is also a random list of TODO's, that I'll drop here in case someone wants to work more on this:
- [x] Streamline the data readers
- [x] Make the data readers use `Record.__alternative_column_names__` for the allowed columns
- [x] Make the data readers use the order of `Record.__alternative_column_names__` to the determine which column to read if there are multiple possible input columns that correspond with a record column (this would solve #1807)
- [x] Move `Record.dataset_row` and `dataset_id` to the `Base` record class? Then any custom implementation would be forced to have them.
- [x] Determine which classes should be put in the `__init__.py` of `asreview.data`
- [x] I'm using no connection pool at the moment. That was giving errors when cleaning up the database file during test. We probably want to use the connection pool, but properly close everything on exit?
- [x] Remove the `ASReviewData` object, or turn it into something that cleans the output from readers
- [x] Create a database constraint that either `title` or `abstract` should be present?
- [x] Are the record fields all we need, for example, do we need `tags`, `prior`?


Rensvandeschoot:
Great work!!! And these changes will open-up many new opportunities!

> Create a database constraint that either title or abstract should be present?

What will you do if both are missing in the dataset? Yes, the model will neither be able to process this record, nor a human can provide a label based on the title+abstract, but such records sometimes are part of the dataset. In a traditional pipeline, you would search for more information online and select a label based on that information. In our case, this record can not be processed. Will we test for it when users import their dataset and present a message/error that the data contains an _x _ amount of records without title+abstract? Or, will we simple remove such records from the data and continue with only those records that do have either a title and/or an abstract?

> Are the record fields all we need, for example, do we need tags, prior?

For post-processing, I think we do need the tags and used_as_prior in the exported data.

asreview/config.py
```diff
@@ -63,10 +63,6 @@
     "notes": ["notes"],
     "keywords": ["keywords"],
     "doi": ["doi"],
-    "tags": [
-        "tags",
-    ],
-    "is_prior": ["asreview_prior", "is_prior"],
```
J535D165, line 69: We removed the Validation/Exploration project type as an explicit project type and integrated Validation with the Exploration/Review project type. 

Therefore, it is fine to deprecate the is_prior column here.

J535D165:
See this interesting issue: https://github.com/asreview/asreview-datatools/issues/48. PR #1820 standardizes the input variables, so this should not be an issue. It can be a nice test case to see if everything works as expected.

tests/citation-file-formatting
J535D165, line 1: My bad. I will remove this.

PeterLombaers:
Locally all my tests pass. I think this is ready to merge. We could do more refactoring and renaming and maybe simplifying the logic in places, but I think that's better to do in new pull requests.

J535D165:
I fixed the docs with a workaround. It's a well-known issue with sqlalchemy and sphinx.

J535D165:
Massive impressive :)

PR #1835, opened 2024-09-05
Cleaner conversion from static colors to dynamic colors

BerkeYazan:
Converted the colors from static to dynamic in `ReviewProgress`, `LabelingHistory`, and `LabelingFrequency.`


J535D165:
Thanks!

PR #1860, opened 2024-09-24
Improve exported results and add support for exporting tags

J535D165:


J535D165:
This PR might have some conflicts with #1820, but should make it easier to finish the logic of #1820. I am happy to help resolve the conflicts.

PR #1861, opened 2024-09-25
Add queue that manages training of models and simulations

cskaandorp:
See /asreview/webapp/queue

start manager.py, and in a different terminal start tester.py


cskaandorp:
use `asreview lab` to start Task Manager automatically, config:

```
ASREVIEW_LAB_TASK_MANAGER_ENDPOINT=localhost:5555 asreview lab
```

Change number of workers:
```
ASREVIEW_LAB_TASK_MANAGER_ENDPOINT=localhost:5555 ASREVIEW_LAB_TASK_MANAGER_WORKERS=4 asreview lab
```

asreview/webapp/task_manager/models.py
```diff
@@ -0,0 +1,16 @@
...
+Base = declarative_base()
+
+
+class ProjectQueueModel(Base):
+    """Queue model"""
+
+    __tablename__ = "queue"
+    id = Column(Integer, primary_key=True)
+    project_id = Column(String(250), nullable=False, unique=True)
+    simulation = Column(Boolean, nullable=False, default=False)
```
J535D165, line 16: I foresee that we will use this queue for more tasks in the future (full text parsing or LLMs). It might be useful to use the project "modes" we are currently using for other things. 

```suggestion
    mode = Column(String(100), nullable=False, default="oracle")
```

asreview/webapp/task_manager/task_manager.py
```diff
@@ -0,0 +1,258 @@
...
+
+from asreview.webapp import asreview_path
+from asreview.webapp.task_manager.models import Base
+from asreview.webapp.task_manager.models import ProjectQueueModel
+from asreview.webapp.task_manager.task_wrapper import RunModelProcess
+from asreview.webapp.tasks import run_task
+
+
+class TaskManager:
+    def __init__(self, max_workers=1, host="localhost", port=5555):
```
J535D165, line 21:
```suggestion
    def __init__(self, max_workers=1, host="localhost", port=5555):
```

Shall we set the max_workers here also at 2? Of the other one at 1? It's nice to align. 

Port 5555 sounds like a port that might be popular for some other applications as well. Maybe use a less popular port as default?

asreview/webapp/task_manager/task_manager.py
```diff
@@ -0,0 +1,258 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--verbose", action="store_true", help="Enable verbose logging")
+    args = parser.parse_args()
+
+    setup_logging(verbose=args.verbose)
+
+    manager = TaskManager(max_workers=2)
+    manager.start_manager()
```
J535D165, line 258: This is almost an entry point. It might be useful to make this an entry point. Users can have difficulty finding this file when installed via pip (and the location changes with an upgrade of Python). There is `entry_points_internal` if you don't want users to see this entry point via `asreview -h`.

asreview/webapp/task_manager/task_wrapper.py
```diff
@@ -0,0 +1,25 @@
+import json
+import multiprocessing as mp
+import socket
+
+
+class RunModelProcess(mp.Process):
```
J535D165, line 6: Move this to `task_manager.py`? It makes it easier to share the port and host (now the default is hardcoded in 2 places).

asreview/webapp/task_manager/tester.py
```diff
@@ -0,0 +1,30 @@
...
+            client_socket.sendall(json.dumps(payload).encode())
+            # time.sleep(0.05)  # Sleep between messages
+
+        print("Closing the connection.")
+        client_socket.close()
+
+
+if __name__ == "__main__":
+    for client_id in range(5):
+        threading.Thread(target=send_payloads, args=(client_id,)).start()
```
J535D165, line 30: Can we make this into a pytest or is that complicated? External devs won't expect this test here.

pyproject.toml
```diff
@@ -85,9 +85,6 @@ auth-tool = "asreview.webapp.entry_points.auth_tool:AuthTool"
 simulate = "asreview.simulation.cli:_cli_simulate"
 algorithms = "asreview.models.cli:cli_algorithms"
 
-[project.entry-points."asreview.entry_points_internal"]
-web_run_model = "asreview.webapp.entry_points.run_model:main"
-
```
J535D165, line 90: Add the task manager here.

J535D165 (changes requested):
Very nice PR and nicely implemented. It looks very reliable to me. 

I added some (small) points of feedback above. We also need the following things: 

- Instructions for developers in the DEVELOPMENT.md
- Instructions for server users in the docs
- Update the VSCode `tasks.json` file linked in the DEVELOPMENT.md (I will do that).

asreview/webapp/api/projects.py
```diff
@@ -96,6 +98,42 @@ def _fill_last_ranking(project, ranking):
...
+            client_socket.connect((host, port))
+            # setup payload
+            project_id = project.config["id"]
+            payload = {
+                "action": "insert",
+                "project_id": project_id,
+                "simulation": simulation,
+            }
+            # send
+            client_socket.sendall(json.dumps(payload).encode("utf-8"))
```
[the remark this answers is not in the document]
cskaandorp: Yes it is, we need to send bytes, not strings. And the utf-8 is a nice standard.

asreview/webapp/app.py
```diff
@@ -164,4 +164,28 @@ def index(**kwargs):
...
+            }
+        else:
+            message = (
+                "Task manager configuration is mandatory in "
+                + "both development and production environments. Please "
+                + "pass the endpoint for the task manager in the "
+                + "ASREVIEW_TASK_MANAGER_ENDPOINT environment variable."
+            )
+            raise RuntimeError(message)
+
```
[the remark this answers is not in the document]
cskaandorp: Will do

asreview/webapp/entry_points/lab.py
```diff
@@ -177,6 +179,19 @@ def lab_entry_point(argv):
...
+    socket_port = app.config["TASK_MANAGER_CONFIG"]["port"]
+    verbose = app.config["TASK_MANAGER_CONFIG"]["verbose"]
+
+    process = multiprocessing.Process(
+        target=run_task_manager, args=(workers, socket_host, socket_port, verbose)
+    )
+    process.start()
+    if process.pid is None:
+        raise RuntimeError("Not able to spin up the task manager.")
+
```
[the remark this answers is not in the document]
cskaandorp: Done

asreview/webapp/task_manager/task_manager.py
cskaandorp: Fully agree, port 5101 and 2 workers

asreview/webapp/task_manager/tester.py
cskaandorp: This was not supposed to be here, I used to to quickly dump some data in the task manager

pyproject.toml
cskaandorp: I've kept asreview.entry_points_internal since it provides a place where more of these "hidden" entry-points can exist. We never know what the future may bring. I like the generic name over a specific name.

cskaandorp:
Still need to add some arguments to task manager entry points call

J535D165 (approved):

J535D165:
Thanks a lot for all the work you put in this. I'm very happy with the result. It sounded like a good idea to use huey, but unfortunately it didn't fit our application well. Your work is a bit step in making asreview more reliable, powerful, and scalable. Thanks again!

PR #1873, opened 2024-10-08
Add roles (frontend) to replace owner_id

J535D165:
This PR replaces the user_id variable in the project config (webapp only) with roles. The role defines what the frontend renders. 

The structure of the roles is simple, but extensible. @cskaandorp will extend this in the future. 

```json
"roles": {"owner": true}
```
```json
"roles": {"owner": false}
```

For example, on the Settings page, an owner sees the delete button, whereas other users won't see this button. 

The backend currently isn't using roles, but will in the future.


J535D165:
FYI @cskaandorp

PR #1884, opened 2024-10-17
Remove use of LABEL_NA from internals

J535D165:
The plan is to deprecate LABEL_NA completely.


J535D165:
We might need several powerful tests to ensure that everything is working correctly. For example,`run_model` has not been extensively tested.

PR #1887, opened 2024-10-17
Bump minimal Python version to 3.10

J535D165:


PeterLombaers:
Looks good to me! Python3.10 seems like a good balance between making life easier for the developers, whilst not forcing too many people to install a new version of Python.

This probably also affects datatools, insights, etc. But then I assume that many changes in v2 will affect them, so they'll need changes anyway.

PeterLombaers (approved):

J535D165:
I removed the `asreview-wordcloud` test, as we might want to integrate or deprecate this package with the release of v2. The package is limited to users who can run the CLI at the moment, and they will most likely also be able to render a word cloud in Python. Users who aren't easy with the CLI will most likely benefit from integration in Asreview LAB.

Rensvandeschoot:
I agree! We created the Wordcloud package many years ago and it is nowadays much easier to create such a workcloud. 

We should not forget to also remove it from the makita workflow!

PR #1900, opened 2024-10-22
Fix Windows tests on delete project

J535D165:


asreview/project/api.py
```diff
@@ -420,8 +420,10 @@ def add_review(
         if state is None:
             state = SQLiteState(fp_state)
             state.create_tables()
+            state.close()
         else:
             state.to_sql(fp_state)
+            state.close()
```
PeterLombaers, line 426: Here you close a state object that was provided in the arguments. If I do `project.add_review(state)`, afterwards `state` is closed.

J535D165: Yeah thanks, nice side effect to find this bug as well. I changed the argument name.

J535D165:
Long waited for :)

PR #1901, opened 2024-10-23
Fix importing oauth parameters in front-end

cskaandorp:
Adding OAuth configuration led to an issue when rendering index.html


J535D165 (approved):

J535D165:
Nice and welcome

PR #1922, opened 2024-11-07
Show keywords when available

J535D165:
image


Rensvandeschoot:
Very nice!

What field of the RIS file format are you using as input?

J535D165:
RIS will use `KW`.

PeterLombaers:
I like the look of it! One thought I had was that if the screen is small or the abstract is long, the keywords won't be visible immediately. If they came before the abstract, they would always be visible. But they might also be distracting above the abstract, so I'm not sure if that would really be an improvement.

J535D165:
True and thanks for the suggestion. I discussed this with Cathelijne, and she favored having them at the bottom. But maybe I can make a setting for this.

PR #1929, opened 2024-11-14
Refactor Feature Extractor

PeterLombaers:
This pull request refactors the feature extractor class. The goal was to make it easier to use the Python API, and also make it easier to create extension overriding specific parts of the feature extraction process.

### DIFF
--- a/asreview/data/record.py
+++ b/asreview/data/record.py
@@ -118,10 +118,10 @@ class Record(Base):
 
     @validates("title", "abstract", "doi", "url")
     def validate_string(self, key, value):
-        if value is None:
-            return ""
-        if not isinstance(value, str):
-            raise ValueError(f"'{key}' should be a string, but is: {value}")
+        if value == "":
+            return None
+        if value is not None and not isinstance(value, str):
+            raise ValueError(f"'{key}' should be a string or None, but is: {value}")
         return value
 
     @validates("included")
--- a/asreview/data/record.py
+++ b/asreview/data/record.py
@@ -98,21 +98,31 @@ class Record(Base):
     title: Mapped[str] = mapped_column(default="")
     abstract: Mapped[str] = mapped_column(default="")
     # authors and keywords could also be in their own separate table.
-    authors: Mapped[list] = mapped_column(default=None)
+    authors: Mapped[list] = mapped_column(default_factory=list)
     notes: Mapped[Optional[str]] = mapped_column(default=None)
-    keywords: Mapped[list] = mapped_column(default=None)
+    keywords: Mapped[list] = mapped_column(default_factory=list)
     year: Mapped[Optional[int]] = mapped_column(default=None)
-    doi: Mapped[Optional[str]] = mapped_column(default=None)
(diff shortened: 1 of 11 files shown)


asreview/models/feature_extraction/tfidf.py
```diff
@@ -42,22 +42,9 @@ class Tfidf(BaseFeatureExtraction):
...
-            sklearn_stop_words = None
-        else:
-            sklearn_stop_words = self.stop_words
-        self._model = TfidfVectorizer(
-            ngram_range=(1, ngram_max), stop_words=sklearn_stop_words
+    def __init__(self, stop_words="english", **kwargs):
+        super().__init__(
+            sklearn_model=TfidfVectorizer,
+            stop_words=stop_words,
+            **kwargs,
```
J535D165, line 49: Nice

asreview/data/record.py
```diff
@@ -98,21 +98,31 @@ class Record(Base):
...
-    authors: Mapped[list] = mapped_column(default=None)
+    authors: Mapped[list] = mapped_column(default_factory=list)
     notes: Mapped[Optional[str]] = mapped_column(default=None)
-    keywords: Mapped[list] = mapped_column(default=None)
+    keywords: Mapped[list] = mapped_column(default_factory=list)
     year: Mapped[Optional[int]] = mapped_column(default=None)
-    doi: Mapped[Optional[str]] = mapped_column(default=None)
-    url: Mapped[Optional[str]] = mapped_column(default=None)
+    doi: Mapped[Optional[str]] = mapped_column(default="")
+    url: Mapped[Optional[str]] = mapped_column(default="")
```
J535D165, line 106: See other comment on the test.

tests/test_readers.py
```diff
@@ -121,12 +121,12 @@ def test_nan_values_ris():
...
-    assert records[1].title is None
-    assert records[3].title is None
+    assert records[1].title == ""
+    assert records[3].title == ""
 
     # Check missing abstracts
-    assert records[0].abstract is None
-    assert records[2].abstract is None
+    assert records[0].abstract == ""
+    assert records[2].abstract == ""
```
J535D165, line 129: I don't fully understand where this change comes from (as there are no changes to the title and abstract in the record file, isn't it?). I think it has many consequences on the front end. 

In my opinion, it should be `None` when there is no value. This makes many downstream tasks more intuitive, easier, and maintainable.

PeterLombaers: In `record.py:L120` I added a validator on the record that tests if certain fields are a string, and this validator also transforms `None` to `""`. That's where the change comes from.

My thought was that it would be easier in the backend to always be certain that title and abstract are strings. For example, when creating the feature matrix I can just concatenate the values for title and abstract, I don't need to fill the missing values with an empty string.

But I'm fine with changing missing values to `None`. The only thing I would like to do then is replace `""` by `None`, so that you don't get a mix of `None` and `""` for missing values.

J535D165: Thanks, let's go for `None` then.

PR #1934, opened 2024-11-19
Continue refactor the labeling frequency

BerkeYazan:
### Performance optimizations:

- Added useMemo and useCallback hooks to memoize calculations and event handlers, reducing unnecessary re-renders.
- Wrapped the LabelingFrequency component with React.memo to prevent re-rendering when props have not changed.
- Simplified the math to work better in large number of records

### User interface improvements:

- Introduced skeleton loading.

- Fixed the slider bugs.

- Fixed colors.

### DIFF
--- a/asreview/webapp/src/ProjectComponents/AnalyticsComponents/LabelingFrequency.js
+++ b/asreview/webapp/src/ProjectComponents/AnalyticsComponents/LabelingFrequency.js
@@ -1,10 +1,4 @@
-import React, {
-  useState,
-  useEffect,
-  useRef,
-  useMemo,
-  useCallback,
-} from "react";
+import React, { useState, useEffect, useRef } from "react";
 import {
   Card,
   CardContent,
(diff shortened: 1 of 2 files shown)


J535D165:
Thanks for the PR. 

General comment about the re-renders (nice you found them): The re-renders are, most likely, caused by the inefficient use of react-query in the Analytics page. Instead of passing `genericDataQuery` and `progressQuery` to components, it is better to query the API again via react-query. The results are efficiently cached and fast.  

This means that 

```javascript
  const progressQuery = useQuery(
    ["fetchProgress", { project_id }],
    ({ queryKey }) =>
      ProjectAPI.fetchProgress({
        queryKey,
      }),
    { refetchOnWindowFocus: false },
  );
  // New unified query for fetching data
  const genericDataQuery = useQuery(
    ["fetchGenericData", { project_id }],
    ({ queryKey }) =>
      ProjectAPI.fetchGenericData({
        queryKey,
      }),
    { refetchOnWindowFocus: false },
  );
```

should be added to each component that makes use of it. This will also make the components easier to reuse and prevent the re-renders. 

When this is added to each component, the use of useMemo and useCallback won't be useful anymore as the component will be rendered only once (unless the users makes a change and re-render is really needed). 
Besides this, the React team will introduce a new compiler that makes useMemo and useCallback obsolete (https://javascript.plainenglish.io/react-usememo-hook-is-going-to-be-obsolete-soon-7fed40d4d511). So, we should look at the proper use of react-query over using useMemo and useCallback.

BerkeYazan:
Thanks for the feedback. I was also refactoring other components with useMemo, it's good that we shifted to this approach early on. I'll implement the necessary changes!

asreview/webapp/src/ProjectComponents/AnalyticsComponents/LabelingFrequency.js
```diff
@@ -171,18 +190,18 @@ const LabelingFrequency = ({ genericDataQuery, progressQuery }) => {
...
               Learn more
-            </a>
+            </Link>
           </Box>
         </Box>
       </Popover>
     </Card>
   );
-};
+});
```
J535D165, line 205:
```suggestion
};
```

J535D165 (changes requested):
Thanks. Let's merge soon. The useMemo is unnecessary (after we refactor the other components in another PR).

PR #1938, opened 2024-11-25
Replace long-lived tokens with short-lived OTP tokens

cskaandorp:
User account creation verifies email addresses with a token that is valid for 24 hours. This is also used in "forgot password": a user gets a 24-hour valid token by email in the form of a link. 
Instead of that, this PR provides a 6-digit one-time-password that is used to verify an email address or allows to change a password.


J535D165:
First impressions look good to me. Do you also plan to make frontend changes in this PR? We can also do that in another PR.

cskaandorp:
I'd like to do this, won't take too much time, can do it in this PR but if you want to organize things differently let me know. I just read your comment after I submitted some frontend work. Let me know if you are OK with this PR.

J535D165:
Let me know when I can start the review (by removing the draft mode label). Ideally with all checks on green.

J535D165 (approved):

PR #1940, opened 2024-11-28
Use `Parameters` instead of `Arguments` in docstrings

### LINKED ISSUE #1911: Standardize docstrings
In function docstrings, for the section describing the input arguments, we use both `Arguments` and `Parameters`. I think `Arguments` is more common, but `Parameters` occurs a couple of times. We should pick one so that automatically generating documentation is easier.

PeterLombaers:
This PR standardizes the docstrings to always use `Parameters` instead of `Arguments`. Closes #1911


J535D165:
Appreciated

PR #1941, opened 2024-11-28
Implement remote user authentication

dometto:
This PR implements support for remote user authentication to the ASReview webapp. This mechanism outsources all authentication to the webserver or middleware that is placed in front of the application in the stack. This allows one to leverage e.g. Nginx or Apache's modules for things like LDAP or OIDC, thereby enabling ASReview to use an organisation's own user directories or authentication schemes. This works as follows:

1. Webserver performs authentication, and if successful, reverse proxies to an ASReview server (typically on localhost). It adds http headers to the request that indicate the user's username, email, etc. For instance, it adds the header: `Remote-User: dometto`
2. ASReview checks for the presence of these headers. If they are found, login is performed for the found user `dometto`, which is created in the database if necessary.

Many other applications import this form of authentication configuration. Read more about it [here](https://docs.galaxyproject.org/en/latest/admin/authentication.html#remote-user-authentication) , for example.

This PR includes tests, and you can test the functionality manually by:

1. running the app with the new example config file `auth_remote_config.toml`
2. Visit `http://localhost:3000/`: you won't be authenticated!
4. Visit `http://localhost:3000/`, but pass along the following header: `'My-User-Identifier-Header': 'foo'`. Now you'll be authenticated!
    * you can use [a browser extension](https://stackoverflow.com/questions/374885/can-i-change-the-headers-of-the-http-request-sent-by-the-browser) to pass along the header
    * or use `curl -H "My-User-Identifier-Header: foo" http://localhost:3000/`

NB: since this mechanisms basically disables all checks on authentication and just 'trusts' the provided headers, it is imperative that it is only used with correctly configured webservers/middleware! This should be made very clear to the enduser.


dometto:
@J535D165 I've added tests and some basic documentation, but I'm sure this can still be improved. I thought I'd open this PR to get some first feedback though! Let me know if you prefer another way of working.

DEVELOPMENT.md
```diff
@@ -70,12 +70,12 @@ When using or developing the authenticated version of ASReview, extra steps
 are needed to configure the application.
 
 Create an authentication config file as instructed in [Authentication]
-(#Authentication). Set the environment variable `FLASK_CONFIGFILE` to the
+(#Authentication). Set the environment variable `ASREVIEW_LAB_CONFIG_PATH` to the
```
dometto, line 73: `FLASK_CONFIGFILE` didn't work for me, so I updated the docs, but maybe it's just me. :)

J535D165: You are right! `ASREVIEW_LAB_CONFIG_PATH ` it is.

asreview/webapp/authentication/decorators.py
```diff
@@ -81,3 +90,32 @@ def decorated_function(*args, **kwargs):
...
+                if not user:
+                    try:
+                        user = User(
+                            **user_info, origin="remote", public=True, confirmed=True
+                        )
+                        DB.session.add(user)
+                        DB.session.commit()
+                    except (IntegrityError, SQLAlchemyError) as e:
+                        DB.session.rollback()
+                        error_500(e)
```
dometto, line 116: This helper is available in `asreview.webapp.app`, but not in this module or the API. Should `error_500` be moved to a common utility module?

asreview/webapp/tests/test_api/test_remote_auth.py
```diff
@@ -0,0 +1,63 @@
...
+    def get_uri(path):
+        return client_remote_auth.get(uri, follow_redirects=True, **custom_headers)
+
+    def find_user(identifier):
+        return User.query.filter(User.identifier == identifier).one_or_none()
+
+    user = find_user(user_identifier)
+    assert user is None
+
+    with pytest.raises(TemplateNotFound):
```
dometto, line 37: Hitting `/` or `/signup` while testing locally throws a template error, presumably because the templates are not setup in this context. But in the `compile_and_test` workflow ([here](https://github.com/asreview/asreview/actions/runs/12072095964/job/33777599102)) the templates are loaded, so this test fails. Any advice on how to handle this?

asreview/webapp/authentication/decorators.py
J535D165: That makes sense indeed!

asreview/webapp/tests/test_api/test_remote_auth.py
J535D165: I'm looking into this. I don't see what's going on at the moment.

J535D165:
Thanks for the PR and extensive documentation, @dometto! This is a fantastic contribution and makes many organizations happy. 

@cskaandorp and @PeterLombaers, can you review this PR?

J535D165:
I can't wait to see a demo with Surf Research Cloud btw.

asreview/webapp/app.py
```diff
@@ -101,11 +103,13 @@ def load_user(user_id):
         with app.app_context():
             # create tables in case they don't exist
             DB.create_all()
-
         # store oauth config in oauth handler
         if bool(app.config.get("OAUTH", False)):
             app.config["OAUTH"] = OAuthHandler(app.config["OAUTH"])
 
+        if bool(app.config.get("REMOTE_USER", False)):
```
PeterLombaers, line 110: You can configure both `REMOTE_USER` and `OAUTH`. Does it ever make sense to have both configured? If not, maybe it makes sense not to allow configuring both, then there is less of a chance of mistakes happening.

asreview/webapp/authentication/decorators.py
```diff
@@ -81,3 +90,38 @@ def decorated_function(*args, **kwargs):
...
+                                confirmed=True,
+                            )
+                            DB.session.add(user)
+                            DB.session.commit()
+                        except (IntegrityError, SQLAlchemyError) as e:
+                            DB.session.rollback()
+                            error_500(e)
+
+                    login_user(
+                        user, remember=True, duration=datetime.timedelta(days=31)
```
PeterLombaers, line 123: Should the duration be configurable?

asreview/webapp/authentication/remote_user_handler.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class RemoteUserHandler:
+    default_headers = {
+        "USER_IDENTIFIER_HEADER": "REMOTE_USER",
+        "USER_NAME_HEADER": False,
+        "USER_EMAIL_HEADER": False,
+        "USER_AFFILIATION_HEADER": False,
+        "DEFAULT_EMAIL_DOMAIN": "localhost",
+        "DEFAULT_AFFILIATION": "",
```
PeterLombaers, line 23: Maybe make the default value `None`. Now a missing identifier will be returned as `None` but a missing affiliations is returned as `""`.

PeterLombaers (commented):
I can't say I know very much about authentication, but this looks like a nice addition! I also like the explanation, makes it very understandable.

I just made a few comments/questions, but they're all a bit nitpicky, so feel free to skip them and focus on getting something working for now.

asreview/webapp/app.py
dometto: I was also wondering about this. In principle, yes, you could allow multiple authentication methods on a single instance of the app. For instance, the webserver could set or not set `REMOTE_USER` depending on the IP address of the client, so that for a certain (say, internal) IP range remote user auth is allowed, but for different IP ranges, users still need to login using regular DB auth or OAuth.

However, I do think this use case will be rare, and the benefit of disallowing remote user auth if any other form of authentication is enabled is that it reduces the chance of misconfiguration.

For what it's worth, Flask-Appbuilder [forces](https://flask-appbuilder.readthedocs.io/en/v3.2.0/security.html#authentication-methods) the user to choose a single authentication type. Perhaps we should go with that approach, and change when it turns out there is demand for mixing auth methods?

dometto: On the other hand: at present there is no configuration setting that enables/disables ordinary password authentication. So remote auth and password auth will always technically co-exist. Maybe then it makes sense for both of these to co-exist with OAuth, as well?

Alternatively, I think a bigger refactor should be implemented that requires the user to set, exclusively, either password authentication, OAuth authentication, or remote user authentication.

asreview/webapp/authentication/decorators.py
dometto: :+1: I've had to refactor the `perform_login_user` function for this.

asreview/webapp/authentication/decorators.py
J535D165: You seem to be right. It looks like Flask has simplified the error handling https://flask.palletsprojects.com/en/stable/errorhandling/. Feel free to change it in this PR if that helps.

dometto: Done.

asreview/webapp/tests/test_api/test_remote_auth.py
dometto: Any luck checking in to this?

dometto:
Taking some inspiration from [Galaxy](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/web/framework/middleware/remoteuser.py#L106-L118), I've added a `REMOTE_AUTH_SECRET` config option. This adds some (light) security against users who can approach the app directly (e.g. on localhost), without any middleware that performs authentication.

asreview/webapp/app.py
PeterLombaers: That use case for multiple authentication methods sounds a bit exotic, and I do think there is value in allowing only a single authentication method. Besides misconfiguration, the code should also be a bit easier to understand if you can have only a single authentication method enabled.

But there's no need to do that right away. I'll make an issue so we don't forget and then you can just continue without too much trouble with this PR.

PeterLombaers (approved):

dometto:
I've fixed the linting issue with the documentation. This leaves the [issue with templating](https://github.com/asreview/asreview/pull/1941#discussion_r1865579387). Any help with that is very welcome :)

asreview/webapp/tests/test_api/test_remote_auth.py
dometto: I've simply added a `try...except` for the `TemplateError` now, so *when* the error is produced the test should just pass, and when it isn't, also. :)

This assumes the templates are not loaded locally due to the way the API tests are configured, so the error is expected in that case.

dometto: So I *think* running the CI tests might yield green lights. :)

dometto:
@J535D165 could you run the CI tests for this again please? 🙏

J535D165:
Thanks @dometto. All checks on green. Great. What do you think? Merge (and refine later if needed)?

J535D165:
Thanks a lot for this work, we can't wait to experiment with it and see your live demo!

PR #1949, opened 2024-12-04
Fix loader animation in reviewer screen

J535D165:
This is a first attempt to fix the "Warmup screen" that randomly pops up on the reviewer page. It is a challenging issue. 

It's hard for me to reproduce this issue on my device/environment. However, we see it more often on the test server. It might have something to do with concurrency to our SQLite database. Maybe @PeterLombaers can think with me about this. 

Basically, what happens:
- Sometimes, querying the ranking table results in an empty table
- Shortly after, the same request will succeed with a filled table. 

I think this has something to do with updating the ranking table from the other thread. However, we haven't seen this before, so that confuses me. In this PR, I merged two sqlite queries (select and insert) to avoid edge cases with concurrency.

### DIFF
--- a/asreview/state/sqlstate.py
+++ b/asreview/state/sqlstate.py
@@ -225,12 +225,9 @@ class SQLiteState:
                 f"{list(RANKING_TABLE_COLUMNS_PANDAS_DTYPES.keys())}."
             )
 
-        cur = self._conn.cursor()
-        cur.execute("delete from last_ranking")
-        self._conn.commit()
-        cur.close()
-
-        last_ranking.to_sql("last_ranking", self._conn, if_exists="append", index=False)
+        last_ranking.to_sql(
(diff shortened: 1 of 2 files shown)


PeterLombaers:
My first thought would be that in [the ranking table update](https://github.com/asreview/asreview/blob/6c66bb8a78816581b925edf1e2d261debe8cda27/asreview/state/sqlstate.py#L221), deleting the existing data and adding the new data do not happen in the same transaction. There is a `conn.commit()` in the middle. If the ranking table gets queried after the old data was deleted, but before it was filled again, the result will be empty.

J535D165:
That's a good one. I think pandas can replace the table directly, so I don't know why I/we didn't do that before. Will implement that.

J535D165:
Closer inspection shows that this function is only used in `migrate.py`. I resolved the connection issue you mention, but it won't solve our issue.

PR #1955, opened 2024-12-05
Add `duplicate_of` column to `Record` class

PeterLombaers:
This pull request prepares for marking records as duplicate inside ASReview. It adds a column `duplicate_of` to the `Record` class. This column can contain a foreign key to the duplicate of a record. Any logic concerning how we actually fill this column is not part of this pull request and needs further discussion.

### DIFF
--- a/asreview/__init__.py
+++ b/asreview/__init__.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from sqlalchemy import event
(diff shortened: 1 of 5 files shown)


J535D165 (approved):

PeterLombaers:
There is still a test that fails because I enabled sqlite to use foreing keys. Apparently a foreign key constraint in a test of webapp is violated? I could look into only enabling foreign keys for the data_store and not for the user database.

PR #1972, opened 2024-12-12
Replace balance strategies by sample weight

PeterLombaers:
This pull request changes the balancers. Instead of sampling the training data, the balancer now computes a vector of sample weights. This is passed as an argument to `classifier.fit`. The hope is that this performs well enough that we do not need to use any other balancing. 

Currently the default ratio of `1.0` is used in the `Balanced` balancer. With this ratio the behavior is the same as passing `"balanced"` in `sklearn.utils.compute_sample_weight`. Some preliminary testing (n=1) seems to indicate that a ratio of `15.0` performs better, and is on par with the current asreview performance.

Some tests still fail because they import historic asreview files which still contain mentions of deprecated balance strategies. So we still need to consider how to read/convert these files.


jteijema:
Looks good! I look forward to using the class weight functionality of the classifiers themselves. Thanks! Ill make sure to work this into Makita/NEMO

jteijema:
Are there other balancing methods that might work with this new format aside from what the balancer does in this PR?

J535D165:
> Are there other balancing methods that might work with this new format aside from what the balancer does in this PR?

Most likely yes!

J535D165:
It was fun to work on this with you!

PR #1993, opened 2024-12-19
Fix dataset PRAGMA error on non SQLite webapp database

J535D165:


PeterLombaers (approved):
Thanks for fixing this!

asreview/state/sqlstate.py
```diff
@@ -89,7 +89,7 @@ def _conn(self):
         if hasattr(self, "_conn_cache"):
             return self._conn_cache
 
-        self._conn_cache = sqlite3.connect(str(self.fp))
+        self._conn_cache = sqlite3.connect(f"sqlite:///{self.fp}")
```
J535D165, line 92:
```suggestion
        self._conn_cache = sqlite3.connect(str(self.fp))
```

Issue #1999, opened 2025-01-09
Doc2Vec and Neural Net options are not installable in v1.6.3


irm-codebase:
**Describe the bug**
According to [this page in the docs](https://asreview.readthedocs.io/en/latest/project_create.html), it's recommended to re-run a screening a second time using fancier models. I'm attempting to do so with the recommended settings (doc2vec plus neural net).

However, attempting to install the necessary `asreview` extensions with the `tensorflow` or `gensim` options does not work in v1.6.3. There appears to be no way to get them to run with it.

**To Reproduce**

Here are three ways to trigger the problem(s):

- Running  `pip install asreview[gensim]` will tell you that the extra does not exist.

```
Pip subprocess error:
WARNING: asreview 1.6.3 does not provide the extra 'gensim'
WARNING: asreview 1.6.3 does not provide the extra 'gensim'
WARNING: asreview 1.6.3 does not provide the extra 'gensim'
```

- Running `pip install asreview[tensorflow]` or `pip install asreview[all]` results in a failure due to invalid `scikit-learn` naming:

```
Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [15 lines of output]
      The 'sklearn' PyPI package is deprecated, use 'scikit-learn'
      rather than 'sklearn' for pip commands.
      
      Here is how to fix this error in the main use cases:
      - use 'pip install scikit-learn' rather than 'pip install sklearn'
      - replace 'sklearn' by 'scikit-learn' in your pip requirements files
        (requirements.txt, setup.py, setup.cfg, Pipfile, etc ...)
      - if the 'sklearn' package is used by one of your dependencies,
        it would be great if you take some time to track which package uses
        'sklearn' instead of 'scikit-learn' and report it to their issue tracker
      - as a last resort, set the environment variable
        SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL=True to avoid this error
      
      More information is available at
      https://github.com/scikit-learn/sklearn-pypi-package
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

**Screenshots**

[...]

PR #2003, opened 2025-01-10
Update project_create.rst with correct doc2vec optional

jteijema:


J535D165 (approved):

jteijema:
As a response to #1999

PR #2010, opened 2025-01-18
Fix docker installation instruction

george-gca:
Fix docker installation instruction, since `latest` tag [doesn't exist](https://github.com/asreview/asreview/pkgs/container/asreview).


J535D165:
Thanks, @george-gca, for this contribution and for pointing out that the latest tag was broken. GitHub is sometimes hard to work with regarding Docker images. I managed to fix the missing latest tag https://github.com/asreview/asreview/pkgs/container/asreview. Now it should work again as described in the documentation. So, I will close this PR, and thank you again!

PR #2016, opened 2025-01-21
Fix uncontrolled error when adding tags from settings page

J535D165:


asreview/webapp/src/ProjectComponents/SetupComponents/ModelCard.js
```diff
@@ -438,7 +438,7 @@ const ModelCard = ({
     },
   );
 
-  console.log(mode);
+  // console.log(mode);
```
J535D165, line 441:
```suggestion
```

PR #2018, opened 2025-01-22
Replace deprecated login_required decorator and rename LOGIN_DISABLED

cskaandorp:
In this PR I have tried to:

* Reintroduce our custom login_required decorator. Due to the removal of Flask-login's LOGIN_DISABLED I think I can conclude that when an app uses Flask-login it will be authenticated in the future. Our app can do both. Our version of the login_required decorator allows us to continue this feature.

* Unify the authentication parameter in back and frontend under the same variable name. Since LOGIN_DISABLED will be removed, I have introduced the AUTHENTICATION paramater that turns authentication on or off.

* ~Create a more robust way for developers to start the app in development mode. React's .env.development file is completely independent from the backend parameters which may cause issues when developing (think authentication yes or no). The proposal tries to read a toml config file in the root directory and creates the .env.development on the basis of the found parameters. If the toml file can not be found, it produces a default  .env.development file.~ Moved to another PR.


cskaandorp:
None of it has been tested by our unit tests. But I would like to know if I am heading in the correct direction.

asreview/webapp/tests/config/auth_basic_config.toml
```diff
@@ -2,5 +2,5 @@ TESTING = true
 DEBUG = true
 SECRET_KEY = "my_very_secret_key"
 SECURITY_PASSWORD_SALT = "my_salt"
-LOGIN_DISABLED = false
+AUTHENTICATION = true
```
J535D165, line 5: I would also love to test the case where AUTHENTICATION is not set. This case should result in an authenticated version of ASReview.

J535D165:
Merge conflicts have been caused by some recent renamings. I think they can be easy to resolve. Will help with that later.

asreview/webapp/package.json
```diff
@@ -24,6 +24,7 @@
     "react-router-dom": "^6.26.1",
     "react-scripts": "^5.0.1",
     "react-share": "^5.1.1",
+    "toml": "^3.0.0",
```
J535D165, line 27: Move to the other PR

PR #2026, opened 2025-01-24
Replace SVC with LinearSVC

timovdk:
Since we are using SVC with a linear kernel, we might as well use the [linear optimized implementation](https://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVC.html)


J535D165:
Can you bump the boundary of the test?

J535D165 (approved):

PR #2028, opened 2025-01-24
Remove int cast for max_features in RF Classifier

### LINKED ISSUE #2027: Random Forest max_features is incorrectly cast to int
The `max_features` parameter is cast to an integer, but according to the [SKlearn documentation ](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html), `max_features` can be `str`, `int`, `float`.

timovdk:
`max_features` can be of types `int, float, str`: [RandomForestClassifier SKLearn Documentation](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html)
Closes #2027


J535D165:
Thanks!

PR #2041, opened 2025-01-30
Add models with parameters for V2 version

J535D165:
PR for models and their params.


timovdk:
What does the u2 vs u4 indicate?

asreview/models/models.py
```diff
@@ -16,18 +16,21 @@
...
 MODELS_CONFIG = [
     {
-        "name": "elas_u4",
-        "label": "ELAS u4",
+        "name": "elas_u2",
+        "label": "ELAS u2",
         "value": ActiveLearningCycleData(
             querier="max",
             classifier="nb",
+            classifier_param={"alpha": 3.822},
```
timovdk, line 24: Why are we using the 1.6 default for alpha here?

asreview/models/models.py
```diff
@@ -16,18 +16,21 @@
 
 MODELS_CONFIG = [
     {
-        "name": "elas_u4",
-        "label": "ELAS u4",
+        "name": "elas_u2",
+        "label": "ELAS u2",
```
timovdk, line 20: What do u4 and u2 refer to?

PR #2044, opened 2025-02-03
Add Normalized Discounted Cumulative Gain Metric

### LINKED ISSUE #2031: Implement Normalized Discounted Cumulative Gain metric
Implementing the Normalized Discounted Cumulative Gain as a performance metric in ASReview https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG would be nice. This algorithm penalizes relevant records lower in the result list.

timovdk:
Implements the NDCG metric
Closes #2031


asreview/metrics.py
```diff
@@ -39,3 +39,30 @@ def loss(labels: list[int]):
     return float(
         (Ny * (Nx - (Ny - 1) / 2) - np.cumsum(labels).sum()) / (Ny * (Nx - Ny))
     )
+
+
+def ndcg(labels: list[int]):
+    """Compute the Normalized Discounted Cumulative Gain (NDCG) https://doi.org/10.48550/arXiv.1304.6480 for binary labels.
```
J535D165, line 45: Can you keep the title short? See PEP 257.

asreview/metrics.py
```diff
@@ -39,3 +39,30 @@ def loss(labels: list[int]):
...
+
+    def dcg_binary(binary_labels):
+        gains = binary_labels  # Gains are the binary labels
+        discounts = 1.0 / np.log2(np.arange(2, len(binary_labels) + 2))
+        return np.sum(gains * discounts)
+
+    ideal_labels = np.sort(labels)[::-1]  # Ideal ranking (sorted labels)
+    actual_dcg = dcg_binary(labels)  # Use binary DCG
+    ideal_dcg = dcg_binary(ideal_labels)  # Use binary DCG for ideal
+
```
J535D165, line 67: I prefer to compute it without the extra function. 

```python
Ny = sum(labels)
Nx = len(labels)
if Ny == 0 or Nx == Ny:
    raise ValueError("Labels must contain two distinct classes.")

dcg = np.sum(labels * / np.log2(np.arange(2, Nx + 2)))
idcg = np.sum(1 * / np.log2(np.arange(2, Ny + 2)))

ndcg = float(dcg/idcg)
```

This might also fix the test.

J535D165 (changes requested):
Thanks for the implementation. I'm curious about this metric. 

Can you also add this metric to the simulation output `Simulate`?

I wonder how this metric performs between datasets. The metric is perfect for comparing prior knowledge and models on a single dataset, but I want to know what happens if you compare datasets of various sizes and inclusion percentages. 

Consider for example

```python
ndcg([1,1,1,0,1,0,0,0]
```
and 
```python
ndcg([1,0,1,0,0]
```

Both cases are 1-off the best solution, but the metric differs, isn't it?

timovdk:
> Thanks for the implementation. I'm curious about this metric.
> 
> Can you also add this metric to the simulation output `Simulate`?
> 
> I wonder how this metric performs between datasets. The metric is perfect for comparing prior knowledge and models on a single dataset, but I want to know what happens if you compare datasets of various sizes and inclusion percentages.
> 
> Consider for example
> 
> ```python
> ndcg([1,1,1,0,1,0,0,0]
> ```
> 
> and
> 
> ```python
> ndcg([1,0,1,0,0]
> ```
> 
> Both cases are 1-off the best solution, but the metric differs, isn't it?

The metric indeed differs. For the first case, `ndcg([1,1,1,0,1,0,0,0]) = 0.98`. For the second case, `ndcg([1,0,1,0,0]) = 0.92`

J535D165:
Thanks for resolving the issues. Can you add it to `Simulate` class?

J535D165:
> > Thanks for the implementation. I'm curious about this metric.
> > Can you also add this metric to the simulation output `Simulate`?
> > I wonder how this metric performs between datasets. The metric is perfect for comparing prior knowledge and models on a single dataset, but I want to know what happens if you compare datasets of various sizes and inclusion percentages.
> > Consider for example
> > ```python
> > ndcg([1,1,1,0,1,0,0,0]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```python
> > ndcg([1,0,1,0,0]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Both cases are 1-off the best solution, but the metric differs, isn't it?
> 
> The metric indeed differs. For the first case, `ndcg([1,1,1,0,1,0,0,0]) = 0.98`. For the second case, `ndcg([1,0,1,0,0]) = 0.92`

Thanks, it makes sense. Let's test it in our parameter optimization study :)

asreview/simulation/simulate.py
```diff
@@ -20,7 +20,7 @@
 import pandas as pd
 from tqdm import tqdm
 
-from asreview.metrics import loss
+from asreview.metrics import loss, ndcg
```
J535D165, line 23:
```suggestion
from asreview.metrics import loss
from asreview.metrics import ndcg
```

tests/test_metrics.py
```diff
@@ -2,7 +2,7 @@
 import pytest
 from numpy.testing import assert_almost_equal
 
-from asreview.metrics import loss
+from asreview.metrics import loss, ndcg
```
J535D165, line 5:
```suggestion
from asreview.metrics import loss
from asreview.metrics import ndcg
```

J535D165 (approved):

J535D165:
Thanks a lot

PR #2051, opened 2025-02-06
Rename navrail items (customize, settings)

J535D165:
As suggested by Berke. 

New: 

localhost_3000_reviews_f212ad6d81584f2695db28b477867f3b_setup

Old: 

localhost_3000_reviews_f212ad6d81584f2695db28b477867f3b_setup (1)

@AnqieDong @BerkeYazan @terrymyc Please help me? Yes or no?


AnqieDong:
> As suggested by Berke.
> 
> New:
> 
> localhost_3000_reviews_f212ad6d81584f2695db28b477867f3b_setup
> 
> Old:
> 
> localhost_3000_reviews_f212ad6d81584f2695db28b477867f3b_setup (1)
> 
> @AnqieDong @BerkeYazan @terrymyc Please help me? Yes or no?



> **Setup** and **Settings** in the new screenshot are confusing. **Settings** are often used for global settings, and the gear icon is used in this case. For project-specific adjustment, **Customize Project**/**Customization** is more appropriate. You don't _set up_ a project after a project has been _set up_.



> **Setup** and **Settings** in the new screenshot are confusing. **Settings** are often used for global settings, and the gear icon is used in this case. For project-specific adjustment, **Customize Project**/**Customization** is more appropriate. You don't _set up_ a project after a project has been _set up_.

I agree. “Customise” could work, or “Configuration”. Or “Tune” with this icon: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:tune:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=custom&icon.size=24&icon.color=%235f6368. If we change “Display” to “Settings”, it should have the settings icon, not the display settings icon.

BerkeYazan:
I think 'Settings' should be changed, I'm okay with all the alternatives like 'customize', 'tune' etc. It should indicate that we're changing the parameters of the model.

For the 'Display' part, I agree with Angie that it should have the settings gear icon if we were to go with that change.

PR #2053, opened 2025-02-07
Fix default authentication behaviour of lab entry-point

cskaandorp:
When the app is started with the asreview lab entry-point without authentication configuration, we assume no authentication. This is undermined by the create_app function in app.py.

### DIFF
--- a/asreview/webapp/_entry_points/lab.py
+++ b/asreview/webapp/_entry_points/lab.py
@@ -117,9 +117,7 @@ def lab_entry_point(argv):
         app.config.get("AUTHENTICATION", False) or args.authentication
     )
 
-    if "--test-mode" in argv:
-        # if this is a test, there is no need to start the server,
-        # return the app.
+    if app.testing:
         return app
 
     # NO MORE APP CONFIGURATION BELOW THIS LINE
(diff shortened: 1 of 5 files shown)


asreview/webapp/_entry_points/lab.py
```diff
@@ -330,4 +338,12 @@ def _lab_parser():
...
     )
 
+    parser.add_argument(
+        "--test-mode",
+        dest="test_mode",
+        default=False,
+        action="store_true",
+        help="Avoid starting the lab server",
+    )
+
```
J535D165, line 348: I like the idea of having a test mode for the entry point. What about using `FLASK_TESTING`? https://flask.palletsprojects.com/en/stable/config/#TESTING Then we don't have to implement an extra argument.

cskaandorp: My initial method, before the first push, was to use `app.testing`, but it turned out to be useless, as is FLASK_TESTING (no doubt the same thing). The testing environment is somehow not passed on to create_app if called from lab_entry_point. I did not do a deep dive into that, it just didn't work as far as I could see. I have opted for an extra argument on the lab side of things since we have defined a large number of parameters for lab, ignoring the TOML config of the Flask route. Did not wanted to mix those.

J535D165: The trick was to use `ASREVIEW_LAB_TESTING` instead of `FLASK_TESTING`. With a monkey patch, the test works now. Thanks for exploring this option. I'm pleased with this new test and the fixes.

PR #2066, opened 2025-02-13
Sign in a user after successful account creation or password reset

cskaandorp:
In the email-verified version of the app, the end-user should be signed in after the token is verified after account creation or password-reset.


cskaandorp:
I get a 404 after account creation, but the email is sent. Weird.

J535D165 (approved):

PR #2070, opened 2025-02-13
Modify to_file to return content without storing by default

DeependraChaddha:
PR updates the to_file method in BaseDataSet class to address the TODO. Previously, it only downloaded the dataset file to a specified path. now it returns the content directly if no path is provided, if path is provided it stores the data to that path. 

Changes: 
1. Modified to_file to return data without storing if path is not provided.
2. Handled possible URL Error which can occur while urlopen is being used

### DIFF
--- a/asreview/datasets.py
+++ b/asreview/datasets.py
@@ -174,9 +174,21 @@ class BaseDataSet:
 
         return self._filename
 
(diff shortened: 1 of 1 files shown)


J535D165:
We have many open issues, so feel free to start with them.

PR #2074, opened 2025-02-17
Add new theme colors and fix color engine implementation

J535D165:
image


asreview/webapp/src/constants/theme.js
```diff
@@ -1,19 +1,42 @@
...
         },
         tertiary: {
-          main: "#ffe08b",
+          main: "#ffe08b", // yellow, alternative: e3c46d
+          dark: "#e3c46d", // yellow, alternative: ffe08b
         },
         background: {
-          default: "#fffbf5",
-          paper: "#f2ede4",
+          default: "#fffdfa",
```
J535D165, line 38: @BerkeYazan I used a slightly lighter background color. The "yellow" background gets very yellow during the evening when modern devices use "night mode"-like features. I also tried to adjust the paper color a bit to it.

asreview/webapp/src/constants/theme.js
```diff
@@ -1,19 +1,42 @@
...
+// - The background color is a light beige color which is used
+// for the background of the app.
+
 export const theme = {
   colorSchemes: {
     light: {
       palette: {
         primary: {
-          main: "#A08F63",
+          main: "#7e7667", // greybrown
```
J535D165, line 28: @BerkeYazan, I'm not entirely sure about this color yet. I'm also unsure if we must use it for the "not relevant" elements. I might move to grey.800 for the not relevant elements.

asreview/webapp/src/constants/theme.js
BerkeYazan: That's a good concern. I agree with it.

asreview/webapp/src/constants/theme.js
BerkeYazan: Let's build everything together, then we can finalize them

BerkeYazan: I know how it looks now, but I must see how it feels

PR #2095, opened 2025-03-10
Python > v3.12 entrypoints are no longer mutable

jteijema:


asreview/models/cli.py
```diff
@@ -21,17 +21,13 @@ def _format_algorithm(values, name, description):
...
-
-        if hasattr(x, "label"):
-            result.append(
-                " " * 22 + f"{x.name}" + " " * (16 - len(x.name)) + f"{x.label}"
-            )
-        else:
-            result.append(" " * 22 + f"{x.name}")
+        padding = " " * 22 + x.name
+        if hasattr(x, "dist") and x.dist:
+            if x.dist.name != "asreview":
```
J535D165, line 26:
```suggestion
        if hasattr(x, "dist") and x.dist.name != "asreview":
```

Is this the same?

jteijema: I think when you do `x.dist.name` when `x.dist` is None, it will throw an error.

jteijema: If you combined these conditions into one, and x.dist is None or falsy (but still has a .dist attribute), this could raise an AttributeError or a NoneType error.

PR #2113, opened 2025-03-20
Fix email constraints when accounts are not created by signup

cskaandorp:
When accounts are created by using an external service (for instance OAuth)  we might not receive an email address of the user. The User model needs to be flexible enough to handle missing email addresses when accounts are not created by signup, but must verify email-address uniqueness when accounts are created by signup.

### DIFF
--- a/asreview/webapp/_authentication/models.py
+++ b/asreview/webapp/_authentication/models.py
@@ -44,7 +44,7 @@ class User(UserMixin, DB.Model):
     id = Column(Integer, primary_key=True)
     identifier = Column(String(100), nullable=False, unique=True)
     origin = Column(String(100), nullable=False)
-    email = Column(String(100), unique=True)
+    email = Column(String(100), nullable=True, unique=True)
     name = Column(String(100))
     affiliation = Column(String(100))
     hashed_password = Column(String(500))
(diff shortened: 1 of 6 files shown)


asreview/webapp/_authentication/models.py
```diff
@@ -126,18 +126,20 @@ def update_profile(
...
-                raise ValueError("Provided old password is incorrect.")
-            else:
-                # old password is verified. The following line will raise
-                # a ValueError if the new password is wrong
-                self.hashed_password = User.create_password_hash(new_password)
-
-        if self.email != email:
-            self.set_token()
-            self.confirmed = False
+        if self.origin == "asreview":
```
J535D165, line 129: Origin is `asreview` means login with email/password isn't it?

asreview/webapp/tests/test_database_and_models/test_user_model.py
```diff
@@ -62,46 +62,123 @@ def test_user_must_have_name(setup_teardown):
...
+    assert user2.origin == "asreview"
     # set to an existing identifier
     user2.email = user1.email
     with pytest.raises(IntegrityError):
         crud.create_user(DB, user2)
 
 
+# test that a single externally created account can miss an
+# email address
+def test_ext_single_account_creation_no_email(setup_teardown):
```
J535D165, line 121: Thanks a lot

J535D165 (approved):

J535D165:
Looks very good to me. Great improvement.

asreview/webapp/_authentication/models.py
cskaandorp: Correct

PR #2116, opened 2025-03-25
Fix gracefully terminating task manager when started from lab

cskaandorp:


cskaandorp:
This is pretty hard since I don't own a Windows machine. I have seen an error on my Mac as well, but only once in a while. A socket error. I have tried a couple of things to make things more robust. I have started the app 30 times and Ctrl-C'd all the time. The socket error didn't show up anymore, but I can't guarantee what Windows will do.

J535D165 (approved):

PR #2117, opened 2025-03-27
Fix reading/writing related bugs

PeterLombaers:
This pull request fixes some bugs for the data readers and writers. We changed the backend so that it will write away the labels provided in ASReview to the column `asreview_label`. We did not change the readers to read in this column again. So if you would export a dataset and then import it, you would see the complete dataset as unlabeled. 

Now we read the column `asreview_label` and will use it's contents as the prior knowledge. @J535D165 I gave the `asreview_label` column the highest priority. So if a dataset contains both `included` and `asreview_label` it will pick the value of `asreview_label`. You agree?

For RIS files, there is no tag corresponding to `asreview_label` so we need to put the data in the notes field. The label will be put as one of the strings `ASReview_relevant`, `ASReview_irrelevant` or `ASReview_not_seen`. This allows for easy searching in a reference manager. Other data columns starting with `asreview_`, such as `asreview_user_name` will be put in the notes field as `asreview_user_name: json.dumps(value)`.

### DIFF
--- a/asreview/data/base.py
+++ b/asreview/data/base.py
@@ -34,6 +34,7 @@ class BaseReader(ABC):
         "abstract": ["abstract", "notes_abstract", "abstract note"],
         "authors": ["authors", "first_authors", "author names"],
         "included": [
+            "asreview_label",
             "included",
             "label",
             "final_included",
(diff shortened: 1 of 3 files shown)


asreview/data/base.py
```diff
@@ -34,6 +34,7 @@ class BaseReader(ABC):
         "abstract": ["abstract", "notes_abstract", "abstract note"],
         "authors": ["authors", "first_authors", "author names"],
         "included": [
+            "asreview_label",
```
J535D165, line 37: 👍

J535D165 (approved):

PR #2118, opened 2025-03-27
Add test that reads dataset with record_id column

PeterLombaers:
Adds a test that checks that we can read datasets that contain a `record_id` column and that we ignore it's value.


J535D165 (approved):

J535D165:
Thanks!!

PR #2125, opened 2025-04-01
Prevent implicit bool conversion on classifier instances

timovdk:
Fixes a bug where Random Forest and AdaBoost were implicitly casted to bool. This casting only works for these models when the model is fit. At this point in the code the models are only initialized, not fitted, hence the error.


J535D165:
Thanks!

PR #2127, opened 2025-04-01
Add serialization tests

timovdk:
I would like feedback on these tests, specifically:
- Suggestions for classifier/feature extractor specific parameters (maybe a dict?)
- Suggestions for handling temp files ("alc1.json")

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -1,0 +1,152 @@
+from itertools import product
+
+import pytest
+
+import asreview as asr
+from asreview.extensions import extensions
+from asreview.models.queriers import Max
+
+# Get all classifiers and feature extractors from ASReview, filtering contrib models
+classifiers = [cls for cls in extensions("models.classifiers")]
+feature_extractors = [fe for fe in extensions("models.feature_extractors")]
+
+# Generate all combinations of classifier and feature extractor
+pairs = list(product(classifiers, feature_extractors))
+
+test_ids = [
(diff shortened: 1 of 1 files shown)


tests/test_serialization.py
```diff
@@ -0,0 +1,152 @@
...
+    )
+
+    alc2_meta = asr.ActiveLearningCycleData(
+        classifier=classifier.load()().name,
+        feature_extractor=classifier.load()().name,
+        balancer="balanced",
+        querier="max",
+    )
+
+    alc1.to_file("./alc1.json")
```
J535D165, line 107: Make use of the tmpdir fixture

asreview/models/feature_extractors.py
```diff
@@ -46,7 +46,58 @@ def transform(self, X):
...
+            contained subobjects that are estimators.
+        instances: bool, default=False
+            If True, will return the instances of the estimators in the pipeline.
+        """
+        params = super().get_params(deep=deep)
+        if not instances:
+            params.pop("text_merger", None)
+            params.pop("vectorizer", None)
+            params.pop("vectorizer__dtype", None)
+        return params
```
timovdk, line 97: I am quite unhappy about these, so I am open to suggestions.
The reason for these pops are:
1. `text_merger` and `vectorizer` are instances, so they are not JSON serializable in the .to_file() function
2. `vectorizer__dtype` is by default a `numpy.float64` which is also not JSON serializable in the .to_file() function

PR #2155, opened 2025-04-22
Fix integration tests

dometto:
While working on #2129 I tried to implement a test for the new functionality allowing the customization of the post-logout redirect URL. I noticed the integration tests where currently not working. I took the liberty of fixing them, which required the following steps, only one of which is major:

* updating some selector paths in the integration tests' `utils.py`, due to changes in the frontend
* updating some expected HTML, due to changes in the frontend
* **fixing an issue with updating the user profile.** At present, when a user updates their email address, this will set the user's `confirmed` status to `False`, *even if email verification is turned off globally*.

I suppose the latter fix (if this is indeed a bug, as I have interpreted it) might also constitute a separate PR?

I have also included an experimental GitHub Actions workflow for running integration tests. I suppose that to save time/resources, this could also be made part of the `ci-webapp` workflow, so things like asset precompilation don't need to be done twice.

No worries if these changes are not helpful, e.g. if you were planning to change the test workflows anyway, but thought I'd contribute!

### DIFF
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -1,0 +1,54 @@
+name: Test ASReview webapp (integration)
+on: [push, pull_request]
+permissions:
+  contents: read
+jobs:
+  selenium-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        browser: # must be valid values that can be passed to pytest's --driver option
+          - 'chrome'
+    env:
+      ASREVIEW_LAB_SQLALCHEMY_DATABASE_URI: sqlite:////tmp/db.sqlite
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+          fetch-tags: true
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "20"
+          cache: "npm"
+          cache-dependency-path: "asreview/webapp/package-lock.json"
+      - uses: nanasess/setup-chromedriver@v2
+        if: matrix.browser == 'chrome'
+      - name: Install requirements
+        run: pip3 install .[test] setuptools gunicorn
+      - name: Compile assets
+        run: python setup.py compile_assets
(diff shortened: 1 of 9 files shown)


.github/workflows/ci-integration.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: pip3 install .[test] setuptools gunicorn
+      - name: Compile assets
+        run: python setup.py compile_assets
+      - name: Start ASReview
+        run: |
+          pip3 install .
+      - name: Run tests
+        run: |
+          export DISPLAY=:99
+          asreview lab --no-browser --config-path asreview/webapp/tests/config/auth_integration.toml --skip-update-check &
```
dometto, line 35: The tests fail because this line seems to immediately exit, so there is no server running to test against.

Weirdly, I get the same result locally when I run:

`asreview lab --config-path asreview/webapp/tests/config/auth_integration.toml`

When I leave out `--config-path`, it works. Any idea what might be going on here?

dometto: This was due to setting `Testing = true` in the app config. Not sure if this is indeed desired behavior, but I'll close this for now.

.github/workflows/ci-integration.yml
```diff
@@ -0,0 +1,54 @@
+name: Test ASReview webapp (integration)
+on: [push, pull_request]
```
dometto, line 2: Since these tests are not *super* lightweight, I could imagine doing some things to optimize the workflow run:

* run it after the normal tests for `ci-webapp`. That way, if the normal tests fail, integration tests are not run. Plus compiling assets will already have been done.
* if we keep this in a separate workflow, run `ruff` in front of the tests, so they don't run on syntax errors etc.?
* *or* run the tests for PRs only if certain files have been altered, e.g. the JS. Not sure about this, since API changes can also break integration tests.

dometto:
Pinging @cskaandorp :)

cskaandorp:
@dometto: my sincere apologies for my very very late entrance. Let me start with thanking you for all your hard work! Great stuff. I am going through your code and will reply very soon.

J535D165 (approved):

Issue #2147, opened 2025-04-17
need to correct calculations of performance metrics in simulation mode


fqixiang:
**Describe the bug**
When we run a simulation study, the asreview code prints the loss and ndcg values automatically, but they are not saved. This is not a problem, since we can just calculate the loss and ndcg values based on labels saved in `self._results['labels']`. However, I noticed that the ndcg values printed during the simulation runs were always a lot higher than the ones I reconstructed later. I figured out this is because the labels used by the simulation code (`asreview.simulation.simulate.py`) included the prior knowledge labels (see the screenshot), while in my re-calculation I excluded them. I think it makes more sense to exclude the prior labels, so that the performance isn't overestimated.

**To Reproduce**
Steps to reproduce the problem:

**Screenshots**

![Image](https://github.com/user-attachments/assets/336c9805-a6a0-45fa-9ff7-eb1ced0c0bc3)

**Version information**

- OS: [e.g. Windows]
- Browser (if applicable) [e.g. Chrome]
- ASReview version [e.g. 0.6.2]

**Additional context**

J535D165:
Thanks for reporting. Can you open a PR?

PR #2157, opened 2025-04-22
Remove prior labels from the calculation of loss and ndcg in simulation mode

fqixiang:
fixed https://github.com/asreview/asreview/issues/2147

### DIFF
--- a/asreview/metrics.py
+++ b/asreview/metrics.py
@@ -36,8 +36,7 @@ def loss(labels: list[int]):
     Ny = sum(labels)
     Nx = len(labels)
     if Ny == 0 or Nx == Ny:
-        warnings.warn("Labels do not contain two distinct classes. Returning NaN.")
-        return np.nan
+raise ValueError("Labels must contain two distinct classes.")
     return float(
         (Ny * (Nx - (Ny - 1) / 2) - np.cumsum(labels).sum()) / (Ny * (Nx - Ny))
     )
@@ -61,8 +60,7 @@ def ndcg(labels: list[int]):
     Ny = sum(labels)
     Nx = len(labels)
     if Ny == 0 or Nx == Ny:
-        warnings.warn("Labels do not contain two distinct classes. Returning NaN.")
-        return np.nan
+raise ValueError("Labels must contain two distinct classes.")
 
     dcg = np.sum(labels / np.log2(np.arange(2, Nx + 2)))
     idcg = np.sum(1 / np.log2(np.arange(2, Ny + 2)))
--- a/asreview/metrics.py
+++ b/asreview/metrics.py
@@ -17,7 +17,6 @@
 __all__ = ["loss", "ndcg"]
 
 import numpy as np
-import warnings
 
 
 def loss(labels: list[int]):
--- a/asreview/metrics.py
+++ b/asreview/metrics.py
@@ -17,6 +17,7 @@
 __all__ = ["loss", "ndcg"]
 
 import numpy as np
+import warnings
 
 
 def loss(labels: list[int]):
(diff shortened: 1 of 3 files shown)


fqixiang:
So the test failed... not so sure if this is an issue with the loss function or with the test data (I can't seem to find the `demo_data.csv` file that is mentioned in the report.

J535D165:
A tiny bit odd, but a brave unit test :) After subtracting the prior knowledge, there might no longer be two distinct labels. You have to catch this error.

asreview/simulation/simulate.py
```diff
@@ -212,9 +212,9 @@ def review(self):
             pbar_rel.close()
             pbar_total.close()
 
-            padded_results = list(self._results["label"]) + [0] * (
-                len(self.labels) - len(self._results["label"])
-            )
+            padded_results = list(
+                self._results.dropna(axis=0, subset="training_set")["label"]
+            ) + [0] * (len(self.labels) - len(self._results["label"]))
```
J535D165, line 217: You can resolve this with an if statement or with try-except (which is more Pythonic). Or a combination of both if you want to provide more informative feedback. 

```python
if len(set(padded_results)) < 2: 
    print("Can't compute loss and gain for labels with only relevant or irrelevant records")
```

asreview/metrics.py
```diff
@@ -35,7 +36,8 @@ def loss(labels: list[int]):
     Ny = sum(labels)
     Nx = len(labels)
     if Ny == 0 or Nx == Ny:
-        raise ValueError("Labels must contain two distinct classes.")
+        warnings.warn("Labels do not contain two distinct classes. Returning NaN.")
+        return np.nan
```
J535D165, line 40: This should raise an error instead of a warning.

tests/test_metrics.py
```diff
@@ -62,9 +62,12 @@ def test_loss_value_function(labels, expected_value):
...
 )
-def test_loss_value_error_cases(labels):
-    with pytest.raises(ValueError):
-        loss(labels)
+def test_loss_warning_message(labels):
+    with pytest.warns(
+        UserWarning, match="Labels do not contain two distinct classes. Returning NaN."
+    ):
+        result = loss(labels)
+        assert np.isnan(result)
```
J535D165, line 70:
```suggestion
def test_loss_value_error_cases(labels):
    with pytest.raises(ValueError):
        loss(labels)
```

tests/test_metrics.py
```diff
@@ -137,9 +140,12 @@ def test_ndcg_value_function(labels, expected_value):
...
 )
-def test_lndcg_value_error_cases(labels):
-    with pytest.raises(ValueError):
-        ndcg(labels)
+def test_lndcg_warning_message(labels):
+    with pytest.warns(
+        UserWarning, match="Labels do not contain two distinct classes. Returning NaN."
+    ):
+        result = ndcg(labels)
+        assert np.isnan(result)
```
J535D165, line 148:
```suggestion
def test_lndcg_value_error_cases(labels):
    with pytest.raises(ValueError):
        ndcg(labels)
```

asreview/metrics.py
```diff
@@ -59,7 +61,8 @@ def ndcg(labels: list[int]):
...
```
J535D165, line 65:
```suggestion
raise ValueError("Labels must contain two distinct classes.")
```

asreview/metrics.py
J535D165, line 40:
```suggestion
raise ValueError("Labels must contain two distinct classes.")
```

asreview/metrics.py
```diff
@@ -35,7 +36,7 @@ def loss(labels: list[int]):
     Ny = sum(labels)
     Nx = len(labels)
     if Ny == 0 or Nx == Ny:
-        raise ValueError("Labels must contain two distinct classes.")
+raise ValueError("Labels must contain two distinct classes.")
```
J535D165, line 39:
```suggestion
        raise ValueError("Labels must contain two distinct classes.")
```

asreview/metrics.py
```diff
@@ -59,7 +60,7 @@ def ndcg(labels: list[int]):
...
```
J535D165, line 63:
```suggestion
        raise ValueError("Labels must contain two distinct classes.")
```

asreview/metrics.py
```diff
@@ -17,6 +17,7 @@
 __all__ = ["loss", "ndcg"]
 
 import numpy as np
+import warnings
```
J535D165, line 20:
```suggestion
```

timovdk (approved):
Nice!

PR #2184, opened 2025-05-06
Fix database migration entry point

cskaandorp:
The previous version of the migration script failed to commit changes. This fix will commit and add default roles to existing user records. Tested on localhost, but also on future.asreview.app


J535D165 (approved):

J535D165:
Lovely. Thanks a lot

PR #2188, opened 2025-05-07
Fix learners label loading time with ast parsing

J535D165:
Not the solution I'm most proud of, but it looks stable and very fast. Extensive research shows that there is no good alternative atm of writing.


timovdk (approved):
Nice to look at the source files, indeed seems like a (not so pretty, but) stable solution for now! Also cool that Dory does not have to implement anything, as long as the name and label variables are there!

PR #2192, opened 2025-05-07
auth_tool: allow adding users with non-standard origin

dometto:
Currently `asreview auth-tool` assumes users added using the `--json` option have `asreview` as their `origin` attribute. That means users must have a valid email and a password.

However, for remote authentication mechanisms, we might want to create users that do not meet this criteria. In particular, we may want to pre-add valid users to the database when using Remote User auth, so that valid users are already visible in the app *before* they are created when they first login.

This PR allows setting the `origin` attribute in the JSON data passed along with the `--json` option.

### DIFF
--- a/asreview/webapp/_entry_points/auth_tool.py
+++ b/asreview/webapp/_entry_points/auth_tool.py
@@ -148,6 +148,7 @@ def insert_user(session, entry):
     into the database."""
     # create a user object
     user = User(
+        entry.get("origin", "asreview").lower(),
         entry["email"].lower(),
         email=entry["email"].lower(),
         name=entry["name"],


cskaandorp:
This makes sense to me, thanks. We will move towards an "admin" feature in which an admin can take care of adding user records, but still, I think the auth tool might still play a role (admin-account injection for instance). I will propose a bit of code for this PR tomorrow in order to enter the origin when a user account is added by manual input (not per json file).

cskaandorp (changes requested):
This PR ensures we can inject user accounts with the auth-tool when Remote User Handling is configured. Thank you. Would you please consider my code to update the `enter_users` function as well to facilitate this for manual input as well:

```
    def enter_users(self):
        while True:
            new_user = input("Enter a new user [Y/n]? ")
            if new_user == "Y":
                email = self._ensure_valid_value_for(
                    "Email address (required)",
                    User.valid_email,
                    "Entered email address is not recognized as a valid email address.",  # noqa
                )
                name = self._ensure_valid_value_for(
                    "Full name (required)",
                    lambda x: bool(x) and len(x) > 2,
                    "Full name must contain more than 2 characters.",
                )
                origin = input("Origin (leave empty if you're not using Remote User Handling): ")
                origin = origin.strip()
                affiliation = input("Affiliation: ")
                password = self._ensure_valid_value_for(
                    "Password (required)",
                    User.valid_password,
                    "Use 8 or more characters with a mix of letters, numbers & symbols.",  # noqa
                )

                entry_data = {
                    "email": email,
                    "name": name,
                    "affiliation": affiliation,
                    "password": password
                }
                # add origin if string is provided
                if origin != "":
                    entry_data["origin"] = origin

                insert_user(self.session, entry_data)
            else:
                break

        return True
```

dometto:
@cskaandorp good idea! I've incorporated your code, which also required updating some more tests. There is a readthedocs workflow that fails now, with this message:

`INFO: pip is looking at multiple versions of asreview to determine which version is compatible with other requirements. This could take a while.
ERROR: Package 'asreview' requires a different Python: 3.9.22 not in '>=3.10'`

This does not seem specific to this PR, I think?

PR #2218, opened 2025-05-19
Quickfix file type MIME type selection

J535D165:
This avoids uploading files in the wrong place (file vs import).

The downside of this method is that custom readers are no longer working cc @PeterLombaers. We need a permanent solution to that (update the readers).

### DIFF
--- a/asreview/webapp/src/ProjectComponents/ImportProject.js
+++ b/asreview/webapp/src/ProjectComponents/ImportProject.js
@@ -70,7 +70,7 @@ const ImportProjectCard = ({ mutate, isLoading, isError, error }) => {
     multiple: false,
     noClick: true,
     accept: {
-      "application/zip": [".asreview"],
+      "application/asreview": [".asreview"],
     },
   });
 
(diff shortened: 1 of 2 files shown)


PeterLombaers:
I don't mind fixing it like this for now, just so that we prevent any more issues coming in where this is happening.

If we want to support custom readers, we would need to declare the MIME type on the reader in addition to the file extension and then update the API so that the frontend can also get MIME type? 

If you want a quick and dirty solution, you could also only update the error message when there is an error adding the dataset. In general, this would be useful: for example if there is an error adding the dataset and the extension is `.txt`, the message say `are you sure the file is a valid RIS file?` or something similar.

Another thought I have is: why do we even distinguish between file vs import?

J535D165:
Thanks for this review. Maybe I should update the readers as you describe it ("if not now, when?"). I just opened a PR to improve the error presentation https://github.com/asreview/asreview/pull/2230, so I will work on this one next. 

> Another thought I have is: why do we even distinguish between file vs import?

Good one. I considered merging them, but uploading a project file under File might be ambiguous. Will it create a new project from that data file inside the project, or import the project? I thought that would be confusing. But UI improvements are clearly needed, as users are reporting issues on this.

Issue #2219, opened 2025-05-19
Export error


sannepeereboom:
**Describe the bug**
I am trying to export the records with decisions but seem to be getting a TypeError. The same thing happens when I try to export unseen records only (i.e. records without tags)

**To Reproduce**
Steps to reproduce the problem:
Within a project,
1. Go to "Collection"
2. Export
3. Any of the export options lead to the error (pasted at end)

Data was imported as an Excel file and the project was migrated to version 2.0 (from 1.6.5).

EDIT: This issue only seems to happen when tags are defined within the project.

**Screenshots**

**Version information**

- OS: Windows 10
- Browser (if applicable): MS Edge
- ASReview version: 2.0

**Additional context**
`ERROR:asreview.webapp.app:Exception on /api/projects/[project id]/export_dataset [GET]
Traceback (most recent call last):
  File "C:\Users\user\AppData\Local\Programs\Python\Python311\Lib\site-packages\flask\app.py", line 1455, in wsgi_app
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\Programs\Python\Python311\Lib\site-packages\flask\app.py", line 869, in full_dispatch_request
    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\Programs\Python\Python311\Lib\site-packages\flask_cors\extension.py", line 176, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
                                                ^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\Programs\Python\Python311\Lib\site-packages\flask\app.py", line 867, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\Programs\Python\Python311\Lib\site-packages\flask\app.py", line 852, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\Programs\Python\Python311\Lib\site-packages\asreview\webapp\_authentication\decorators.py", line 50, in decorated_view
[...]

PR #2222, opened 2025-05-19
Fix export of migrated projects with tags

J535D165:


asreview/webapp/_api/projects.py
```diff
@@ -1070,8 +1070,15 @@ def _flatten_tags(results, tags_config):
         del results["tags"]
         return results
 
+    print(results["tags"])
+
```
J535D165, line 1074:
```suggestion
```

Rensvandeschoot:
solves https://github.com/asreview/asreview/issues/2219

PR #2229, opened 2025-05-20
Resolve IndexError for older versions of Pandas

timovdk:
For older versions of Pandas, we would get an index error when trying to index an `np.ndarray` with a `pandas.core.arrays.integer.IntegerArray`. This change ensures that we always index with an `np.ndarray`.


J535D165 (approved):

J535D165:
Grazie mille!

Issue #2252, opened 2025-06-03
Unable to Create a New Project with version 2.0.2


Phillip-100:
Hi,
I wanted to create an ASReview Project with the new version I just upgraded to. I'm using a partially labeled dataset with around 140k records that has been cleaned and verified. However, I get the same error message even if I drop the file or select it from a folder, and even if I change the file type from CSV to XLSX.
Error message: Failed to create project for this dataset. JSONDecodeError.__init__() missing 2 required positional arguments: 'doc' and 'pos'

**Screenshots**

![Image](https://github.com/user-attachments/assets/6c7eca49-3743-4f5b-8434-aff070624269)

**Version information**

- OS: Windows 11
- Browser (if applicable) [e.g. Chrome]
- ASReview version 2.0.2

**Additional information**
I had used the same dataset with less labels before (using version 1.6.6), and it didn't show any issues. Would it be possible to revert back to a previous version of ASReview?

Phillip-100:
Hi, I just sent the dataset via email. 
Thanks for your support!

J535D165:
The traceback for this error (cc @PeterLombaers )

```
INFO:werkzeug:127.0.0.1 - - [04/Jun/2025 09:53:29] "GET /api/projects?subset=oracle HTTP/1.1" 200 -
ASReview_1st_Model_Edited.xlsx
ERROR:root:JSONDecodeError.__init__() missing 2 required positional arguments: 'doc' and 'pos'
Traceback (most recent call last):
  File "/Users/Bruin056/Documents/GitHub/asreview/asreview/webapp/_api/projects.py", line 250, in api_create_project
    project.add_dataset(
  File "/Users/Bruin056/Documents/GitHub/asreview/asreview/project/api.py", line 262, in add_dataset
    records = _from_file(save_fp, dataset_id=dataset_id)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Bruin056/Documents/GitHub/asreview/asreview/data/loader.py", line 76, in _from_file
    return reader.read_records(fp, dataset_id=dataset_id, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Bruin056/Documents/GitHub/asreview/asreview/data/base.py", line 68, in read_records
    df = cls.clean_data(df)
[...]

PR #2276, opened 2025-06-19
Handle character limits for cells in CSV and Excel

PeterLombaers:
This PR should fix the problem encountered in #2252. The problem is that when we read a RIS file there can be arbitrarily many authors in a list of authors of a record. When writing to CSV or Excel there is a maximum limit to the number of characters in a single cell. What happens if the number of characters exceeds this limit is not clearly defined and can lead to errors when reading the CSV/Excel file again.

What I did is:
- When reading new input data, if the authors or keywords field contains a string, I convert it to a list by splitting on `;`. If that doesn't give the correct list of authors, the user should format their file differently.
- When exporting a dataset, if the input file was RIS and the output is CSV/Excel, I convert the list fields to string by joining on `;`, then truncating the string if it exceeds the character limit. This should mean that it's always possible to read the file again, though possible some data is lost in the process if you have a huge list of authors or keywords.

I did not yet write a test yet, you can see I removed a test in the last commit because it was not the correct test. You should use the webapp and an integration test, because it only happens the scenario with RIS input and CSV/Excel output.

The old `convert_to_list` is still used in one spot in `RISReader`. I did not yet have time to check what that code is exactly doing and how it should be changed for the new way of parsing lists.


PeterLombaers:
I have added a test RIS file with a lot of authors. Feel free to use it in an integration test, or to remove the file completely.

PeterLombaers:
I fixed the last bugs and added an API test. Should be ready now if the checks all pass.

J535D165:
Very welcome!

PR #2288, opened 2025-06-26
Fix overlap when y-axis data has 4 digits

BerkeYazan:
- Increased the space between the y-axis label and the y-axis data for `Recall` and `Wave` charts


J535D165:
Is this not relevant for the density plot?

J535D165:
You are right. I was waiting for the weekend I think.

J535D165:
Grazie mille!

PR #2290, opened 2025-06-26
Add option to delete your account from the server

J535D165:
Add an option to delete your profile from the server. 

@cskaandorp this might also be visible during profile setup with OAuth, isn't it? What is an easy way to prevent this?

@dometto Do we need to disable this on your setup, or is it fine when they delete their account? I think that's fine, isn't it?

localhost_3000_signup(iPad Pro)


dometto:
@J535D165 I indeed expect no issues with remote user auth from this PR

cskaandorp (commented):
No issues

PR #2293, opened 2025-06-28
Pr 2290 suggestions

cskaandorp:
Not sure about deleting the projects from the ASReview folder, if that is not what we want please remove. I have added some tests as well. Looks good to me.


cskaandorp:
Committed as a standalone PR but was meant for PR 2290

PR #2305, opened 2025-07-08
Use email as identifier when creating users with remote user auth

dometto:
I ran into a bug when using `asreview auth-tool` to periodically add users to ASReview on ResearchCloud:

1. When users are created with auth-tool, `email` (not `username`) is used as the user's identifier
2. But the remote user auth mechanism now checks for the pre-existence of a user using `username` as an identifier

So after user `foo` is created with auth-tool, and when logging in `foo` for the first time (using remote auth):

1. ASReview checks to see if there is already a user with identifier `foo` (no, because auth-tool created a user with identifier `foo@localhost`)
2. ASReview attempts to create a new user with username `foo` and identifier `foo`. This fails, because a user with those attributes already exists.

To make things more robust, this PR changes the user creation mechanism for remote user auth to use `email` as an identifier. I think this is preferred anyway, looking at other authentication implementations?

### DIFF
--- a/asreview/webapp/_authentication/decorators.py
+++ b/asreview/webapp/_authentication/decorators.py
@@ -162,8 +162,10 @@ def login_remote_user(f):
                     user_info = remote_user_handler.handle_request(request.environ)
                 except HTTPException as e:
                     return jsonify({"message": e.description}), 401
+                except ValueError:
+                    user_info = None
 
-                if user_info["identifier"]:
+                if user_info:
                     user = User.query.filter(
                         User.identifier == user_info["identifier"]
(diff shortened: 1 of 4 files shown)


dometto:
@J535D165 @cskaandorp  who could best review this PR?

J535D165:
> To make things more robust, this PR changes the user creation mechanism for remote user auth to use email as an identifier. I think this is preferred anyway, looking at other authentication implementations?

I agree with this (although the email requirement is a struggle for us, as not all OAuth providers require it).

J535D165:
Thanks

PR #2405, opened 2025-11-24
Revamp team membership workflow with token-based project invitations

cskaandorp:
This PR introduces a complete overhaul of the team membership workflow by replacing the old pending-invitations system with a secure, token-based invitation mechanism.

Key Changes:

* Token-based project invites: Project owners and administrators can now generate a unique invitation token (with an optional QR code) that can be shared with prospective team members.
* Persistent tokens: Invitation tokens remain valid until they are regenerated or explicitly revoked by the project owner.
* Streamlined onboarding flow: Anyone with the token link is prompted to join the project. If the user is not logged in, they will be redirected to sign in first and then automatically returned to the join-project screen.
* Removal of obsolete pending-invitations table: The former pending_invitations table is no longer needed and has been removed.
This cleanup is handled via the database migration entry point.

### DIFF
--- a/asreview/webapp/tests/test_database_and_models/test_database_creation.py
+++ b/asreview/webapp/tests/test_database_and_models/test_database_creation.py
@@ -24,9 +24,7 @@ def test_database_exists_after_starting_auth_app(auth_app):
 
 
 # checks if all tables were created
-@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


J535D165:
Maybe I prefer a M level QR code if possible.

J535D165 (approved):

PR #2420, opened 2025-12-01
Feat: Add upload progress indicator for uploading datasets and project imports

cskaandorp:
Users uploading large datasets experienced long periods with no visual feedback, causing uncertainty about whether uploads were progressing or stalled. This change adds real-time upload progress bars during file transfers for both dataset uploads and project imports. The progress is tracked client-side using axios callbacks and displays "Data has been uploaded" when complete.


J535D165:
You will make many people very happy with this!

PR #2511, opened 2026-04-13
Revert batch data ingestion

J535D165:


J535D165:
This PR shows that the issue @PeterLombaers fixed in https://github.com/asreview/asreview/pull/2504 is Windows related.

PR #2523, opened 2026-04-22
Import an original_id for consistent mapping with external datasets.

michi-zuri:
I would like to preserve my imported Zotero keys, so that it is easier to write my screening decisions back to the corresponding articles in Zotero. This pull requests adds a column with this id. 

Co-authored by: Claude Opus 4.6

### DIFF
--- a/asreview/data/base.py
+++ b/asreview/data/base.py
@@ -47,6 +47,7 @@ class BaseReader(ABC):
             "include",
         ],
         "title": ["title", "primary_title"],
+        "original_id": ["original_id", "id"],
     }
 
     # Dictionary {column name : function to apply to the column} of function that clean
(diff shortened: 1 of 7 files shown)


J535D165:
Did you import the records from a CSV file? Are the original keys not already preserved in the output. By design, ASReview should preserve the original data.

michi-zuri:
Hmm, I don't think the id was preserved previously, but i'll double check and report back my findings. At least the SQLite database doesn't seem to be able to save non integer ids.

michi-zuri:
Ah I see what you mean now, there is a dataset_row column, which maps the record_id to the imported dataset. But row number is not the Zotero_id, it's the row number... When I export my results I want to backport them to my Zotero library and not just the exported data sample. 

I imported both from RIS and CSV. By default Zotero doesn't export the zotero_id to CSV, but I modded to RIS translator to include an ID  - tag in the export. 

As I understand record_id has to be integer because it is also used as accessor in the machine learning data structures. 

What would an alternative solution be that solves my problem of needing to backport the screening decisions back to my Zotero library? DOI mapping is not perfect, because 5% of articles lack this and with duplicate articles (same title) I would need to fallback to abstract mapping which feels wrong to me.

Issue #2294, opened 2025-06-29
Adding and editing a note resets selected tags


soelderer:
**Describe the bug**
I have a list of tags in a screening project. When screening an abstract, I selected one or more of the tags.
Then I add a note. After saving the note, the tags are all reset, i.e. none of the tags is selected anymore. So I have to select the tags again.
This also happens if I first add a note, then select tags and then edit the note. Editing resets the tags.

I would expect that adding and editing notes do not alter the selected tags.

**To Reproduce**
(1) Create a screening project
(2) Define some tags
(3) Start screening
(4) For any record: (i) select one or more tags, (ii) create and save a note

**Version information**

- OS: Arch Linux 6.15.2-arch1-1
- Browser: Brave 1.79.123 (Official Build) (64-bit) Chromium: 137.0.7151.104
- ASReview version: 2.1

PR #2524, opened 2026-04-22
added white-space:pre-wrap to the user-entered note field

michi-zuri:
so that notes are as readable after saving as while they are being edited.


J535D165:
Thanks for your PR! Does this close https://github.com/asreview/asreview/issues/2294?

michi-zuri:
Ah no, it's just am improvement for the note view, but unsaved Tag selections are not rescued by this pr, let me make a separate pr for that this week, would be fun to fix!

PR #2526, opened 2026-04-24
Make TextMerger.transform return a list instead of pd.Series

cskaandorp:
Dory's sentence-transformer feature extractors broke after the refactor because `TextMerger.transform returns a pd.Series`, and sentence-transformers versions reject it with `ValueError: Unsupported input type: Series. Expected one of: str, dict, PIL.Image.Image, np.ndarray, torch.Tensor`.

This changes `TextMerger.transform` to return a plain `list[str]`. A list is the most universally accepted "iterable of strings" across the libraries we and downstream extensions compose with — sklearn vectorizers, sentence-transformers, HuggingFace tokenizers, and gensim all accept it without conversion.

### DIFF
--- a/asreview/models/feature_extractors.py
+++ b/asreview/models/feature_extractors.py
@@ -43,7 +43,7 @@ class TextMerger(TransformerMixin, BaseEstimator):
         return self
 
     def transform(self, X):
-        return X[self.columns].fillna("").apply(lambda x: self.sep.join(x), axis=1)
+        return X[self.columns].fillna("").apply(lambda x: self.sep.join(x), axis=1).tolist()
 
 
 class Tfidf(Pipeline):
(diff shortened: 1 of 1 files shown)


J535D165:
Is this caused by upstream changes in the sentence transformers library?

cskaandorp:
As far as I understand: yes. I want to be careful here since I'm not an expert, but from what I can tell, strict input type validation was introduced in version 5.4.0.

J535D165:
@timovdk can you review this? 

I'm in favor of fixing this downstream as this might be a breaking change in the ASReview API

timovdk:
Thanks for flagging this, I fixed this in [Dory](https://github.com/asreview/asreview-dory/pull/47). New version of Dory is released, closing this PR

PR #2531, opened 2026-04-30
Fix tag selections being lost when saving a note during screening

### LINKED ISSUE #2334: Bug when adding note
When you add & save a note while screening all checked tags get unchecken

**To Reproduce**
Steps to reproduce the problem:
create tags, check tags while screening, add and save a note,
**Screenshots**

**Version information**

- OS: windows
- Browser edge
- ASReview version 2.1

**Additional context**

michi-zuri:
The RecordCardLabeler component's React key included the record's note value. When a note was saved, the note changed from null to the note text, which changed the key and caused React to unmount/remount the component. This reinitialized the local tag checkbox state from the server data (null for unlabeled records), losing any checked tags.

Fixed by removing note from the component key, and adding a useEffect in NoteDialog to reset internal state when the dialog opens (previously handled implicitly by the key-triggered remount).

Fixes #2334 and #2294

### DIFF
--- a/asreview/webapp/src/ProjectComponents/ReviewComponents/RecordCard.js
+++ b/asreview/webapp/src/ProjectComponents/ReviewComponents/RecordCard.js
@@ -206,12 +206,7 @@ const RecordCard = ({
           <Grid size={landscape ? 2 : 5}>
             <RecordCardLabeler
               key={
-                "record-card-labeler-" +
-                project_id +
-                "-" +
-                record?.record_id +
-                "-" +
-                record?.state?.note
+                "record-card-labeler-" + project_id + "-" + record?.record_id
               }
               project_id={project_id}
               record_id={record.record_id}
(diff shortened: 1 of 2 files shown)


asreview/webapp/src/ProjectComponents/ReviewComponents/RecordCard.js
```diff
@@ -206,12 +206,7 @@ const RecordCard = ({
           <Grid size={landscape ? 2 : 5}>
             <RecordCardLabeler
               key={
-                "record-card-labeler-" +
-                project_id +
-                "-" +
-                record?.record_id +
-                "-" +
-                record?.state?.note
```
michi-zuri, line 214: I suspect this line should not be deleted. The current state is not displayed correctly in the record card of the labelling history. Will check and fix.

michi-zuri (commented):
needs fixing

asreview/webapp/src/ProjectComponents/ReviewComponents/RecordCard.js
michi-zuri: When loading the articles with a filter, the tag states are not shown. Needs further investigation.

michi-zuri: okay never mind, the bug was caused by a wrong implementation in #2533, which is now fixed.

Issue #2320, opened 2025-07-17
Feature Requests: Summarizing Tagging Functionality Improvements


Rensvandeschoot:
Some suggestions to improve the tagging functionality in ASReview:

---

#### Edit Tags

- Allow users to **re-label tags** within the collection screen. https://github.com/asreview/asreview/issues/2260

#### Remove Tags

- Allow users to **delete tags** after creation, either during project setup or later in the customization screen.

#### Re-order Tags

- Allow users to **delete tags** after creation, either during project setup or later in the customization screen. https://github.com/asreview/asreview/issues/2133

#### Filter by Tags

- Enable **filtering the collection by tag** to focus on specific subsets of records, directly in the collection screen.

#### Toggle Tags off/on
- 
- Toggle of ' show tags', for example, when viewing the collection now it takes up a lot of space and sometimes you do not need to see them. https://github.com/asreview/asreview/issues/2261

#### Allow Open Text Tags (Free-Text Entry)

- Currently, ASReview supports **multiple-choice tags** only.
- Add support for **open text tags**, **numeric field**, or **forced 1-answer options**

Issue #2316, opened 2025-07-17
Feature Request: Enhanced Collection Management and Screening History


Rensvandeschoot:
## Problem Description

Currently, the **collection screen** in ASReview displays all labeled records—both relevant and irrelevant. However, in practice, the **collection of records for output products** (e.g., systematic review, meta-analysis, policy report) usually only includes the **relevant studies**.

At the same time, with the introduction of **tags**, the collection screen has become more than a list of labeled records—it is increasingly used as a **workspace for managing the final output**.

---

#### 1. Dedicated Collection Screen

- Create a **dedicated collection view** that only contains **records labeled as relevant**.
- This collection represents the **output selection** for further use in reviews or reports.
- Allow users to:
  - **Filter on tags**
  - **Filter by screener**
  - **Search for specific records** within the collection.

- Q: Maybe **Rename the "Include" button** to **"Add to collection"** to better reflect the purpose.

#### 2. History Screen

- Provide a **History view** with an **ordered list of labeling actions**, (which is the current collection screen) including:
  - The sequence in which records were labeled.
  - Who labeled each record.
  - The model used (if applicable).
- Allow users to **search for specific records** in the history.
- Enable moving a record to the **Collection** (if decisions need correction or adjustment).

---

### Benefits

- Clarifies the difference between **labeling for training** and **building a final collection**.
- Supports **audit trails** and **decision transparency**.
- Provides better management of the **final output set**, especially for systematic reviews or reports.
- Improves usability for projects involving **multiple screeners and tags**.

michi-zuri:
Please don't remove the view of irrelevant articles, it is very helpful! 

The current 3-collection view is great in my opinion. 
<img width="314" height="66" alt="Image" src="https://github.com/user-attachments/assets/c72f000a-1dee-4efe-96c0-dc27d6b60f23" />

I don't really get the problem that this issue is raising apart from wanting to implement better searchability and discoverability on the collection screen. This doesn't necessitate removing the view of irrelevant articles though. Maybe @Rensvandeschoot you could elaborate on "Clarifies the difference between labeling for training and building a final collection."? Does re-labelling an article not trigger training queue currently? I was guessing it would retrigger training, but maybe my intuition is wrong. If training is not triggered, what is the reasoning behind this? Is it a technical limitation, that the trainer can only deal with new decisions and doesn't know how to deal with removing previous decisions? I am guessing the latter:

[...]

PR #2533, opened 2026-05-06
Redesign collection filters with 3-state toggles and tag filtering

michi-zuri:
Replace checkbox-based filters with True/False chip toggles. 
Support filtering not only for presence or absence of notes, prior knowledge status but also absence of these and dynamically loaded project tags. False-state chips get a red tint; labels truncate with native tooltips on hover.

Currently: 
image
image

What this pull request proposes:
image
image

Looking forward to feedback, are pull requests like this welcome?
All the best, 
Michael


michi-zuri:
Relates to 
* #2316
  image
  * I could also implement the other two, what do you think?

* #2320
  image

michi-zuri:
there was a bug in this pr where label checkboxes were not loading correctly if a label filter was being used. This is now fixed with commit 2bfd477.

PR #2548, opened 2026-06-01
Avoid OperationalError: database is locked

cskaandorp:
In multi-user projects, the model trainer occasionally collides with concurrent label writes on the per-project SQLite database, producing the user-facing message "OperationalError: database is locked. Change model on the Customize tab." This change suppresses that specific exception in the trainer. Labels are not lost, because the next label event automatically re-triggers training and picks up anything missed.

### DIFF
--- a/asreview/webapp/_tasks.py
+++ b/asreview/webapp/_tasks.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sqlite3
+
 import asreview as asr
 from asreview.models.queriers import TopDown
 from asreview.models.stoppers import IsFittable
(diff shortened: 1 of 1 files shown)


J535D165 (approved):

michi-zuri:
How long does the app wait until it decides that it's not worth waiting anymore on a lock release? The waiting time could make the app feel slow if this exception happens often.

cskaandorp:
@michi-zuri : appreciate the concern. The app doesn't wait when the trainer hits the lock, it raises immediately. Now, the catch block returns, and the next label event re-triggers training and picks up the missed cycle. After v3.0.5 (Apr 20) we've seen this fire exactly once under heavy real-world load (20+ people doing 2–4 papers/min for hours on end), so it seems the exception doesn't happen often.

PR #2549, opened 2026-06-03
Fix dropped ranking table when add_last_ranking faces lock

J535D165:
Alternative approach to the issue reported by @michi-zuri in PR https://github.com/asreview/asreview/pull/2545. 

This solution doesn't need WAL to work. WAL might be a problem for some of our users working over a network. 

Inspired by the work of Michi, I changed the verification script to this solution. It looks like everything passes. 

```python
"""Reproduce the SQLite "database is locked" data loss and verify our fix.

Background
----------
ASReview rewrites the ``last_ranking`` table on every model update. The old
implementation did this with::

    DataFrame.to_sql("last_ranking", conn, if_exists="replace")

which pandas expands into three *separately committed* steps:

    1. DROP TABLE last_ranking      (committed)
    2. CREATE TABLE last_ranking    (committed)
    3. INSERT the new ranking       (committed)

This runs in a background task thread while gunicorn workers read the same
project database from other connections. SQLite allows only one writer, so if a
worker holds a lock when step 2 or 3 runs, that step fails with
``sqlite3.OperationalError: database is locked``. Because step 1 already
committed, the table is now **gone** -- and every later reader crashes with
``no such table: last_ranking``. The review is broken until the file is
repaired by hand.

A WAL journal would avoid the lock contention, but WAL is unsafe on networked
filesystems (and ASReview may serve a project directory over a network share),
so we do not want to rely on it.

The fix (in ``asreview/database/database.py``)
----------------------------------------------
``add_last_ranking`` no longer drops the table. It writes inside a single
``BEGIN IMMEDIATE`` transaction::

    BEGIN IMMEDIATE
    CREATE TABLE IF NOT EXISTS last_ranking (...)   -- self-heal if missing
    DELETE FROM last_ranking
    executemany(INSERT ...)
    COMMIT                                           -- or ROLLBACK on failure

This makes the write **atomic** (a lock failure rolls back, leaving the
*previous* ranking intact instead of a missing table) and **self-healing** (the
``CREATE TABLE IF NOT EXISTS`` rebuilds a table that an older build already
lost). No WAL required.

This script demonstrates all three properties against the real ASReview
``Database`` class.

Run:        python demo_issue.py
Requires:   asreview (this checkout), pandas
"""

import os
import sqlite3
import tempfile

import asreview as asr
from asreview.data.record import Record

N_ROWS = 5_000
RANKING_ARGS = ("nb", "max", "balanced", "tfidf", N_ROWS)


def _make_project_db(path):
    """Create a project database with records and an initial ranking."""
    db = asr.Database(path)
    db.create_tables()
    db.input.add_records([Record(i, "foo") for i in range(N_ROWS)])
    # Fail fast instead of blocking the default 5s when a lock is contended.
    db._conn.execute("PRAGMA busy_timeout = 500")
    db.add_last_ranking(list(range(N_ROWS)), *RANKING_ARGS)
    return db


def _hold_reader(path):
    """Open a second connection and hold a shared read lock open.

    Mimics a gunicorn worker reading project state mid-request: it starts a
    SELECT and does not finish it, so SQLite keeps a shared lock on the whole
    database file and any writer's COMMIT cannot acquire the exclusive lock.
    """
    reader = sqlite3.connect(path, uri=True)
    reader.execute("BEGIN")  # open a read transaction...
    reader.execute("SELECT count(*) FROM last_ranking").fetchone()  # ...holds SHARED
    return reader


def _ranking_state(path):
    """Report the table state as seen by a fresh independent connection."""
    probe = sqlite3.connect(path, uri=True)
    try:
        n = probe.execute("SELECT count(*) FROM last_ranking").fetchone()[0]
        return f"{n} rows"
    except sqlite3.OperationalError as exc:
        return f"BROKEN ({exc})"
    finally:
        probe.close()


def experiment_1_legacy_loses_table():
    """The old non-atomic write loses the table when interrupted by a lock.

    We replay pandas' ``to_sql(if_exists='replace')`` sequence step by step on a
    raw connection and let a reader grab a lock after the DROP has committed --
    exactly the race a background ranking update hits in production.
    """
    print("Experiment 1: legacy to_sql(if_exists='replace')  [the bug]")
    with tempfile.TemporaryDirectory() as tmp:
        path = os.path.join(tmp, "project.sqlite")
        _make_project_db(path).close()
        print(f"    before write           : {_ranking_state(path)}")

        writer = sqlite3.connect(path, uri=True)
        writer.execute("PRAGMA busy_timeout = 500")

        # Step 1: pandas drops the old table and commits it.
        writer.execute("DROP TABLE last_ranking")
        writer.commit()

        # A worker now starts reading and holds a lock (the race window).
        reader = sqlite3.connect(path, uri=True)
        reader.execute("BEGIN")
        reader.execute("SELECT count(*) FROM record").fetchone()

        # Step 2: pandas tries to recreate the table -> blocked -> fails.
        try:
            writer.execute("CREATE TABLE last_ranking (record_id INTEGER)")
            writer.commit()
            print("    write                  : OK (no contention this run)")
        except sqlite3.OperationalError as exc:
            print(f"    write                  : FAIL ({exc})")
        finally:
            reader.close()
            writer.close()

        print(f"    after failed write     : {_ranking_state(path)}")
        print("    -> the ranking table is gone; readers now crash.\n")


def experiment_2_fixed_keeps_previous():
    """The fixed add_last_ranking rolls back, keeping the previous ranking."""
    print("Experiment 2: fixed add_last_ranking under lock  [atomic rollback]")
    with tempfile.TemporaryDirectory() as tmp:
        path = os.path.join(tmp, "project.sqlite")
        db = _make_project_db(path)
        print(f"    before write           : {_ranking_state(path)}")

        # A worker holds a read lock for the entire write attempt (worst case).
        reader = _hold_reader(path)
        try:
            db.add_last_ranking(list(reversed(range(N_ROWS))), *RANKING_ARGS)
            print("    write                  : OK (no contention this run)")
        except sqlite3.OperationalError as exc:
            print(f"    write                  : FAIL ({exc})")
        finally:
            reader.close()

        print(f"    after failed write     : {_ranking_state(path)}")
        # The original ranking must still be fully present and readable.
        order = db.get_last_ranking_table()["record_id"].to_list()
        intact = order == list(range(N_ROWS))
        print(f"    previous ranking intact: {intact}")

        # And once the reader is gone, the write simply succeeds.
        db.add_last_ranking(list(reversed(range(N_ROWS))), *RANKING_ARGS)
        new_order = db.get_last_ranking_table()["record_id"].to_list()
        print(f"    retry without lock     : OK ({new_order[:3]}... updated)")
        db.close()
        print("    -> no data loss, no WAL needed.\n")


def experiment_3_self_heal():
    """A table lost by an older build is rebuilt on the next read/write."""
    print("Experiment 3: recovery of an already-broken database  [self-heal]")
    with tempfile.TemporaryDirectory() as tmp:
        path = os.path.join(tmp, "project.sqlite")
        db = _make_project_db(path)

        # Simulate the damage left behind by the old code.
        db._conn.execute("DROP TABLE last_ranking")
        db._conn.commit()
        print(f"    broken state           : {_ranking_state(path)}")

        # Reads recover gracefully instead of raising "no such table".
        empty = db.get_last_ranking_table().empty
        print(f"    read recovers (empty)  : {empty}")

        # The next ranking update repopulates the table.
        db.add_last_ranking(list(range(N_ROWS)), *RANKING_ARGS)
        print(f"    after next update      : {_ranking_state(path)}")
        db.close()
        print("    -> existing broken projects repair themselves.\n")


def main():
    print(
        f"{N_ROWS} rows per ranking; default rollback-journal mode (no WAL).\n"
    )
    experiment_1_legacy_loses_table()
    experiment_2_fixed_keeps_previous()
    experiment_3_self_heal()


if __name__ == "__main__":
    main()

```


michi-zuri:
There is likely a bug in this commit that breaks type casting from SQLite to python causing issue #2554, see also reported problems in #2551 and #2552.

```
"""
Minimal repro for the BLOB `record_id` regression introduced in v3.0.7 by
#2549 (commit ff9825b1c5f1383cb7f3bfb3317d679b4d3e9a7e).

The new `_write_last_ranking` writes via `df.itertuples()` + `cur.executemany()`.
When the input frame uses pandas' nullable `Int64` dtype (the dtype declared in
RANKING_TABLE_COLUMNS_PANDAS_DTYPES), `itertuples` yields `numpy.int64` scalars,
which `sqlite3` has no adapter for and stores via the buffer protocol as BLOB.
The next `read_sql_query(..., dtype={"record_id": "Int64", ...})` then fails to
cast bytes -> Int64 with the user-facing error in #<issue>.

Replacing `Int64` with plain `int64` (or routing through `df.to_sql` as v3.0.6
did) makes the bug disappear -- which is also why the regression tests added in
#2549 do not catch it: they pass `record_ids` as a Python list, so the resulting
column infers to numpy `int64` and never exercises the broken branch.
"""
import sqlite3
import pandas as pd

# Exact write pattern from asreview/database/database.py::_write_last_ranking
def write_last_ranking(con, df, columns):
    rows = list(df[columns].itertuples(index=False, name=None))
    placeholders = ", ".join(["?"] * len(columns))
    con.execute(f"CREATE TABLE IF NOT EXISTS last_ranking "
                f"({', '.join(c + ' INTEGER' for c in columns)})")
    con.execute("DELETE FROM last_ranking")
    con.executemany(
        f"INSERT INTO last_ranking ({', '.join(columns)}) "
        f"VALUES ({placeholders})", rows)
    con.commit()

# Exact read pattern from asreview/database/database.py::get_last_ranking_table
def read_last_ranking(con):
    return pd.read_sql_query("SELECT * FROM last_ranking", con,
                             dtype={"record_id": "Int64", "ranking": "Int64"})

con = sqlite3.connect(":memory:")
columns = ["record_id", "ranking"]

# Nullable Int64 -- the dtype asreview's own DTYPES dicts declare.
df = pd.DataFrame({
    "record_id": pd.array([177, 42], dtype="Int64"),
    "ranking":   pd.array([0,   1],  dtype="Int64"),
})

write_last_ranking(con, df, columns)

print("Stored on disk:")
for typ, hx in con.execute(
        "SELECT typeof(record_id), hex(record_id) FROM last_ranking"):
    print(f"  typeof={typ}  hex={hx}")

print("\nRead-back via the asreview pattern:")
try:
    print(read_last_ranking(con))
except Exception as e:
    print(f"  {type(e).__name__}: {e}")

```

PR #2588, opened 2026-08-04
Fill missing title values from alternative RIS tags (T1/TI)

### LINKED ISSUE #1154: [RIS-Import] Error during processing of title fields T1 or TI during RIS import
### Discussed in https://github.com/asreview/asreview/discussions/1153

<div type='discussions-op-text'>

<sup>Originally posted by **felixhorch** June  8, 2022</sup>
Hello!

When processing a .RIS file exported from Citavi, I noticed an error when processing the `T1` or `TI` title fields: Apparently, the title of the publications is only recognized correctly by ASReview if the title field is labeled `TI`. But if the title is exported as `T1`, as it is apparently implemented in Citavi, ASReview does not show any titles, neither in the search nor in the review. However, as far as I understand the RIS specification, both are allowed and correct. 

Couldn't this be added relatively easily in the RIS processing?

Best regards

Felix</div>

@felixhorch thanks for reporting, what version of ASReview are you using? Thanks for the attempted debug. Will look into it.

abhinavaditya811:
Changes proposed in this pull request:

- `BaseReader.standardize_column_names` now fills missing values in a record column
  from its alternative columns, instead of taking the first alternative present and
  ignoring the rest. A dataset can use different tags for the same field across
  records, so the alternatives are complementary rather than mutually exclusive.
- Adds a regression test for a RIS file where one record uses `TI` for the title and
  another uses `T1`.

Fixes #1154.

  **Checklist**

- [x] Unit tests are added for new features and bug fixes
- [ ] Documentation is added for new features — not applicable, this is a bug fix
- [x] Title of the pull request is self-explaining

### DIFF
--- a/asreview/data/base.py
+++ b/asreview/data/base.py
@@ -29,7 +29,8 @@ class BaseReader(ABC):
     # title column. The format is {record_column_name: [list of input column names]},
     # where the list of input column names is in order from most important to least
     # important. So when the input dataset contains two possible input columns for a
-    # record column, it will pick the first it finds in the list.
+    # record column, it will pick the values of the first it finds in the list and use
+    # the other columns to fill the values that are missing.
     # If a field is not in this mapping, only the record column is allowed as input
     # column.
     __alternative_column_names__ = {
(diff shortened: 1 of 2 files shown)


EmilyWes:
Hi [abhinavaditya811](https://github.com/abhinavaditya811),

Thank you for the suggestion and the PR. We have looked into it but decided not to merge.
The current solution is very generic, which can lead to unwanted edgecases (currently empty fields that get incorrect data).
It is possible to make the implementation more specific, but we rather keep ASReview as clean as possible and leave this part to other applications/processes (like a reference manager).

Emily

PR #2597, opened 2026-08-19
Introduce new ELAS u5, h4, and l3 model parameters

timovdk:
Optimized on the new [SYNERGY+ dataset](https://doi.org/10.34894/DDCVCV), we see slightly lower loss and SD compared to u4, h3, and l2. The new ELAS models were optimized on the train set of SYNERGY+ (91 systematic reviews) and validated and compared with the previous ELAS models on the test set of SYNERGY+ (23 systematic reviews).

image

### DIFF
--- a/asreview/models/models.py
+++ b/asreview/models/models.py
@@ -16,7 +16,30 @@ from asreview.learner import ActiveLearningCycleData
 
 AI_MODEL_CONFIGURATIONS = [
     {
-        # ELAS u4 is the default model for this version ASReview LAB. The model
+        # ELAS u5 is the default model for this version ASReview LAB. The model
+        # parameters have been optimized on the SYNERGY dataset. After expert
+        # elicitation, the model and parameters have been chosen.
+        # Optimization setup: https://github.com/asreview/asreview-optuna/releases/tag/
+        "name": "elas_u5",
+        "label": "ELAS u5",
+        "type": "ultra",
+        "value": ActiveLearningCycleData(
+            querier="max",
+            classifier="svm",
+            classifier_param={"loss": "squared_hinge", "C": 0.0240, "max_iter": 5000},
+            balancer="balanced",
+            balancer_param={"ratio": 9.242},
+            feature_extractor="tfidf",
+            feature_extractor_param={
+                "ngram_range": (1, 2),
+                "sublinear_tf": True,
+                "min_df": 1,
+                "max_df": 1.0,
+            },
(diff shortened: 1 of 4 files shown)


asreview/models/models.py
```diff
@@ -79,7 +121,26 @@
         ),
     },
     {
-        # ELAS h3 is the heavy model for this version ASReview LAB. The model
+        # ELAS h4 is the heavy model for this version ASReview LAB. The model
+        # parameters have been optimized on the SYNERGY dataset. After expert
+        # elicitation, the model and parameters have been chosen.
```
J535D165: What do you mean with expert elicitation?

asreview/models/models.py
```diff
@@ -16,7 +16,30 @@
...
+            feature_extractor_param={
+                "ngram_range": (1, 2),
+                "sublinear_tf": True,
+                "min_df": 1,
+                "max_df": 1.0,
+            },
+        ),
+    },
+    {
+        # ELAS u4 is the default model for version 2 of ASReview LAB. The model
```
J535D165: Also in version 3. Maybe use an upper bound.

J535D165 (changes requested):
Thanks a lot. Im happy to see this result. I think its worth implementing even though the changes are incredibly small. 

Would you like to advertise the new synergy already in model.rst or do you open a new PR after you release synergy?

Several tests fail, partially because of an issue with ruff, but also one related to this PR. It's straightforward to fix (just change the number). Can you fix this?

J535D165:
Can you rebase? That will resolve most of the failing tests.

asreview/models/models.py
timovdk: Oh I copied this description from the previous elas models, let me be precise about how I rounded the numbers rarther than "expert elicitation".

timovdk:
> Thanks a lot. Im happy to see this result. I think its worth implementing even though the changes are incredibly small.
> 
> Would you like to advertise the new synergy already in model.rst or do you open a new PR after you release synergy?
> 
> Several tests fail, partially because of an issue with ruff, but also one related to this PR. It's straightforward to fix (just change the number). Can you fix this?

I will create the PR for model.rst once the synergy-dataset-py PR is merged, then the full user flow works as expected, so before then it will just lead to issues/discussion topics I think.

J535D165 (approved):

J535D165:
Amazing!

PR #2608, opened 2026-09-10
Fix SYNERGY Dataset Download

timovdk:
SYNERGY+ introduced a new Python tool that handles errors differently. This fix ensures datasets are downloaded correctly in case the user does not have SYNERGY+ downloaded to their device already.

This error only shows up if SYNGERY+ is not yet downloaded to your device. If you downloaded SYNERGY+ via `synergy get` first, this bug won't show up for you.

### DIFF
--- a/asreview/datasets.py
+++ b/asreview/datasets.py
@@ -22,7 +22,6 @@ __all__ = [
 
 import json
 import socket
-import tempfile
 from abc import ABC
 from abc import abstractmethod
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


J535D165:
But why does `df = sd.Dataset(self.dataset_id).to_frame()` not raise an error? It's very hard to catch this masked error now if you don't know the API in great detail (as we're seeing now). I think EAFP is the best solution here.

asreview/datasets.py
```diff
@@ -396,17 +395,20 @@ def reader(self):
...
+        # partially). Download into the default synergy cache.
+        if df.empty:
+            sd.download_raw_subset(self.dataset_id)
+            df = sd.Dataset(self.dataset_id).to_frame()
 
-        raise ValueError("Synergy dataset not found")
+        if df.empty:
+            raise ValueError(
+                f"Synergy dataset '{self.dataset_id}' is empty after download"
+            )
```
J535D165, line 409: What does it mean when the dataset is empty after download? You downloaded it, didn't you? Then it should always exist or an error occured.

timovdk:
> But why does `df = sd.Dataset(self.dataset_id).to_frame()` not raise an error? It's very hard to catch this masked error now if you don't know the API in great detail (as we're seeing now). I think EAFP is the best solution here.

Hmm I do agree, that's a good point. Let me close this PR and open one for the py tool instead. We should error, not send an empty dataset.

